PR #107, opened 2021-02-20
Show instructions for active tab

sashahonchar:
TODO: Add instructions for "Trace" tab

@dustinbyrne , could you provide the instructions for Trace diagram?

Fixes: #91


dustinbyrne:
I believe this is now stale. Feel free to re-open if that's not the case.

PR #123, opened 2021-03-13
fix: force re-render search panel when AppMap was updated

sashahonchar:


src/components/DetailsSearch.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div class="details-search">
+  <div class="details-search" :key="renderKey">
```
dustinbyrne, line 2: Usually if you require this pattern, something is wrong. In this case, the state rendered is non-reactive because it is derived from a data property. The solution would be to move it to a computed property or `props`, but I'm inclined to let this merge as-is given this component is entirely reliant on global state (also an anti-pattern). I'm sure we'll have to revise this sometime in the future, but we can refactor later.

This branch needs to be rebased. Please double-check the base branch still requires this change as well. One of the branches merged was quite large and _may_ (unlikely) have fixed this.

dustinbyrne (approved):

src/components/DetailsSearch.vue
sashahonchar: Thanks for the detailed comment, Done https://github.com/applandinc/appmap-js/pull/123/commits/a3e8a44640e8ba65e0e2f89681a2ac31ffb619bd

dustinbyrne:
I believe this was merged in with another PR.

PR #128, opened 2021-03-18
feat: add fqid for codeObjects and events

sashahonchar:


src/pages/VsCodeExtension.vue
```diff
@@ -184,6 +199,51 @@ export default {
...
+      this.$root.$emit('selectedObject', definition);
+    },
+
+    setSelectedObject(objectDefinition) {
+      const defMatch = objectDefinition.match(/^([a-z]+):(.+)/);
+      const type = defMatch[1];
+      const object = defMatch[2];
+
+      /* eslint-disable no-case-declarations */
+      switch (type) {
```
dustinbyrne, line 212: This kind of API would be useful on `AppMap`
```js
appmap.get('label:authentication')
appmap.get('event:19')
appmap.get('object:com/myorg/MyClass#myMethod')
```

src/lib/models/codeObject.js
```diff
@@ -340,4 +340,8 @@ export default class CodeObject {
         return this.name;
     }
   }
+
+  get definition() {
+    return encodeURI(`${this.type}:${this.id}`);
+  }
```
dustinbyrne, line 346: Can we name this something along the lines of `fullyQualifiedId`/`fqid`?
https://en.wikipedia.org/wiki/Fully_qualified_name

sashahonchar: Good Point and I fixed https://github.com/applandinc/appmap-js/pull/128/commits/c6b275659c7937dfb3dd27848b1bdf63b8065590

src/pages/VsCodeExtension.vue
sashahonchar: Done https://github.com/applandinc/appmap-js/pull/128/commits/c6b275659c7937dfb3dd27848b1bdf63b8065590

src/lib/models/appMap.js
```diff
@@ -61,23 +61,6 @@ export default class AppMap {
...
-    };
-
-    const fqidMatch = fqid.match(/^([a-z]+):(.+)/);
-    if (fqidMatch) {
-      result.type = fqidMatch[1]; // eslint-disable-line prefer-destructuring
-      result.object = fqidMatch[2]; // eslint-disable-line prefer-destructuring
-    }
-
-    return result;
-  }
```
dustinbyrne: After thinking about it, this doesn't really make sense because there's no first-class support for labels as objects (similar to an event or code object).

src/lib/models/codeObject.js
```diff
@@ -342,6 +342,6 @@ export default class CodeObject {
   }
 
   get fqid() {
-    return encodeURI(`${this.type}:${this.id}`);
+    return `${this.type}:${this.id}`;
```
dustinbyrne, line 345: Let the library be agnostic to the encoding. This will prevent us from having to encode/decode elsewhere.

src/pages/VsCodeExtension.vue
```diff
@@ -263,42 +263,34 @@ export default {
...
+          return;
+        }
+
+        selectedObject = events.find((e) => e.id === eventId);
+      } else {
+        selectedObject = classMap.codeObjects.find((obj) => obj.fqid === fqid);
+      }
+
+      if (selectedObject) {
+        this.$store.commit(SELECT_OBJECT, selectedObject);
```
dustinbyrne, line 292: I'd moved this to `AppMap.get`, but without an object to return for labels, ended up moving everything back inside this method.

tests/unit/components/VsCodeExtension.spec.js
```diff
@@ -0,0 +1,31 @@
...
+    const expected = {
+      'label:json': 'json',
+      'event:44': 'User.find_by_id!',
+      'class:app/models/User': 'User',
+    };
+
+    Object.entries(expected).forEach(([fqid, header]) => {
+      wrapper.vm.setSelectedObject(fqid);
+      expect(wrapper.get('details-label__header-title').text()).toMatch(header);
+    });
```
dustinbyrne, line 29: We need to make sure we're testing the method and it's working as expected. There's a couple of issues with this test, however. First, we're exporting `__esModule` somewhere in the dependency tree (`Trace.vue`). This causes an issue in Jest, but it's also necessary. We'll need a workaround.

Second, with the `__esModule` export removed, this test still raises an `Invalid Component definition` error. I'm not sure why.

dustinbyrne: See the comment here in Trace.vue for the `__esModule` issue:
https://github.com/applandinc/appmap-js/blob/main/src/components/trace/Trace.vue#L54-L56

Essentially we're trying to hit this block, but IIRC the ESM bundling causes it to fail: 
https://github.com/vuejs/vue/blob/b51430f598b354ed60851bb62885539bd25de3d8/src/core/vdom/helpers/resolve-async-component.js#L20

The explicit export resolves the issue.

dustinbyrne (approved):

Issue #129, opened 2021-03-19
A query details panel exists


dustinbyrne:
- Instead of referencing queries as events which contain a query, reference the query code object directly.
- If the query code object contains a single event, navigate directly to the event

This will clean up the usage of prepared statements:
![image](https://user-images.githubusercontent.com/8737782/111809686-a0977e00-88ab-11eb-9cc7-9dcb2cd6adc4.png)

dustinbyrne:
Here's a good sample AppMap (drop the txt extension)
[org_owasp_webgoat_GeneralLessonTest_insecureLogin.appmap.json.txt](https://github.com/applandinc/appmap-js/files/6172647/org_owasp_webgoat_GeneralLessonTest_insecureLogin.appmap.json.txt)

kgilpin:
Did we get WebGOAT working?

sashahonchar:
@dustinbyrne, Is that a correct AppMap?
Why SQL query is a child of SQL query?

![image](https://user-images.githubusercontent.com/17301016/112029086-9794fe00-8b84-11eb-9ba6-fcc750d1ce83.png)

kgilpin:
Seems impossible...

PR #132, opened 2021-03-22
feat: show SQL queries as code objects in Details panel

sashahonchar:
Display code object with type `query` in Details panel.

image

Fixes: #129

### DIFF
--- a/src/components/DetailsPanelQuery.vue
+++ b/src/components/DetailsPanelQuery.vue
@@ -1,0 +1,47 @@
+<template>
+  <div>
+    <v-details-panel-header object-type="SQL query" :title="title" />
+    <v-details-panel-list title="Events" :items="object.events" />
+  </div>
+</template>
+
+<script>
+import VDetailsPanelHeader from '@/components/DetailsPanelHeader.vue';
+import VDetailsPanelList from '@/components/DetailsPanelList.vue';
+import { getSqlLabelFromString } from '@/lib/util';
+import { SELECT_OBJECT, POP_OBJECT_STACK } from '../store/vsCode';
+
+export default {
+  name: 'v-details-panel-query',
+  components: {
+    VDetailsPanelList,
+    VDetailsPanelHeader,
+  },
+  props: {
+    object: {
+      type: Object,
+      required: true,
+    },
+  },
+  computed: {
+    title() {
+      return getSqlLabelFromString(this.object.name);
+    },
+
+    events() {
+      return this.$store.state.appMap.events.filter(
+        (e) => e.isCall() && e.codeObject && e.codeObject.id === this.object.id
+      );
+    },
+  },
+
+  created() {
+    if (this.object.events.length === 1) {
+      this.$store.commit(POP_OBJECT_STACK);
+      this.$store.commit(SELECT_OBJECT, this.object.events[0]);
+    }
+  },
+};
+</script>
+
+<style scoped lang="scss"></style>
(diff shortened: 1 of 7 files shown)


src/components/DetailsPanelQuery.vue
```diff
@@ -0,0 +1,39 @@
...
+  </div>
+</template>
+
+<script>
+import VDetailsPanelHeader from '@/components/DetailsPanelHeader.vue';
+import VDetailsPanelList from '@/components/DetailsPanelList.vue';
+import { getSqlLabelFromString } from '@/lib/util';
+
+export default {
+  name: 'v-details-panel-function',
```
dustinbyrne, line 14: Can we check if `events` only contains a single event, and if it does, automatically select it?

dustinbyrne (changes requested):

src/components/DetailsPanelQuery.vue
sashahonchar: Done. https://github.com/applandinc/appmap-js/pull/132/commits/267ef09fa0ae3b9da25552747ff4230d4d11ac96

src/components/DetailsPanelQuery.vue
```diff
@@ -0,0 +1,46 @@
...
+    },
+
+    events() {
+      return this.$store.state.appMap.events.filter(
+        (e) => e.isCall() && e.codeObject && e.codeObject.id === this.object.id
+      );
+    },
+  },
+
+  mounted() {
```
dustinbyrne, line 38: [`created`](https://vuejs.org/v2/api/#created) would be a better choice here. [`mounted`](https://vuejs.org/v2/api/#mounted) is run after the component has been rendered.

src/components/DetailsPanelQuery.vue
```diff
@@ -0,0 +1,46 @@
...
+    if (this.object.events.length === 1) {
+      this.$store.commit(SELECT_OBJECT, this.object.events[0]);
```
dustinbyrne, line 40: This has the side-effect of leaving the query code object in the selected objects stack:
image

dustinbyrne (changes requested):

src/components/DetailsPanelQuery.vue
sashahonchar: Done https://github.com/applandinc/appmap-js/pull/132/commits/2a589435914d26694bfed048ec78c74598d04621

dustinbyrne:
Is it possible to add the count badges to the label items? Similar to what we have elsewhere:
image

It's harder to find duplicate queries without some indicator.

src/components/DetailsSearch.vue
```diff
@@ -90,34 +104,54 @@ export default {
...
+        item.data = item.data.sort((a, b) => {
+          const aStr =
+            a.object instanceof CodeObject
+              ? a.object.prettyName
+              : String(a.object);
+          const bStr =
+            b.object instanceof CodeObject
+              ? b.object.prettyName
+              : String(b.object);
+          return aStr.localeCompare(bStr);
```
sashahonchar, line 152: @dustinbyrne This comparision doesn't work well with `Query` code objects, because when we call `object.prettyName` it returns `SQL Select`/`SQL Insert`/`SQL Update`, etc.
Are we okay with that?

dustinbyrne: I'm OK with this, I think we need another method to sort queries anyway.

dustinbyrne (approved):

PR #137, opened 2021-03-25
style: update ux for instructions

typaulhus:
- [x]  the word "calls" is split in two lines, "call" and "s"
- [x]  let's add a "legend" title to the legend sections
- [x]  can we make the orbs in the trace view a bit bigger to match the sizes of the orbs in the trace?
- [x]  can we make the line spacing the same for the Dependency Map and Trace legends?

### DIFF
--- a/src/components/Instructions.vue
+++ b/src/components/Instructions.vue
@@ -157,8 +157,6 @@ export default {
     position: absolute;
     right: 0;
     bottom: 100%;
-    display: flex;
-    flex-direction: column;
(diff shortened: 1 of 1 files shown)


dustinbyrne (changes requested):
Looks good! I noticed with less vertical space, the text overflows out of the instructions container. The addition of the header makes this more apparent. Can we fix it before merging?

image

dustinbyrne (approved):

PR #144, opened 2021-04-01
feat: Emit user events from the VS Code Extension

dustinbyrne:


kgilpin:
Should we add more, while we are at it? What would we like to know? How the user is progressing in their usage of AppMap?

sashahonchar (approved):

dustinbyrne:
This should be enough to capture a clear picture of how the user is moving around and interacting with the AppMap. Let's follow up with another round of information gathering after the initial telemetry release.

PR #146, opened 2021-04-02
Fix: Filter out all objects which do not originate from root events

dustinbyrne:


sashahonchar:
Looks great!
I've tested this with different `AppMap files` and it works well.

sashahonchar (approved):

PR #157, opened 2021-04-12
feat: methods to read/write application state

sashahonchar:


src/pages/VsCodeExtension.vue
```diff
@@ -307,6 +307,27 @@ export default {
       }
     },
 
+    getState() {
+      const state = {
+        currentView: this.currentView,
+        selectedObject:
```
kgilpin, line 313: Can you omit this from the object if it's `null`?

dustinbyrne (approved):

Issue #62, opened 2021-01-29
Package labels are not taken into account during layout


dustinbyrne:
If a label expands the package box, the layout is not adjusted

![image](https://user-images.githubusercontent.com/8737782/106334591-fe6afa80-6258-11eb-936e-6e8c698f2a06.png)

Issue #105, opened 2021-02-19
Class names should be simplified if they have a complex form


dustinbyrne:
Discourse AppMap has classes with very long names. In the diagrams, these names should be simplified.

<img width="709" alt="Screen Shot 2020-12-11 at 6 01 25 PM" src="https://user-images.githubusercontent.com/86395/101962771-01435a80-3bdb-11eb-9cc9-391923e3437f.png">

[DraftController_saves_a_draft_on_update.appmap.json.zip](https://github.com/applandinc/diagrams-js/files/5681985/DraftController_saves_a_draft_on_update.appmap.json.zip)

PR #164, opened 2021-04-15
fix: cut long label when package is expanded

sashahonchar:
image

Fixes: #62

### DIFF
--- a/src/lib/diagrams/components/componentDiagram/graph/groups/labelGroup.js
+++ b/src/lib/diagrams/components/componentDiagram/graph/groups/labelGroup.js
@@ -78,17 +78,40 @@ export default class LabelGroup {
       this.element.setAttribute('opacity', 0);
     }
 
-    const text = createSVGElement('text');
-    const tspan = createSVGElement('tspan');
-    tspan.setAttribute('space', 'preserve');
-    tspan.setAttribute('dy', '1em');
-    tspan.setAttribute('x', 1 + (icon ? iconSize + 6 : 0));
-    tspan.textContent = label;
-    text.appendChild(tspan);
-    this.element.appendChild(text);
+    this.textElement = createSVGElement('text');
+    this.tspanElement = createSVGElement('tspan');
+    this.tspanElement.setAttribute('space', 'preserve');
+    this.tspanElement.setAttribute('dy', '1em');
+    this.tspanElement.setAttribute('x', 1 + (icon ? iconSize + 6 : 0));
+    this.tspanElement.textContent = label;
+    this.textElement.appendChild(this.tspanElement);
+    this.element.appendChild(this.textElement);
   }
 
   getBBox() {
     return this.element.getBBox();
+  }
+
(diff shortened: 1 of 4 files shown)


src/lib/diagrams/components/componentDiagram/graph/groups/labelGroup.js
```diff
@@ -78,17 +78,40 @@ export default class LabelGroup {
...
+  }
+
+  cutToWidth(width) {
+    const textEl = this.textElement;
+    const textSpanEl = this.tspanElement;
+    let string = textEl.textContent;
+
+    while (textEl.getComputedTextLength() >= width && string.length > 0) {
+      string = string.slice(0, -1);
+      textSpanEl.textContent = `${string}...`;
```
dustinbyrne, line 112: Thinking out loud here, please let me know your thoughts.

I was originally going to suggest we trim the beginning of the text instead. For example, say you have two packages named `com.myorganization.myapplication.models` and `com.myorganization.myapplication.controllers`. The shortened form may be something like `com.myorganization.myapplication...` which doesn't help differentiate the two packages.

Instead, we could trim the beginning of the text, displaying something like `...zation.myapplication.models` and `...yapplication.controllers`. This definitely feels like an improvement to me over the former solution.

However, I was just thinking, what if we cut down the font size instead? What do you think is acceptable in this situation?

sashahonchar: Hmm, that's a good point about cutting the beginning of the label, and actually we have related issue: #105 

What about to make font-size a bit smaller and add 'title' to package name?
User will be able to see package's full name by mouse hover over label:
image

dustinbyrne: I think if we trim the end of the text we'll end up needing to change it anyway when it's requested as a UX enhancement.

I've had some more time to think about it, and I think it'd be best if we first split on package separator (`/`). Then start removing elements from the front of the array until it fits. This means we won't have any awkward partial names:
`.../myapplication/models` instead of `...zation/myapplication/models`

If we have just a single token and it's still over our size constraint, then we can break the word and trim the end:
`internationalization` becomes `international...`

What do you think?

sashahonchar: That's good idea to split on `/`. I'm going to implement it.
What do you think about add `title` to label?
Browser will show the full package name on mouse hover.

sashahonchar: Done: https://github.com/applandinc/appmap-js/pull/164/commits/48b2b8a3d03a52288186ae7ec07671340921ee30

dustinbyrne:
I've restored full Java package names. We were only using the last token to cut down on the size of the string, but this logic is way better.

dustinbyrne (approved):

Issue #38, opened 2021-01-15
Edge list items detail both to and from


dustinbyrne:
![image](https://user-images.githubusercontent.com/8737782/104746192-ef466180-571c-11eb-8121-5e66e9bbf519.png)

How should we be listing these edges? Currently we only display the `from` event. Each item in this list is actually comprised of two events: `from`, `to`.

To make things more interesting, from/to can also be a route or SQL.

PR #166, opened 2021-04-16
feat: show pair of 'from' and 'to' objects when clicking on edge

sashahonchar:
image

NB! Probably we need a styling to display from/to pairs.

Fixes: #38

### DIFF
--- a/src/components/DetailsPanelEdge.vue
+++ b/src/components/DetailsPanelEdge.vue
@@ -1,18 +1,18 @@
 <template>
   <div class="details-panel-edge">
     <v-details-panel-header object-type="Edge" :title="title" />
-    <v-details-panel-list title="Events" :items="events" />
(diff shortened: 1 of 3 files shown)


src/components/DetailsPanelEdge.vue
```diff
@@ -32,7 +32,8 @@ export default {
       return from.allEvents
         .map((e) => e.children)
         .flat()
-        .filter((e) => e.codeObject.ancestors().includes(to));
+        .filter((e) => e.codeObject.ancestors().includes(to))
+        .map((e) => ({ from: e.parent.codeObject, to: e.codeObject }));
```
dustinbyrne, line 36: `to`/`from` should both link directly to an event.

dustinbyrne (changes requested):

Issue #165, opened 2021-04-16
Selected events don't always highlight an object in the dependency map


dustinbyrne:
via https://5fe18baf66f2280021e634d0-skuggsmcdc.chromatic.com/iframe.html?id=pages-vs-code--extension&viewMode=story

Given the following event is selected:
![image](https://user-images.githubusercontent.com/8737782/115050198-82706e00-9ea9-11eb-83fe-09d0c1049a7c.png)

And the user clicks the dependency map tab, then no object is highlighted:
![image](https://user-images.githubusercontent.com/8737782/115050236-8d2b0300-9ea9-11eb-8283-e60fddc16e1d.png)

It's expected that `Spee::Admin::RootController` would receive a highlight.

PR #167, opened 2021-04-17
fix: highlight function's class on Dependency Map when function was selected in Trace view

sashahonchar:
Fixes: #165

### DIFF
--- a/src/components/DiagramComponent.vue
+++ b/src/components/DiagramComponent.vue
@@ -70,6 +70,10 @@ export default {
         return;
       }
 
+      if (codeObject.type === CodeObjectType.FUNCTION) {
+        codeObject = codeObject.classObject;
+      }
+
       if (this.componentDiagram.hasObject(codeObject)) {
         this.componentDiagram.highlight(codeObject);
       } else {


src/components/DiagramComponent.vue
```diff
@@ -70,6 +70,10 @@ export default {
         return;
       }
 
+      if (codeObject.type === CodeObjectType.FUNCTION) {
+        codeObject = codeObject.classObject;
+      }
```
dustinbyrne, line 75: The `else` block from the following `if` statement is supposed to handle this case (https://github.com/applandinc/appmap-js/blob/273d71ae819d14c27e4200a1f73778dafe4839df/src/components/DiagramComponent.vue#L80-L82). Adding this statement here short circuits that logic, but do you know why it wasn't working previously?

sashahonchar: I spent some time to debug why it didn't work and this is what I discovered:
when we expand first parent:
https://github.com/applandinc/appmap-js/blob/923a4021fe137c4107042f07a897b6d03295b2aa/src/components/DiagramComponent.vue#L87

function `highlightSelectedComponent` is called again:
https://github.com/applandinc/appmap-js/blob/923a4021fe137c4107042f07a897b6d03295b2aa/src/components/DiagramComponent.vue#L110

but handler is called too with event (which you selected in Trace view):
https://github.com/applandinc/appmap-js/blob/923a4021fe137c4107042f07a897b6d03295b2aa/src/components/DiagramComponent.vue#L49-L53

so `highlightSelectedComponent` is called once again with function code object, since we doesn't show functions in Dependency Map it's not highlighted

Issue #40, opened 2021-01-15
Event list items are visibly distinct


dustinbyrne:
![image](https://user-images.githubusercontent.com/8737782/104746817-c377ab80-571d-11eb-9192-c4c240a675ad.png)

Is there a better way we can list distinct events without including a numeric ID?

ptrdvrk:
We could visually separate function calls from REST and SQL calls, and within function calls, identify calls within the same package vs. calls to other packages?

PR #168, opened 2021-04-19
feat: view event in Trace from function events list

sashahonchar:
Show event in Trace view when clicking on eye icon in function events list

image

Fixes: #40


src/components/DetailsPanelFunction.vue
```diff
@@ -11,7 +11,11 @@
         </v-details-button>
       </template>
     </v-details-panel-header>
-    <v-details-panel-list title="Events" :items="object.events" />
+    <v-details-panel-list
+      title="Events"
+      :items="object.events"
+      :event-quickview="true"
```
dustinbyrne, line 17: This property should be present for any detail panel that lists events (routes, labels, etc.). To that point, should the quick view icon just be present for any list item that contains an event? It feels like we're missing some abstraction here.

src/components/trace/TraceEventBlock.vue
```diff
@@ -120,6 +120,13 @@ export default {
       height: 0,
     };
   },
+  watch: {
+    '$store.state.focusedEvent': {
+      handler() {
+        this.initialize();
+      },
+    },
+  },
```
dustinbyrne, line 129: I'd consider adding this as a prop instead of adding it to the global store. In the past, repurposing components which have a hard dependency on the store has been difficult.

dustinbyrne (changes requested):
Please consider the items I've commented on and add a test that clicks on the 'quick view'.

src/components/trace/TraceEventBlock.vue
sashahonchar: I made it with `watch` because I thought it's not necessary to pass a prop through long chain:
VsCodeExtention -> DiagramTrace -> Trace -> TraceEventBlock

Now focused event is passing as prop: https://github.com/applandinc/appmap-js/pull/168/commits/87b4aa91f0c1f4f00155bb050d2da0dfb0838de2

src/components/DetailsPanelFunction.vue
sashahonchar: Done: https://github.com/applandinc/appmap-js/pull/168/commits/cef7b01d1182af41ae67c17bbf5accc41053f1df

src/components/trace/TraceEventBlock.vue
dustinbyrne: In the future, I think we'll end up needing to add this state directly to the `Event` object. I agree, propagating props through many components (they call this prop drilling) feels a bit odd, but it ends up keeping the components more clean as they don't rely on global state.

src/pages/VsCodeExtension.vue
```diff
@@ -194,6 +195,18 @@ export default {
         );
       },
     },
+    '$store.getters.focusedEvent': {
+      handler(event) {
+        if (event) {
+          this.setView(VIEW_FLOW);
+          window.requestAnimationFrame(() => {
```
dustinbyrne, line 202: I'm not sure if this is reliable. If you're waiting for the DOM to update, consider `this.$nextTick(() => { ... })`?
https://vuejs.org/v2/api/#Vue-nextTick

dustinbyrne (changes requested):

src/pages/VsCodeExtension.vue
sashahonchar: Yes, inside Vue component it's better to use Vue methods.
Done: https://github.com/applandinc/appmap-js/pull/168/commits/41108ac4f7366b2505d5d038559cd59f32b6a2ce

dustinbyrne (approved):

PR #173, opened 2021-04-20
Manage app state with getState/setState API

sashahonchar:
**NB!** There is a breaking change, because `setSelectedObject` method and `selectedObject` event were removed from `VsCodeExtension` component.

With this PR we must read/write app state only with:

- `getState/setState` methods
- `stateChanged` event

### DIFF
--- a/src/pages/VsCodeExtension.vue
+++ b/src/pages/VsCodeExtension.vue
@@ -172,7 +172,7 @@ export default {
       handler(view) {
         this.onChangeTab(this.$refs[view]);
         this.$refs.tabs.activateTab(this.$refs[view]);
-        this.$root.$emit('stateChanged');
+        this.$root.$emit('stateChanged', 'currentView');
       },
     },
     '$store.getters.selectedObject': {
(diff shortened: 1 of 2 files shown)


dustinbyrne (changes requested):
In the extension, we listen for the `selectedObject` event to send telemetry data on what kinds of objects our users are inspecting. With this change, we don't have any notion of what caused the state change. In other words, `stateChange` may be fired because the user selected a new object or due to the view changing. Can we find a way of supporting the telemetry use case?

Thinking out loud, a few options:
- We send both `stateChanged` and `selectedObject` upon selecting a new object. As of the current changes, [we're still sending `changeTab`](https://github.com/applandinc/appmap-js/blob/feature/app-state/src/pages/VsCodeExtension.vue#L276) in addition to `stateChanged` when the view changes.
- When emitting `stateChanged`, we send the key that has changed along with the event. Then the consumer can decide how to respond based on which key has changed.
- We roll up all the state and telemetry events (`changeView`, `selectedObject`, `showInstructions`, `clearSelection`) into a new `userAction` event which indicates the action and, if applicable, the key of the state change.

sashahonchar:
Second option with key which has been changed looks good. https://github.com/applandinc/appmap-js/pull/173/commits/5e2041bd3c7965161a4bbfa9664f4039bb9a4aeb

PR #175, opened 2021-04-21
fix: Depends fixes

kgilpin:
Fix bugs in `depends` command:

* Log to stderr instead of stdout, to avoid polluting CLI output
* Fix handling of `baseDir`
* Yield results incrementally


.eslintrc.js
```diff
@@ -10,6 +10,7 @@ module.exports = {
   },
   rules: {
     'no-param-reassign': ['error', { props: false }],
+    'no-underscore-dangle': ['off'],
```
dustinbyrne, line 13: Is this necessary?
https://airbnb.io/javascript/#naming--leading-underscore

At the very least, `allowAfterThis` seems more applicable. We've been side-stepping elsewhere by prepending 'internal' class member variables with `$` when using getters/setters that perform additional calculations. Everything is otherwise assumed public.
https://eslint.org/docs/rules/no-underscore-dangle#allowafterthis

If it's overly cumbersome, we can move things around to better support multiple ESLint configurations in the future.

dustinbyrne (changes requested):

dustinbyrne:
I've changed it to allow dangling underscores after `this`, allowing ESLint to verify that it's not accessed outside of the class definition scope.

dustinbyrne (approved):

sashahonchar (approved):

Issue #171, opened 2021-04-19
HTTP server request `status` comes from `.status || .status_code`


dustinbyrne:
According to the specification, the status code returned from an HTTP request should be recorded in the response as `http_server_response.status_code` (https://github.com/applandinc/appmap#http-server-request-return-attributes). 

However, the Ruby and Java clients record it instead as `http_server_response.status`.

Add support for both these properties. The status code should be present in the trace view for Ruby, Java and Python.

PR #184, opened 2021-04-26
fix: display HTTP server response status in 'status' property

sashahonchar:
In Details panel HTTP `status` will be displayed with status key, despite [`the official specification`](https://github.com/applandinc/appmap#http-server-request-return-attributes) where HTTP response event must have required `status_code` key.

image

Fixes: #171

### DIFF
--- a/src/lib/models/event.js
+++ b/src/lib/models/event.js
@@ -106,12 +106,7 @@ export default class Event {
   }
 
   get httpServerResponse() {
-    const response = { ...this.returnEvent.http_server_response };
-    if (response.status_code) {
-      response.status = response.status_code;
-      delete response.status_code;
-    }
-    return response;
+    return this.returnEvent.http_server_response;
   }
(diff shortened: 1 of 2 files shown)


src/lib/models/event.js
```diff
@@ -106,7 +106,12 @@ export default class Event {
   }
 
   get httpServerResponse() {
-    return this.returnEvent.http_server_response;
+    const response = { ...this.returnEvent.http_server_response };
+    if (response.status_code) {
+      response.status = response.status_code;
+      delete response.status_code;
+    }
+    return response;
```
dustinbyrne: I moved this to occur during normalization as I think it makes the most sense. I also wanted to avoid the allocation on line 109. I'm not sure if it would break anything in this case, but we want to maintain stable references where ever we can.

dustinbyrne (approved):

PR #195, opened 2021-04-30
Monorepo

dustinbyrne:
Move the repository structure to a monorepo. Packages are placed in the root level `packages` directory and are versioned independently via `semantic-release`.

`npm` replaced with `yarn`.

`lerna` is a tool which makes it trivial to run commands on each repository. There's some overlap with `yarn`, so it's worth taking a look at the documentation.


kgilpin:
Cool. Is this at all a standardish JS thing to do? Curious.

dustinbyrne:
Lerna monorepos are pretty common. Check out [`babel`](https://github.com/babel/babel) as a good example.

Issue #206, opened 2021-05-06
Bug: non-responsive package org.hsql.jdbc in an AppMap recorded from a Java WebApp


ptrdvrk:
Package org.hsql.jdbc cannot be selected or expanded in the Dependency Map

AppMap file: https://app.land/api/scenarios/3b14b4b8-1e07-438e-86ca-acf42c69f527
AppMap: https://app.land/scenarios/3b14b4b8-1e07-438e-86ca-acf42c69f527?state=%257B%2522currentView%2522%3A%2522viewComponent%2522%257D

appmap.yml:
```yaml
name: bodgeit
packages:
  - path: org.apache.jsp
```

PR #207, opened 2021-05-06
fix: filter out ClassMap nodes without events

sashahonchar:
With this change we don't filter out CodeObjects array anymore because it used by Dependency Map where packages/classes without events are still present.

Fixes: #206

### DIFF
--- a/packages/components/tests/e2e/specs/vsCodeExtension.spec.js
+++ b/packages/components/tests/e2e/specs/vsCodeExtension.spec.js
@@ -483,7 +483,7 @@ context('VS Code Extension', () => {
         .contains('Functions')
         .parent()
         .within(() => {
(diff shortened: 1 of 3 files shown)


dustinbyrne:
If a CodeObject has no events, it shouldn't be visible. It effectively has no relevance to the rest of the objects or events.

sashahonchar:
Done: https://github.com/applandinc/appmap-js/pull/207/commits/7d27353ee774fca3ec78cff52a3f335cd7254c50

kgilpin:
@dustinbyrne i agree we don’t want these to be visible in the UI, but we don’t want them to be removed from the AppMap completely. 

Sometimes there are source locations such as view template files that will be in the class map but have no associated event (due to the way view rendering works).

PR #210, opened 2021-05-07
feat: display only details panel in mobile views

sashahonchar:
Display only Details panel for screens width less than 800px.

image

Fixes: #208


dustinbyrne:
Stale

PR #211, opened 2021-05-08
feat: add Upload button to VSCode webview

artemvaskoo:
Add Upload button to vscode extension page


kgilpin:
This control is also used in app.land so the upload button should be disabled there?

sashahonchar (approved):

kgilpin:
What happens in other environments? app.land and JetBrains?

dustinbyrne:
I was expecting to see this as a command in the extension instead of a button in the view. Is this the right place for it?

artemvaskoo:
Updated the component to have a prop that specifies if appmap is uploadable, false by default

Issue #198, opened 2021-05-04
HTTP client request support


dustinbyrne:
- [x] Event has an `httpClientRequest` accessor
- [x] Event has an `httpClientResponse` accessor
- [x] HTTP client requests are styled in the Trace view
- [x] HTTP client request section is present in the search panel
- [x] HTTP client requests exist in the class map similar to HTTP server requests
- [x] Dependency map has a custom style for HTTP client requests

PR #213, opened 2021-05-11
feat: HTTP client requests

sashahonchar:
image

image

AppMap for tests: 
[http_client_requests.json.txt](https://github.com/applandinc/appmap-js/files/6461826/http_client_requests.json.txt)

Fixes: #198

### DIFF
--- a/packages/components/src/components/DetailsPanelHttp.vue
+++ b/packages/components/src/components/DetailsPanelHttp.vue
@@ -1,9 +1,6 @@
 <template>
   <div>
-    <v-details-panel-header
-      object-type="HTTP server requests"
-      :object="object"
-    />
+    <v-details-panel-header :objectType="object.data.name" :object="object" />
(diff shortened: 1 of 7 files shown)


dustinbyrne:
@Daniel-Warner-X @typaulhus We need a design for HTTP client requests. These events occur when the application makes an outbound request to some other service.

dustinbyrne (changes requested):
While we wait for an official design, can we change the colors of the client requests to differentiate it from a server request? Then we can merge as a v0.

dustinbyrne:
Tracking HTTP request visual updates in #231.

dustinbyrne:
Let's cut it down to two options:
- [x] Limit root events to HTTP
- [x] Show unlabeled events 

`Limit root events to HTTP` should default to on (as it is now) if at least one event contains a HTTP server request.
`Show unlabeled events` should be on by default.

dustinbyrne:
Moved to #257

Issue #214, opened 2021-05-13
Configurable filters can be applied globally


dustinbyrne:
The filters used in the diff view are ported to the VS code extension page and applied to all code objects

PR #217, opened 2021-05-18
feat: global filters

sashahonchar:
Button with filters was added to top right panel next to reload button.

image

Fixes: #214


dustinbyrne (changes requested):
Looking good! I found an issue where, after unselecting all boxes, the empty AppMap screen appears. I've added an acceptance test to test this case. Once fixed, this is good to merge.

kgilpin:
The view I will want is to hide unlabeled functions. Is this part of the
design?

kgilpin:
Instead of "unlabeled events" and "labeled events" we may want to say
"unlabeled functions" and "labeled functions", since other code objects
don't have labels.

Also, what is the point of hiding HTTP server requests and SQL?

kgilpin:
I would expect to see "events outside an HTTP server request" be hidden by
default, but an option to bring them back (when applicable)

sashahonchar:
>
Also, what is the point of hiding HTTP server requests and SQL?
I took those filters from Diff page, so I left all of them

> I would expect to see "events outside an HTTP server request" be hidden by default, but an option to bring them back (when applicable)
What do you mean? @kgilpin  As I see all events have a root of HTTP server request.

packages/components/src/pages/Diff.vue
```diff
@@ -9,57 +9,53 @@
...
+              <label for="labeled-events">Labeled events</label>
+            </div>
+
+            <div>
+              <input
+                type="checkbox"
+                id="http-server-requests"
+                v-model="filters.http.on"
+              />
+              <label for="http-server-requests">HTTP server requests</label>
```
kgilpin, line 45: Why would I disable viewing http events?

packages/components/src/pages/Diff.vue
```diff
@@ -9,57 +9,53 @@
...
+            </div>
+
+            <div>
+              <input
+                type="checkbox"
+                id="sql-queries"
+                v-model="filters.sql.on"
```
kgilpin, line 52: Why would I disable SQL?

packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -73,6 +73,46 @@
...
+                v-model="filters.allRootEvents.on"
+              />
+              <label for="all-root-events">All root events</label>
+            </div>
+            <div>
+              <input
+                type="checkbox"
+                id="all-http-requests"
+                v-model="filters.allHttpRequests.on"
+              />
```
kgilpin, line 111: Why would I disable this?

packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -73,6 +73,46 @@
...
+              <label for="unlabeled-events">Unlabeled events</label>
+            </div>
+            <div>
+              <input
+                type="checkbox"
+                id="labeled-events"
+                v-model="filters.labeledEvents.on"
+              />
+              <label for="labeled-events">Labeled events</label>
```
kgilpin, line 96: Why would I disable labeled events?

kgilpin:
@dustinbyrne i don’t understand why the user would want to hide sql, http server requests, or labeled events. These filters don’t seem to have a plausible user need to me. What am I missing?

sashahonchar:
I checked [your comment](https://github.com/applandinc/appmap-js/pull/213#issuecomment-853409924) and list of available filters now is:
- Limit root events to HTTP
- Show unlabeled events
- Hide media HTTP requests

packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -245,11 +295,48 @@ export default {
...
+        rootEvents = rootEvents.filter((e) => e.httpServerRequest);
+      }
+
+      let events = rootEvents.reduce((callTree, rootEvent) => {
         rootEvent.traverse((e) => callTree.push(e));
         return callTree;
       }, []);
 
+      events = events.filter(
+        (e) => this.filters.unlabeledEvents.on || e.labels.size > 0
```
dustinbyrne, line 310:
```js
this.filters.unlabeledEvents.on ||
e.labels.size > 0 ||
e.codeObject.type !== CodeObjectType.FUNCTION
```

This filter should only apply to functions.

packages/components/tests/e2e/specs/vsCodeExtension.spec.js
```diff
@@ -524,6 +524,35 @@ context('VS Code Extension', () => {
...
+    it('shows all root events with disabled filter', () => {
+      cy.get('.tabs .tab-btn').last().click();
+
+      cy.get('.trace-node').should('have.length', 4);
+
+      cy.get('.popper__button').click();
+      cy.get('#limit-root-events').click();
+
+      cy.get('.trace-node').should('have.length', 12);
+    });
```
dustinbyrne, line 555: Please add a test for the media request filter.

packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -245,11 +295,48 @@ export default {
...
+          'text/css',
+          'video/.+',
+        ];
+        const excludedEvents = [];
+        events.forEach((e) => {
+          if (
+            e.http_server_response &&
+            e.http_server_response.mime_type &&
+            mediaTypes.some((type) =>
+              new RegExp(type, 'ig').test(e.http_server_response.mime_type)
```
dustinbyrne, line 331: Compiling the regex for each iteration of this loop is quite slow. Can we compile these once?
https://jsbench.me/tnkph4i56y/1

packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -193,6 +229,20 @@ export default {
...
+          on: true,
+          default: true,
+        },
+        unlabeledEvents: {
+          on: true,
+          default: true,
+        },
+        hideMediaRequests: {
+          on: true,
+          default: true,
```
dustinbyrne, line 243: What does the `default` key imply? Wouldn't the default be the initial value of `on`?

packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -193,6 +229,20 @@ export default {
       versionText: '',
       VIEW_COMPONENT,
       VIEW_FLOW,
+      filters: {
+        limitRootEvents: {
+          on: true,
+          default: true,
```
dustinbyrne, line 235: The default of this filter should depend on whether or not the event array contains any HTTP server requests.

If there **aren't** any HTTP server requests, it should be off by default so that we don't display an empty Trace.
If there **are** HTTP server requests, it should be on by default to reduce noise.

dustinbyrne (changes requested):
Looking good. I left a few minor comments.

packages/components/src/pages/VsCodeExtension.vue
sashahonchar: `default` key is used in computed property `filtersChanged` to determine any changes in filters since initial load:
https://github.com/applandinc/appmap-js/blob/feat/global-filters/packages/components/src/pages/VsCodeExtension.vue#L393-L398

We need it to show small tip on filters icon when something was changed

packages/components/src/pages/VsCodeExtension.vue
sashahonchar: Added new condition

packages/components/tests/e2e/specs/vsCodeExtension.spec.js
sashahonchar: I've added new HTTP server request event to `scenario.json` and added new test case for media requests filter. https://github.com/applandinc/appmap-js/pull/217/commits/82134753fbbc8ed84c34ccf49e283e08dc3dc948

packages/components/src/pages/VsCodeExtension.vue
sashahonchar: Good point. Done https://github.com/applandinc/appmap-js/pull/217/commits/4c26a606138eb215a3cd4b7d3ff78bf88c329d54

packages/components/src/pages/VsCodeExtension.vue
sashahonchar: Done https://github.com/applandinc/appmap-js/pull/217/commits/4c26a606138eb215a3cd4b7d3ff78bf88c329d54

kgilpin:
Personally, I would like to have this feature available.

dustinbyrne (changes requested):
This looks and works great. Please look into the following issue and then we can merge:
See https://www.chromatic.com/test?appId=5fe18baf66f2280021e634d0&id=61157283d41976003bde8a1d

sashahonchar:
@dustinbyrne , I think this is normal behavior because when we detect the `appmap` without HTTP root events, we disable the first filter "Limit root events to HTTP":
https://github.com/applandinc/appmap-js/blob/feat/global-filters/packages/components/src/pages/VsCodeExtension.vue#L409-L411

In the Chromatic screenshot, there is the pink dot on the filter icon, so it means we disabled the first filter to try to display something on the diagram.

dustinbyrne:
From an end-user perspective, there's nothing being displayed so it should tell me why. If we try to recover by disabling some filters that's good, but if it still fails to find anything useful, we should continue to fall back to the "no data" screen.

dustinbyrne (approved):

PR #228, opened 2021-05-25
fix: trace lines rendered with CSS

sashahonchar:
Before:
image

After:
image

Fixes: #82, #75


dustinbyrne (changes requested):
Fantastic work! Once the test is fixed, this is good to merge.

dustinbyrne (approved):

PR #239, opened 2021-06-01
fix: Improve bind events

kgilpin:
Improve the performance and correctness of the code which binds events to code objects.

When loading many AppMaps, the `O(N*M)` nature of event registration was a problem. 

Scanning all code objects (M) for every event (N) is the issue: https://github.com/applandinc/appmap-js/pull/239/files#diff-2208f30747bef29ba042cc56931d47112b3bf25d8b47d9760ad31178e249a5e8L84

### DIFF
--- a/packages/models/src/classMap.js
+++ b/packages/models/src/classMap.js
@@ -1,33 +1,8 @@
 import CodeObject, { CodeObjectType } from './codeObject';
 
-// Recursively add a code object to an array
-// If the code object exists in the array already, add its children to the
-// existing code object instead.
-function addCodeObject(codeObjectArray, codeObject, parent = null) {
-  if (
-    !codeObjectArray ||
-    !Array.isArray(codeObjectArray) ||
-    !codeObject ||
-    !(codeObject instanceof CodeObject)
-  ) {
-    return;
-  }
-
-  // TODO.
-  // This ignores static/non-static function collisions and function overloads, though this method
-  // is never currently called in a context where those edge cases exist.
-  const existingObject = codeObjectArray.find(
-    (obj) => obj.type === codeObject.type && obj.name === codeObject.name
-  );
-
-  if (!existingObject) {
-    codeObject.parent = parent; // eslint-disable-line no-param-reassign
-    codeObjectArray.push(codeObject);
-  } else {
-    codeObject.children.forEach((child) =>
-      addCodeObject(existingObject.children, child, existingObject)
-    );
-  }
+function indexCodeObject(co, codeObjects, codeObjectsById) {
+  codeObjects.push(co);
+  codeObjectsById[co.id] = co;
 }
 
 export default class ClassMap {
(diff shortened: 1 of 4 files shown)


packages/models/src/classMap.js
```diff
@@ -1,64 +1,40 @@
...
-
-  if (!existingObject) {
-    codeObject.parent = parent; // eslint-disable-line no-param-reassign
-    codeObjectArray.push(codeObject);
-  } else {
-    codeObject.children.forEach((child) =>
-      addCodeObject(existingObject.children, child, existingObject)
-    );
-  }
+function indexCodeObject(co, codeObjects, codeObjectsById) {
```
kgilpin, line 3: This is shared across the constructor and the event registration.

packages/models/src/classMap.js
```diff
@@ -1,64 +1,40 @@
...
+  codeObjects.push(co);
+  codeObjectsById[co.id] = co;
 }
 
 export default class ClassMap {
   constructor(classMap) {
     this.codeObjectsByLocation = {};
+    this.constructedCodeObjects = {};
```
kgilpin, line 11: These are the only code objects for which we used `event.toString()` as a key. Other code objects are keyed by location.

packages/models/src/classMap.js
```diff
@@ -1,64 +1,40 @@
...
-          this.codeObjectsByLocation[location] = codeObjects;
-        }
-        codeObjects.push(co);
+      (data.children || []).forEach((child) => {
+        buildCodeObject(child, co);
       });
-    }
 
-    return co;
+      if (co.type !== 'package') {
```
kgilpin, line 23: I'm not sure why packages are excluded from this logic, but I've left it.

packages/models/src/classMap.js
```diff
@@ -81,9 +57,27 @@ export default class ClassMap {
   }
 
   codeObjectFromEvent(event) {
-    return this.codeObjects.find(
-      (co) => co.id.indexOf(event.toString()) !== -1
-    );
+    let codeObject;
+    // These types of events should not be reporting path and lineno, but sometimes
+    // they do.
+    if (!(event.httpServerRequest || event.httpClientRequest || event.sql)) {
```
kgilpin, line 63: This is a point of weakness in the code; if we add new event types, we may need to add them to this exclusion clause.

packages/models/src/classMap.js
```diff
@@ -81,9 +57,27 @@ export default class ClassMap {
...
+      const { path, lineno } = event;
+      const location = [path, lineno].filter((e) => e).join(':');
+      if (location !== '') {
+        const codeObjects = this.codeObjectsAtLocation(location);
```
kgilpin, line 67: This is the way that the agent spec expects events and code objects to be correlated.

packages/models/tests/unit/fixtures/spring_petclinic.json
```diff
@@ -1,973 +1,963 @@
...
+      "method_id": "testInitCreationForm"
+    },
+    "framework": {
+      "name": "junit"
+    }
+  },
+  "events": [
+    {
+      "event": "call",
+      "http_server_request": {
```
kgilpin, line 31: "Function call" fields are removed from this event, per the AppMap spec.

dustinbyrne (changes requested):
See comments below. `constructedCodeObjects` needs property assignments, or we can safely remove it at a slight performance cost.

packages/models/src/classMap.js
```diff
@@ -116,12 +110,38 @@ export default class ClassMap {
...
+              }
+              indexCodeObject(
+                newCodeObject,
+                this.codeObjects,
+                this.codeObjectsById
+              );
+            }
+
+            return newCodeObject;
+          };
```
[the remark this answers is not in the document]
kgilpin: When I see something nested I know it's not going to be called from anywhere else. So that helps me understand its role within the codebase.

packages/models/src/classMap.js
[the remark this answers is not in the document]
kgilpin: Right, good point. I need to take a look.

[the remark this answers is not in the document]
kgilpin: I've removed this, so I think this branch is ready to merge now.

dustinbyrne (approved):

PR #249, opened 2021-06-11
feat: initial commit

lachrist:
Initial version, I'm not sure I setup things correctly though. Also have not fully tested the schema yet so it might not be 100% accurate.


kgilpin:
The coverage may be checked in by accident?

dustinbyrne (approved):
I'm having some issues validating [`scenario.json`](https://github.com/applandinc/appmap-js/blob/c4776a0514c333746846b8ffca88465f8c2739ee/packages/components/src/stories/data/scenario.json), so there's likely some more work to be done. Though I haven't dug in to verify whether it's an issue with the schema itself or a violation of the AppMap spec.

In any case, I don't want to hold this PR up. We can continue to iterate on this.

lachrist:
I believe it is because it is version 1.3 and I validate against 1.6. I'm not sure whether I should spend some time to support legacy versions.

lachrist:
I mean I explicitely look for `"version": "1.6"`

PR #255, opened 2021-06-15
fix: Trace nodes are always visible after keyboard navigation

dustinbyrne:


packages/components/src/components/Container.vue
```diff
@@ -31,8 +31,10 @@ export default {
     translateTo(x, y) {
       this.container.translateTo(x, y);
     },
-    panToElement(element) {
-      panToNode(this.container, element);
+    panToElement(element, parent = null) {
+      const parentContainer = parent ?? this.container.element;
+      const coords = getParentRelativeOffset(element, parentContainer);
+      this.container.translateTo(coords.left, coords.top);
```
sashahonchar, line 37: Now we will use this method for panning to Trace node

packages/components/tests/e2e/specs/vsCodeExtension.spec.js
```diff
@@ -542,6 +559,21 @@ context('VS Code Extension', () => {
...
+        .click()
+        .type('{rightarrow}');
+
+      cy.get('body').trigger('keydown', { keycode: 38 }); // arrow up
+      cy.get('.trace-node.highlight').should('be.visible');
+
+      for (let i = 0; i < 50; ++i) {
+        cy.get('body').trigger('keydown', { keycode: 40 }); // arrow down
+        cy.get('.trace-node.highlight').should('be.visible');
+      }
```
sashahonchar, line 575: I changed Cypress code to press keyboard arrows keys

sashahonchar (approved):

dustinbyrne:
@sashahonchar Can you take a look at this?
https://www.chromatic.com/test?appId=5fe18baf66f2280021e634d0&id=60ca5116c637e00039e81fd0

dustinbyrne:
It looks like it's centering the bottom right corner of these nodes. It's not really noticeable, except when the node is larger than average and the viewport is on the smaller side. I'd like to see it focus the center of the bounding box instead.
image

sashahonchar (approved):

sashahonchar:
I just reviewed your fix of node centering and it looks good @dustinbyrne

PR #267, opened 2021-07-01
feat: quickstart steps

### LINKED ISSUE #259: Initial outline of the step-by-step installation wizard exists
__Figma:__ https://www.figma.com/proto/fqM92VioD2fKNkwcz5VYjO/user-milestones?node-id=461%3A10475&scaling=min-zoom&page-id=461%3A10452

Acceptance criteria:
- [x] A single page exists to step through each step
- [x] When creating the page, a property exists to specify which step is initially visible
-  [x] The creator of the page can request a callback on button clicks which perform an action (e.g. `create the appmap.yml file`, `install the gem`)
- [x] Action buttons resolve asynchronously: Upon clicking the button, a callback is called, returning a promise. The button should be in the 'loading' state until the promise is resolved, when it will be replaced with a 'next' button.
- [x] An end-to-end test clicks through the entirety of the walkthrough.

Example Vue definition for creating the page:
```ts
const app = new Vue({
  el: '#app',
  props: {
    language: 'ruby', // The language detected
    configuration: 'name: my_git_project\n\npackages:\n  - path: app/models', // The content of appmap.yml
    initialStep: Steps.CREATE_CONFIGURATION, // The step we should open to
    stepsComplete: [Steps.INSTALL_AGENT], // Which steps are completed so far

    // This callback is called when an action button is clicked, allowing the owner to handle these events. A promise is
    // returned, running the length of the action, and resolving (or throwing an exception) once complete.
    onAction(language: string, stepId: string): Promise<void> {
      switch(step) {
        case Steps.INSTALL_AGENT: {
          return await agent.install();
        }

        case Steps.CREATE_CONFIGURATION: {
          return await agent.init();
        }

        case Steps.RUN_TESTS: {
          return await runTests();
        }

        default: {
          throw new Error('unknown step');
          // TODO.
        }
      }
    },
  }
});
```

### LINKED ISSUE #271: Project selection UI exists
Figma: https://www.figma.com/file/fqM92VioD2fKNkwcz5VYjO/user-milestones?node-id=571%3A6967

```ts
interface ProjectOption = { name: string, path: PathLike };
```

The component accepts an array of `ProjectOption` objects and emits an event after a selection has been made. The event contains the option selected.

### LINKED ISSUE #272: "Record AppMap" steps displays detected test framework and command.
Currently, the displayed detected test framework is hardcoded to "Minitest".

### LINKED ISSUE #273: "Record AppMap" step displays actual test command.
Currently, the step always displays a minitest command line.

### LINKED ISSUE #274: "Record AppMap" step: user can edit the test command prior to its execution.
Currently, users cannot modify the test command for recording AppMap.

### LINKED ISSUE #275: "Record AppMap" step: when no test command detected, allow user to enter their test command prepended by "APPMAP=true".
Currently, an "empty command" is run when no known test framework is detected, with a false positive result.

### LINKED ISSUE #277: "Record AppMap" step: tests can be re-run later for returning users.
Currently, one the first run of the tests have been completed, the UI stays "AppMap recorded" and user cannot run the tests using the same UI later.

### LINKED ISSUE #280: The "Contact an AppLand developer directly." link opens the new "Help me" page
Currently the link opens a wrong page.
Needs to open https://appland.com/appmap/support.html

### LINKED ISSUE #283: Remove the "appmap.yml" link from the "Configure AppMap" step.
Let's remove this link for now.

image

### LINKED ISSUE #284: "Configure AppMap" step: user can open the newly created appmap.yml file with the link in the Webview.
At this moment, the UI only displays "appmap.yml" under the success message.

image

sashahonchar:
Fixes #259 
Fixes #271
Fixes #272
Fixes #273
Fixes #274 
Fixes #275 
Fixes #277 
Fixes #280
Fixes #283 
Fixes #284 
Fixes #https://github.com/applandinc/vscode-appland/issues/267


packages/components/src/pages/Quickstart.vue
```diff
@@ -0,0 +1,571 @@
...
+            @click="nextStep"
+          >
+            <StepRightIcon class="qs-head__btn-icon" />
+          </button>
+        </div>
+      </div>
+      <section class="qs-step" v-if="currentStep === 1">
+        <div class="qs-step__head">
+          <h1 class="qs-title">Install AppMap Agent</h1>
+          <select class="qs-select" v-model="language">
```
dustinbyrne, line 32: Props should not be edited from within the component or its children. `vmodel="language"` will end up writing to the prop value and raising a console error.

packages/components/src/pages/Quickstart.vue
```diff
@@ -0,0 +1,571 @@
...
+    },
+    languages() {
+      const languages = {
+        ruby: 'Ruby',
+        java: 'Java',
+        python: 'Python',
+      };
+
+      if (Object.keys(languages).includes(this.language)) {
+        languages[this.language] += ' (detected)';
```
dustinbyrne, line 255: `(detected)` should only be appended to the language if passed by prop. If the user selects another language from the dropdown, it shouldn't display `(detected)`.

packages/components/src/pages/Quickstart.vue
```diff
@@ -0,0 +1,571 @@
...
+            <div class="qs-loader__dot"></div>
+            <div class="qs-loader__dot"></div>
+          </div>
+        </div>
+        <div class="qs-step__success" v-if="step2Completed">
+          <span class="qs-step__success-title">
+            <SuccessIcon class="qs-step__success-icon" />
+            AppMap configuration file created
+          </span>
+          <span class="qs-step__success-subtitle">appmap.yml</span>
```
dustinbyrne, line 111: Can this also be a link that emits an event upon being clicked? Then we can listen for the event from the code editor and open the file.

packages/components/src/pages/Quickstart.vue
```diff
@@ -0,0 +1,571 @@
...
+# 'name' should generally be the same as the code repo name.
+ name: my_project
+ packages:
+ - path: app/controllers
+ - path: app/models
+ # Include the gems that you want to see in the dependency maps.
+ # These are just examples.
+ - gem: activerecord
+ - gem: devise</code
+          >
```
dustinbyrne, line 96: Please allow us to supply this content via prop.

packages/components/src/pages/Quickstart.vue
```diff
@@ -0,0 +1,571 @@
...
+            </option>
+          </select>
+        </div>
+        <div class="qs-step__block" v-if="!step1Completed">
+          <p>
+            This will add the following snippet to the top of your Gemfile and
+            run bundle to install the AppMap gem.
+          </p>
+          <code class="qs-code" @click="select"
+            >gem "appmap", "= 0.53.0", :groups => [:development, :test]</code
```
dustinbyrne, line 48: How can we specify this text for the other languages (Python, Java)? We may need to break this component up into smaller components so that we may reuse them with minor content differences.

packages/components/src/pages/Quickstart.vue
```diff
@@ -0,0 +1,571 @@
...
+    },
+  },
+
+  data() {
+    return {
+      currentStep: this.initialStep,
+      step1Completed: this.completedSteps.includes(1),
+      step2Completed: this.completedSteps.includes(2),
+      step3Completed: this.completedSteps.includes(3),
+      step4Completed: this.completedSteps.includes(4),
```
dustinbyrne, line 228: I'm not 100% sure this is the behavior, but I believe `data` is run once upon component initialization and not updated when props change. So if the extension changes the `completedSteps` array, these values will not update. For this case, a `computed` property would work better.

packages/components/src/pages/Quickstart.vue
```diff
@@ -0,0 +1,571 @@
...
+          this[`step${this.currentStep}Completed`] = true;
+        }
+      } catch (e) {
+        console.error(e);
+      }
+
+      this.isActionRunning = false;
+    },
+    select(event) {
+      console.log(this.language);
```
dustinbyrne, line 295: `console.log`

packages/components/src/pages/Quickstart.vue
```diff
@@ -0,0 +1,571 @@
...
+  name: 'Quickstart',
+
+  components: {
+    HelpIcon,
+    StepLeftIcon,
+    StepRightIcon,
+    SuccessIcon,
+  },
+
+  props: {
```
dustinbyrne, line 201: Here's another prop: `error: string`

If `error` is truthy, add a code box to display the error.

packages/components/src/pages/Quickstart.vue
```diff
@@ -0,0 +1,571 @@
...
+      default: null,
+    },
+    initialStep: {
+      type: Number,
+      default: 1,
+    },
+    completedSteps: {
+      type: Array,
+      default: () => [],
+    },
```
dustinbyrne, line 213: Can we update the content of the `completedSteps` array? The milestones have three states: `complete` | `incomplete` | `error`.

Instead of an array of completed steps, the array should hold the state of all steps.
```
[  'complete', 'complete', 'error', 'incomplete' ]
```

packages/components/src/pages/Quickstart.vue
sashahonchar: How that error should be displayed?
There is no styling in Figma for error text.

dustinbyrne: I've reviewed the Figma. Let's just stick with that.
https://www.figma.com/file/fqM92VioD2fKNkwcz5VYjO/user-milestones?node-id=571%3A7490

We should display this page when `onAction` throws an exception from the install page, then re-throw the exception.

packages/components/src/pages/Quickstart.vue
```diff
@@ -0,0 +1,696 @@
...
+          <code class="qs-code" @click="select"
+            >APPMAP=true bundle exec rake test</code
+          >
+          <button class="qs-button" v-if="!isActionRunning" @click="runAction">
+            Run tests to create AppMaps
+          </button>
+          <div
+            class="qs-loader"
+            v-if="isActionRunning"
+            data-process="5 AppMaps created"
```
dustinbyrne, line 142: We'll need to be able to provide the count through a prop.

packages/components/src/pages/Quickstart.vue
```diff
@@ -0,0 +1,762 @@
...
+            Next step : Record AppMaps ->
+          </button>
+        </div>
+      </section>
+      <section class="qs-step" v-if="currentStep === 3">
+        <div class="qs-step__head">
+          <h1 class="qs-title">Record AppMaps</h1>
+          <select class="qs-select">
+            <option>minitest (detected)</option>
+          </select>
```
dustinbyrne, line 136: The options here will also need to be provided via prop. All options will technically be detected, so we won't need that status explicitly called out.

We'll also need a way of knowing what option the user has selected, maybe it's provided via another param on `onAction`?

```ts
onAction(language: string, step: number, testFramework?: string | null | undefined): Promise<void>
```

packages/components/src/pages/Quickstart.vue
```diff
@@ -0,0 +1,775 @@
...
+              :value="lang"
+            >
+              {{ languages[lang] }}
+            </option>
+          </select>
+        </div>
+        <div class="qs-step__block" v-if="!step1Completed && !error">
+          <p>
+            This will add the following snippet to the top of your Gemfile and
+            run bundle to install the AppMap gem.
```
dustinbyrne, line 45: This text depends on the language selected, similar to the `installSnippet`.

packages/components/src/pages/Quickstart.vue
```diff
@@ -0,0 +1,775 @@
...
+      <section class="qs-step" v-if="currentStep === 2">
+        <div class="qs-step__head">
+          <h1 class="qs-title">Configure AppMap</h1>
+        </div>
+        <div class="qs-step__block" v-if="!step2Completed">
+          <p>
+            This will create an <a href="#">appmap.yml</a> config file in the
+            root directory of your project. This will tell AppMap what code to
+            record. You can run these defaults or add more packages, gems, or
+            specific functions. You can edit this file at any time.
```
dustinbyrne, line 99: This text will also need to be changed depending on the language selected, or the text needs to be made more generic @ptrdvrk

packages/components/src/pages/Quickstart.vue
```diff
@@ -0,0 +1,775 @@
...
+            </option>
+          </select>
+        </div>
+        <div class="qs-step__block" v-if="!step3Completed">
+          <p>
+            An easy way to create AppMaps is by running your tests. This will
+            run a standard command to run your tests and generate AppMap data.
+          </p>
+          <code class="qs-code" @click="select"
+            >APPMAP=true bundle exec rake test</code
```
dustinbyrne, line 146: This should be provided via prop.

packages/components/src/pages/Quickstart.vue
```diff
@@ -0,0 +1,775 @@
...
+          </button>
+        </div>
+      </section>
+      <section class="qs-step" v-if="currentStep === 2">
+        <div class="qs-step__head">
+          <h1 class="qs-title">Configure AppMap</h1>
+        </div>
+        <div class="qs-step__block" v-if="!step2Completed">
+          <p>
+            This will create an <a href="#">appmap.yml</a> config file in the
```
dustinbyrne, line 96: `appmap.yml` won't exist yet, so there will be no way to link to it. Maybe this should be `<code>` instead?

packages/components/src/pages/Quickstart.vue
```diff
@@ -0,0 +1,775 @@
...
+        await this.onAction(
+          this.selectedLanguage,
+          this.currentStep,
+          this.selectedTestFramework
+        );
+      } catch (e) {
+        console.error(e);
+      }
+
+      this.isActionRunning = false;
```
dustinbyrne, line 367: I think it makes sense now to wait for an external system to change the state. Can we display the loading indicator regardless of when `this.onAction` returns? It should only disappear once the state has changed.

packages/components/src/pages/Quickstart.vue
sashahonchar: I can raise a `viewAppmapYml` event when user click on that link and extension will open file.

packages/components/src/pages/Quickstart.vue
sashahonchar: Apparently, new languages won't be added often.
Maybe we will write texts for every existing language and swap them inside Quickstart.vue when the user change language in select?

dustinbyrne: I'm fine with that for now.

packages/components/src/pages/Quickstart.vue
```diff
@@ -0,0 +1,829 @@
...
+      default: () => {},
+    },
+    appmapYmlSnippet: {
+      type: String,
+      default: '',
+    },
+    runTestsSnippet: {
+      type: String,
+      default: '',
+    },
```
dustinbyrne, line 297: The snippet will vary for each framework sent along in the `testFrameworks` prop. Can we map test snippets to language? Otherwise the display logic for the snippet UI becomes the responsibility of the quickstart owner (the extension plugin).

dustinbyrne: e.g.
```js
{
  minitest: 'APPMAP=true bundle exec rails test',
  rspec: 'APPMAP=true bundle exec rake',
}
```

dustinbyrne:
The AppMap list overflows horizontally (see attached picture).
image

Can the name wrap instead?

Issue #293, opened 2021-07-20
Bug: "Record AppMaps" step not working as expected after tests are started.


ptrdvrk:
Issues:
 - The "Run tests" button not replaced with the counter of recorded AppMaps
 - "AppMaps recorded" message never shows up
 - The "Next step" button never shows up

https://www.loom.com/share/dc686b23095c4eecb28bd1279d84af76

PR #294, opened 2021-07-20
fix: restore 'appmapsProgress' prop to Quickstart

sashahonchar:
Fixes: #293


ptrdvrk:
This looks OK for the next UX testing build, but there are still flaws in the logic of the 3rd step.
- a new appmap event has to be sent to the current webview in order to display the actual # of AppMaps recorded
  - the "0 AppMap recorded" is not looking good
- the UI slightly different when the IDE is open the 2nd time - the Edit button of the tests is not present the 1st time

https://www.loom.com/share/98256cc3d916411ebb6428839fa404fe
@Daniel-Warner-X ^

dustinbyrne:
Stale.

PR #303, opened 2021-08-03
fix: Reword instructions for python agent install

apotterri:
These changes fix the wording and the flows for both flavors of python
agent installation.

### DIFF
--- a/packages/cli/src/agentInstaller/pythonAgentInstaller.js
+++ b/packages/cli/src/agentInstaller/pythonAgentInstaller.js
@@ -3,7 +3,6 @@
 // @ts-check
 
 const { promises: fsp } = require('fs');
-const childProcess = require('child_process');
 const BuildToolInstaller = require('./buildToolInstallerBase');
 const CommandStruct = require('./commandStruct');
 const InstallAgentStep = require('./installAgentStep');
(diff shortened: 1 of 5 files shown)


kgilpin:
Ok maybe with an asciinema I could review without having to clone and run it?

apotterri:
Sure, that's a great idea. Here you go: https://asciinema.org/a/qpF5JO6jmVaLPI5UrBD1qMP67 .

packages/cli/src/agentInstaller/pythonAgentInstaller.js
```diff
@@ -76,15 +78,18 @@ class PipInstaller extends BuildToolInstaller {
...
 
   /**
    * @returns {string}
    */
   get postInstallMessage() {
     return `The AppMap Python package ("appmap") has been added to your ${REQUIREMENTS_FILE}. You should open this file and check that
-    it looks clean and correct.`;
+it looks clean and correct.
+    
+Once you've done that, you'll complete the installation of the AppMap package by running the following command
```
kgilpin, line 91: This is the step where it didn’t sound totally clear that pushing enter was going to run the confirmation command.

packages/cli/src/agentInstaller/pythonAgentInstaller.js
```diff
@@ -35,30 +34,33 @@ class PoetryInstaller extends BuildToolInstaller {
    * @returns {string}
    */
   get assumptions() {
-    return `Your project contains a ${POETRY_LOCK_FILE}. Therefore, it looks like a poetry project,
-so we will install the AppMap Python package "appmap".`;
+    return null;
```
kgilpin, line 37: It seems more consistent to me if this step is required and states why the flow is choosing this path (because a poetry lock file exists). 

Then the install command is run and the post install message is shown. Then a confirmation command is also run to verify the install - which may be a formality, since the success of the install command is technically enough. But it’s helpful to the user, I think, to always have a confirmation step?

packages/cli/src/agentInstaller/pythonAgentInstaller.js
kgilpin: Maybe instead of “once you’ve done that”, simply “Now…”. Or maybe we should show them a diff of the file?

packages/cli/src/agentInstaller/pythonAgentInstaller.js
apotterri: If we make this step required, the user sees the confirmation message:

```
Press enter to continue, 'a' abort, or 'm' to run it yourself manually:
```

which doesn't make sense here.

I think if we want to completely rework this flow, it should probably happen in a separate PR?

kgilpin: Why doesn't it make sense?

kgilpin (approved):

packages/cli/src/agentInstaller/pythonAgentInstaller.js
kgilpin: Assumptions was designed to always be not-null, because it's explaining what path the installer has chosen - information that the user will always want to have.

apotterri: But what does "run it yourself manually" mean without an action? Isn't that going to be confusing?

I hear what you're saying about `assumptions` providing information the user needs, though. AS we move forward, I think we'll want to make the API a little more flexible so it can accommodate the requirements of each of the languages.

kgilpin: It would probably mean:

Developer: "I let you try and do this once, but you mangled my file, so I understand what we are trying to do here, and I am going to do it myself, then let you continue with the rest of your flow".

PR #307, opened 2021-08-11
Add code snippet component

dustinbyrne:
Sample here:
https://5fe18baf66f2280021e634d0-pghmhrhkfs.chromatic.com/iframe.html?id=appland-ui-code-snippet--code-snippet&args=&viewMode=story

### DIFF
--- a/packages/components/src/components/CodeSnippet.vue
+++ b/packages/components/src/components/CodeSnippet.vue
@@ -1,0 +1,46 @@
+<template>
+  <v-popper placement="top" text="Click to copy" ref="popper">
+    <pre class="code-snippet" @click="copyToClipboard" ref="code"><slot /></pre>
+  </v-popper>
+</template>
+
+<script>
+import VPopper from './Popper.vue';
+
+export default {
+  name: 'CodeSnippet',
+  components: {
+    VPopper,
+  },
+  props: {
+    clipboardText: {
+      type: String,
+      required: false,
+      default: null,
+    },
+  },
+  methods: {
+    copyToClipboard() {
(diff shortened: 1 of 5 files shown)


packages/components/src/components/CodeSnippet.vue
```diff
@@ -0,0 +1,46 @@
...
+    clipboardText: {
+      type: String,
+      required: false,
+      default: null,
+    },
+  },
+  methods: {
+    copyToClipboard() {
+      const text = this.clipboardText || this.$refs.code.innerText;
+      if (text === '' || !navigator || !navigator.clipboard) {
```
sashahonchar, line 25: shouldn't we check `clipboard.writeText` function is exists and open popper with error message for older browsers?
```suggestion
      if (text === '' || !navigator || !navigator.clipboard || !navigator.clipboard.writeText) {
        this.$refs.popper.flash('Can\'t copy to your clipboard!');
```

dustinbyrne: Yeah, good point. Right now it'll just silently fail.

It's not too important now, though, because this is guaranteed be used in a modern browser (via vscode or IntelliJ). I've logged an issue to fail gracefully if the user is on an old browser.

https://github.com/applandinc/appmap-js/issues/309

sashahonchar:
@dustinbyrne This PR should be closed, because all changes were migrated to `docs-based-flow` branch: https://github.com/applandinc/appmap-js/pull/310/commits/277cc59a984f1fb156de534d7e71696070e5f857

dustinbyrne:
Merged with #310.

PR #308, opened 2021-08-12
feat: remove non-crucial dependencies to make run on the browser.

lachrist:
BREAKING CHANGE: the main function does no longer accept a path, instead the json data must be provided as first argument.


apotterri (approved):
LGTM

PR #316, opened 2021-08-20
feat: Rework the `agent-install` user experience

dustinbyrne:
- User input prompts are now asked ahead of running any automation.
- Prompts have been cleaned up, and the overall look and feel has been improved.
- Command output has been hidden behind the `-v/--verbose` flag. Users will be instructed to re-run with this flag if an error occurs.

Tests will be failing. I've yet to update them.

https://asciinema.org/a/Ci4tclAJraPE5lZyBo1QrPNhY


apotterri:
@dustinbyrne If at all possible, can you restructure PR this so that refactoring changes and functional changes happen in separate commits, please? It will be easier to review that way.

packages/cli/src/cmds/agentInstaller/pythonAgentInstaller.ts
```diff
@@ -0,0 +1,120 @@
...
+      ['add', '--dev', '--allow-prereleases', 'appmap'],
+      this.path
+    );
+  }
+
+  initCommand(): CommandStruct {
+    return new CommandStruct('poetry', ['run', 'appmap-agent-init'], this.path);
+  }
+
+  installAgent(): void {
```
apotterri, line 53: @dustinbyrne this isn't right: there's only one step in the installation of the python agent in a poetry project (the `poetry add` command above).

This is the reason I introduced the notion of a workflow based on the installer type. Any thoughts on how you'd like to address this?

dustinbyrne: Where was the logic for the `poetry` install?

apotterri: `poertry add` updates `pyproject.toml`, and actually installs the `appmap-python` package. Nothing else needs to be done.

dustinbyrne: Ah - the verify command isn't actually run anywhere. Is `verify` the same as `validate`? It makes more sense to me that the install step is responsible for running `poetry add`.

apotterri: No, `verify` isn't the same as `validate` (i.e. the step that calls out to `appmap-agent-validate`).

FWIW, @kgilpin originally came up with these steps, based on the requirements for Ruby and Java projects. They didn't map well to poetry projects. Structure them however seems best to you, based on your new framework. As long as `poetry add` gets called in a poetry project to install the agent, things should be fine.

PR #317, opened 2021-08-24
Prototype of the `assert` CLI command to check performance assertions

torchello:


packages/cli/src/cmds/assert/assert.ts
```diff
@@ -0,0 +1,77 @@
...
+    const { appMapsDir } = argv;
+
+    if (!fs.existsSync(appMapsDir)) {
+      throw new ValidationError(`AppMaps directory ${appMapsDir} does not exist.`)
+    }
+
+    let summary = { passed: 0, failed: 0 }
+    const checker = new AssertionChecker();
+    const assertions = [
+      new Assertion('http_server_response', '', 'e.elapsed < 1'),
```
torchello, line 42: "Hardcoded" assertion for the prototype

packages/cli/src/cmds/assert/assertionChecker.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export default class AssertionChecker {
+  private strategies: Strategy[] = [
+    new HttpRequestStrategy(),
+  ];
+
+  check(appMapData: AppMapData, assertion: Assertion): Boolean {
+    for (let strategy of this.strategies) {
+      if (strategy.supports(assertion)) {
+        return strategy.check(appMapData, assertion);
```
torchello, line 15: Simple strategy pattern here to decouple different scope for the high-level checker.

packages/cli/src/cmds/assert/strategy/httpRequestStrategy.ts
```diff
@@ -0,0 +1,19 @@
...
+      if (e.hasOwnProperty('http_server_response') && !eval(assertion.assert)) {
+        return false;
+      }
+    }
+
+    return true;
+  }
+
+  supports(assertion: Assertion): Boolean {
+    return assertion.scope === 'http_server_response';
```
torchello, line 17: "Mapping strategies by `scope`"

packages/cli/src/cmds/assert/strategy/httpRequestStrategy.ts
```diff
@@ -0,0 +1,19 @@
+import Strategy from './strategy';
+import { AppMapData } from '../../../appland/types';
+import Assertion from '../assertion';
+
+export default class HttpRequestStrategy implements Strategy {
+  check(appMap: AppMapData, assertion: Assertion): Boolean {
+    for (let e of appMap.events) {
+      if (e.hasOwnProperty('http_server_response') && !eval(assertion.assert)) {
```
torchello, line 8: We also need to apply `where` condition.

torchello, line 8: Is it safe to `eval` like this? Are there better/safer ways?

dustinbyrne: Generally I think we can consider this as trusted, peer reviewed code, but take a look at the [`vm` module](https://nodejs.org/api/vm.html#vm_vm_executing_javascript). `eval` has access to everything within the current scope and can introduce side effects as a result.

packages/cli/src/cmds/assert/assert.ts
```diff
@@ -0,0 +1,77 @@
...
+import { buildAppMap } from '../../search/utils';
+import AssertionChecker from './assertionChecker';
+import Assertion from './assertion';
+
+
+exports.command = 'assert <appMapsDir>';
+exports.describe = 'Run assertions for AppMaps in the directory';
+
+exports.builder = (args) => {
+  args.positional('appMapsDir', {
```
kgilpin, line 17: Other commands use: ```
      args.option('appmap-dir', {
        default: 'tmp/appmap',
      });
```

File: packages/cli/src/cmds/assert/assert.ts [hunk 6]
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    let summary = { passed: 0, failed: 0 }
+    const checker = new AssertionChecker();
+    const assertions = [
+      new Assertion('http_server_response', '', 'e.elapsed < 1'),
+    ];
+
+    fs.readdirSync(appMapsDir).forEach((file: string) => {
+      const filePath = appMapsDir + '/' + file;
```
kgilpin, line 46: Consider `path.join`

packages/cli/src/cmds/assert/assertion.ts
```diff
@@ -0,0 +1,6 @@
+import { Scope } from './types';
+
+export default class Assertion {
+  constructor(public scope: Scope, public where: string, public assert: string) {
```
kgilpin, line 4: What does public mean on a parameter?

packages/cli/src/cmds/assert/assertionChecker.ts
kgilpin: It would be easier for me to understand the "strategy" design if there was at least one more example of a strategy.

packages/cli/src/cmds/assert/strategy/httpRequestStrategy.ts
kgilpin, line 8: `Event` has a bunch of built-in functions that you can use, for example `e.httpServerRequest` rather than digging in with `hasOwnProperty`

packages/cli/src/cmds/assert/types.d.ts
```diff
@@ -0,0 +1 @@
+export type Scope = 'appmap' | 'http_server_response' | 'http_server_request';
```
kgilpin, line 1: The `response` is always paired with the `request` so I think only `appmap` and `http_server_request` are needed here.

packages/cli/src/cmds/assert/assert.ts
torchello: Changed in 2bbe5f902fdfa10d8e47c7dd76c8f91f401aaf4c.

packages/cli/src/cmds/assert/assertion.ts
torchello: The `public` keyword instructs the TypeScript compiler that it’s OK to access the property from outside the class.

packages/cli/src/cmds/assert/strategy/httpRequestStrategy.ts
torchello: > Generally I think we can consider this as trusted, peer reviewed code

That's what I had in mind too.

> `eval` has access to everything within the current scope and can introduce side effects as a result.

Right, we need to be careful about naming in our code.

packages/cli/src/cmds/assert/assert.ts
```diff
@@ -20,28 +23,32 @@ exports.builder = (args) => {
   });
   args.option('config', {
     describe: 'path to assertions config file',
-    default: 'assertions.yml',
+    default: './assertionsConfig',
     alias: 'c',
   });
+  args.option('format', {
+    describe: 'output format (progress, pretty)',
```
kgilpin, line 30: You can use `options` to define the option values as code. Example: 

```javascript
      args.option('format', {
        describe: 'output format (text, yaml, json)',
        options: ['text', 'yaml', 'json'],
        default: 'text',
      });
```

packages/cli/src/cmds/assert/assert.ts
```diff
@@ -20,28 +23,32 @@ exports.builder = (args) => {
...
+    default: 'progress',
+    alias: 'f',
```
kgilpin, line 32: `f` is often an alias for `file` - just a thought.

packages/cli/src/cmds/assert/strategy/methodStrategy.ts
```diff
@@ -0,0 +1,12 @@
+// @ts-ignore
+import { Event } from '@appland/models';
+import Strategy from './strategy';
+import { Scope } from '../types';
+
+export default class MethodStrategy extends Strategy {
+  protected scope: Scope = 'method';
+
+  protected isEventApplicable(event: Event): Boolean {
+    return event.methodId;
```
kgilpin, line 10: If you want to select just the events which are functions, try `Event.isFunction`

packages/cli/src/cmds/assert/types.d.ts
```diff
@@ -1 +1 @@
-export type Scope = 'appmap' | 'http_server_request' | 'sql_query';
+export type Scope = 'appmap' | 'http_server_request' | 'sql_query' | 'method';
```
kgilpin, line 1: "Method" is kind of a Ruby-ism, so we generally refer to these as "functions".

packages/cli/src/cmds/assert/strategy/httpRequestStrategy.ts
```diff
@@ -0,0 +1,12 @@
+// @ts-ignore
+import { Event } from '@appland/models';
+import Strategy from './strategy';
+import { Scope } from '../types';
+
+export default class HttpRequestStrategy extends Strategy {
+  protected scope: Scope = 'http_server_request';
+
+  protected isEventApplicable(event: Event): Boolean {
+    return event.httpServerRequest && event.elapsed;
```
kgilpin, line 10: Here is `event.elapsed` again, which I think will always be null.

torchello: Is it enough to check for `event.httpServerRequest`?

packages/cli/src/cmds/assert/strategy/strategy.ts
```diff
@@ -0,0 +1,34 @@
+// @ts-ignore
+import { Event } from '@appland/models';
+import Assertion from '../assertion';
+import { AppMapData } from '../../../appland/types';
+import { Scope } from '../types';
+
+export default abstract class Strategy {
+  protected abstract scope: Scope;
+  protected abstract isEventApplicable(event: Event): Boolean;
```
dustinbyrne, line 9: Just a suggestion, but if `Strategy` and `Scope` are strongly connected, in that you're always adding a new `Scope` string when creating a new `Strategy`, this can be simplified by the type checker:

```ts
interface Strategy {
  isEventApplicable(event: Event): boolean;
}

const Strategies = {
  sql_query: {
    isEventApplicable(event: Event) {
      return Boolean(event.sql);
    },
  },
  http_server_request: {
    isEventApplicable(event: Event) {
      return Boolean(event.httpServerRequest);
    },
  },
} as const;

type Scope = keyof typeof Strategies;

export default Strategies;
```

Of course, this requires all of the the `Strategy` implementations to exist in a single object. YMMV. Something to play around with.

packages/cli/src/cmds/assert/strategy/httpRequestStrategy.ts
kgilpin: Yes. `elapsed` is actually on the return event anyway.

packages/cli/src/cmds/assert/strategy/strategy.ts
dustinbyrne, line 9: Prefer the primitive `boolean` type over the object form (`Boolean`)

packages/cli/src/cmds/assert/assert.ts
```diff
@@ -0,0 +1,105 @@
...
+import * as console from 'console';
+import { buildAppMap } from '../../search/utils';
+import AssertionChecker from './assertionChecker';
+import Assertion from './assertion';
+import { glob } from 'glob';
+import AssertionsConfig from './assertionsConfig';
+import ProgressFormatter from './formatter/progressFormatter';
+import PrettyFormatter from './formatter/prettyFormatter';
+
+exports.command = 'assert';
```
dustinbyrne, line 16: Prefer ES6 `export` instead of `module.exports`

packages/cli/src/cmds/assert/assert.ts
```diff
@@ -0,0 +1,105 @@
...
+    const assertions = (new AssertionsConfig()).assertions;
+
+    const files: string[] = await new Promise((resolve, reject) => {
+      glob(`${appmapDir}/**/*.appmap.json`, (err, globFiles) => {
+        if (err) {
+          return reject(err);
+        }
+        return resolve(globFiles);
+      });
+    });
```
dustinbyrne, line 63: Consider this:

```ts
import { glob as globCallback } from 'glob';
import { promisify } from 'util';

const glob = promisify(globCallback);

[...]

const files: string[] = await glob(`${appmapDir}/**/*.appmap.json`);
```

packages/cli/src/cmds/assert/assert.ts
```diff
@@ -0,0 +1,105 @@
...
+  return args.strict();
+};
+
+exports.handler = async (argv) => {
+  verbose(argv.verbose);
+
+  const commandFn = async () => {
+    const { appmapDir, config, format } = argv;
+
+    if (!fs.existsSync(appmapDir)) {
```
dustinbyrne, line 43: Prefer asynchronous

dustinbyrne (changes requested):
Looking really great. I've added a couple of (nit-picky) comments below.

packages/cli/src/cmds/assert/assert.ts
```diff
@@ -0,0 +1,110 @@
...
+        `AppMaps directory ${appmapDir} does not exist.`
+      );
+    }
+
+    let summary = { passed: 0, failed: 0, skipped: 0 };
+    const checker = new AssertionChecker();
+    const formatter =
+      format === 'progress' ? new ProgressFormatter() : new PrettyFormatter();
+    await import(config);
+    const assertions = new AssertionsConfig().assertions;
```
kgilpin, line 54: I think this is only working because ./assertionsConfig is imported at the top of the file.

packages/cli/src/cmds/assert/assert.ts
kgilpin: If there's anywhere it's resonable to do it, it's in setup code like this though :-)

packages/cli/src/cmds/assert/strategy/strategy.ts
torchello: Thanks for the idea. I think we may want to have a strategy that does not iterate over events, but appmaps only at some point. So I'd keep it as is to be more flexible if needed.

kgilpin (approved):

torchello:
Closed as abandoned.

PR #323, opened 2021-08-30
feat: Add validate-config subcommand

apotterri:
Add the validate-config subcommand to the CLI. Currently, only supports
Java projects, and uses a hard-coded JSON schema for the config.

Here's what it looks like: https://asciinema.org/a/gVYKc91UGuM1Yb6W8z1uhgo7C

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -42,6 +42,8 @@
   },
   "dependencies": {
     "@appland/models": "workspace:*",
+    "@sidvind/better-ajv-errors": "^0.9.1",
+    "ajv": "^8.6.2",
     "async": "^3.2.0",
     "chokidar": "^3.5.1",
     "cli-progress": "^3.9.0",
@@ -52,6 +54,7 @@
     "glob": "^7.1.6",
     "js-yaml": "^3.14.1",
     "jsdom": "^16.6.0",
+    "jsonschema": "^1.4.0",
     "moo": "^0.5.1",
     "semver": "^7.3.5",
     "w3c-xmlserializer": "^2.0.0",
--- a/packages/cli/src/cmds/configValidator/validate-config.js
+++ b/packages/cli/src/cmds/configValidator/validate-config.js
@@ -1,0 +1,73 @@
+const fs = require('fs');
+const { join } = require('path');
+const process = require('process');
+
+const yaml = require('js-yaml');
+const Yargs = require('yargs');
+
+const { verbose } = require('../../utils');
+const validator = require('../../service/config/validator');
+
+exports.command = 'validate-config <project type>';
+exports.describe = 'Validate AppMap configuration';
+
+const projectTypes = ['java', 'python', 'ruby'];
+exports.builder = (yargs) => {
+  yargs
+    .positional('project type', {
+      describe: 'the type of project to validate',
+    })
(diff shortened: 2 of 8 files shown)


packages/cli/package.json
```diff
@@ -1,22 +1,17 @@
 {
   "name": "@appland/appmap",
-  "version": "3.1.0",
+  "version": "3.0.0",
```
kgilpin, line 3: This is going down a minor version, is that expected?

packages/cli/package.json
```diff
@@ -1,22 +1,17 @@
...
   "description": "",
-  "main": "built/cli.js",
-  "bin": "built/cli.js",
+  "main": "src/cli.js",
+  "bin": "src/cli.js",
```
kgilpin, line 6: Is this a conflict with the TypeScript migration?

kgilpin (commented):
What's the longer-term plan for this command to get access to the JSON schema? Will the agent CLI have a command to print the schema to stdout?

packages/cli/package.json
apotterri: Version is managed by semantic-release, so I'm expecting this to get updated after it's merged.

packages/cli/package.json
apotterri: @dustinbyrne What should happen here?

apotterri:
I hadn't gotten that far, but having the an agent CLI subcommand to print out the schema seems like a good way to go.

packages/cli/package.json
dustinbyrne: Without looking at the `git log`, it looks like this branch is based off the commit that includes typescript and this commit rolls all those changes back.

packages/cli/package.json
dustinbyrne: The latest live release is `3.1.0`, so this change should not be made.

apotterri:
@kgilpin It occurs to me that we could include the schema in the output from the agent's `validate` subcommand. What do you think about that?

packages/cli/tests/unit/service/config/validator.spec.js
```diff
@@ -0,0 +1,33 @@
...
+  it('returns no errors for a valid config', () => {
+    const config = {
+      name: 'a-config',
+    };
+
+    const result = validateConfig(schema, config);
+    expect(result.valid).toBeTruthy();
+  });
+
+  it('returns errors for a valid config', () => {
```
kgilpin, line 24: For an invalid config?

packages/cli/src/cmds/configValidator/validate-config.js
```diff
@@ -0,0 +1,57 @@
...
+  logger.error(`\n${result.errors}\n`);
+
+  return 1;
+};
+
+exports.handler = async (argv) => {
+  verbose(argv.verbose);
+
+  const schema = yaml.load(
+    fs.readFileSync(join(__dirname, 'java-config-schema.yml'))
```
kgilpin, line 49: It seems a bit problematic to merge this if it's hard-coded to the Java config schema.

apotterri:
This is pretty far out of date now, I'll open a new PR when I get back to adding this subcommand.

PR #324, opened 2021-09-01
feat: Add validation step after install

dustinbyrne:
**The history needs cleanup here**, so this PR shouldn't be merged till that happens.

### DIFF
--- a/packages/cli/src/cmds/agentInstaller/agentInstaller.ts
+++ b/packages/cli/src/cmds/agentInstaller/agentInstaller.ts
@@ -5,10 +5,10 @@ export default interface AgentInstaller {
   readonly name: string;
   readonly buildFile?: string;
 
-  installAgent(): void | Promise<void>;
-  validateAgent?(): void | Promise<void>;
-  initCommand(): CommandStruct | Promise<CommandStruct>;
-  verifyCommand?(): CommandStruct | Promise<CommandStruct>;
-  postInstallMessage?(): string | Promise<string>;
-  available(): boolean | Promise<boolean>;
+  installAgent(): Promise<void>;
+  validateAgentCommand?(): Promise<CommandStruct>;
+  initCommand(): Promise<CommandStruct>;
+  verifyCommand?(): Promise<CommandStruct>;
+  postInstallMessage?(): Promise<string>;
+  available(): Promise<boolean>;
 }
--- a/packages/cli/src/cmds/agentInstaller/agentInstaller.ts
+++ b/packages/cli/src/cmds/agentInstaller/agentInstaller.ts
@@ -6,6 +6,7 @@ export default interface AgentInstaller {
   readonly buildFile?: string;
 
   installAgent(): void | Promise<void>;
+  validateAgent?(): void | Promise<void>;
   initCommand(): CommandStruct | Promise<CommandStruct>;
   verifyCommand?(): CommandStruct | Promise<CommandStruct>;
   postInstallMessage?(): string | Promise<string>;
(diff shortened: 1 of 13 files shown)


packages/cli/src/cmds/agentInstaller/agentInstaller.ts
```diff
@@ -5,9 +5,10 @@ export default interface AgentInstaller {
...
-  initCommand(): CommandStruct | Promise<CommandStruct>;
-  verifyCommand?(): CommandStruct | Promise<CommandStruct>;
-  postInstallMessage?(): string | Promise<string>;
-  available(): boolean | Promise<boolean>;
+  installAgent(): Promise<void>;
+  validateAgentCommand?(): Promise<CommandStruct>;
+  initCommand(): Promise<CommandStruct>;
+  verifyCommand?(): Promise<CommandStruct>;
+  postInstallMessage?(): Promise<string>;
+  available(): Promise<boolean>;
```
dustinbyrne, line 13: This is mostly a code-style decision, but I prefer the ability for an implementation not to require a Promise return value when it's not actually returning a promise. Consider:

```ts
async initCommand(): Promise<CommandStruct> {
  return new CommandStruct('poetry', ['run', 'appmap-agent-init'], this.path);
}
```

This method is neither asynchronous nor does it return a Promise. It just so happens that the JavaScript interpreter will handle `await new CommandStruct(...)` as if it were a Promise, but it's not immediately clear what's going on. This also doesn't fully abide by the Promise interface in that you cannot `initCommand().then(...)`, because `then` doesn't exist. Unfortunately neither the compiler nor the linter seems to care.

By keeping the union return type, both signatures are valid implementations of the interface and we can accurately describe what the method actually returns.
```ts
initCommand(): CommandStruct {
  return new CommandStruct('poetry', ['run', 'appmap-agent-init'], this.path);
}

async initCommand(): Promise<CommandStruct> {
  return new CommandStruct('poetry', ['run', 'appmap-agent-init', await something()], this.path);
}
```

apotterri: > This method is neither asynchronous nor does it return a Promise. It just so happens that the JavaScript interpreter will handle await new CommandStruct(...) as if it were a Promise, but it's not immediately clear what's going on. This also doesn't fully abide by the Promise interface in that you cannot initCommand().then(...), because then doesn't exist. Unfortunately neither the compiler nor the linter seems to care.

It's true that this method isn't asynchronous. It **does** return a Promise, though, courtesy of the interpreter: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function#description , which says:

>  Async functions always return a promise. If the return value of an async function is not explicitly a promise, it will be implicitly wrapped in a promise. 

So, the interface does describe what the function returns, which is (probably) why the compiler isn't complaining about it.

apotterri: FWIW, I tested this by rewriting
```javascript
    if (writeAppMapYml) {
      const initCommand = await installer.initCommand();
      const { stdout } = await run(initCommand);
      const json = JSON.parse(stdout);

      await fs.writeFile(
        join(this.path, 'appmap.yml'),
        json.configuration.contents
      );
    }
```
as
```javascript
    if (writeAppMapYml) {
      await installer.initCommand().then(async (initCommand) => {
        const { stdout } = await run(initCommand);
        const json = JSON.parse(stdout);

        await fs.writeFile(
          join(this.path, 'appmap.yml'),
          json.configuration.contents
        );
      });
    }
```
and confirming that it works for `initCommand` implemented as
```javascript
  async initCommand(): Promise<CommandStruct> {
    return new CommandStruct('appmap-agent-init', [], this.path);
  }
```

PR #328, opened 2021-09-02
feat: re-styled CodeSnippet component

### LINKED ISSUE #309: CodeSnippet component may fail to write to the clipboard
When using the `CodeSnippet` component, the popper and pointer cursor should be disabled if the user is on an old browser that does not support `navigation.clipboard.writeText`.

### LINKED ISSUE #326: Copy-to-clipboard interaction is clear to the user
The 'copy to clipboard' field in step 2 of onboarding includes an icon and clear success message.

The interaction should resemble the one GitHub uses when you copy a repo address to clone.

image

image
 
As part of this story, the border on the welcome message should be changed to white because it's not interactive.

image

sashahonchar:
image

Fixes #309 
Fixes #326


packages/components/src/components/Popper.vue
```diff
@@ -1,9 +1,11 @@
 <template>
-  <div class="popper" @mouseover="hover = true" @mouseleave="hover = false">
+  <div class="popper">
```
dustinbyrne, line 2: This breaks the value hover on nodes in the trace view. Perhaps we need a prop that controls whether or not the popper becomes active on hover or on click?

dustinbyrne (changes requested):

packages/components/src/pages/quickstart-docs/InstallAgent.vue
```diff
@@ -32,6 +35,14 @@ export default {
     VCodeSnippet,
   },
 
+  data() {
+    return {
+      codeSnippet: 'npx @appland/appmap install-agent',
+      messageSuccess:
+        '<b>Copied!</b><br>Paste into your terminal<br>to run the installer',
```
Daniel-Warner-X, line 42: In the storybook, the success message just says 'Copied to clipboard.' Is there some way I can see the full text of this success message? 
image

packages/components/src/components/CodeSnippet.vue
```diff
@@ -1,50 +1,122 @@
...
-        typeof navigator.clipboard.writeText !== 'function'
-      ) {
-        return;
-      }
+      const text = this.clipboardText.trim();
 
-      navigator.clipboard.writeText(text);
-      this.$refs.popper.flash('Copied to clipboard!');
+      if (text && this.hasClipboardAPI) {
+        navigator.clipboard.writeText(text);
```
dustinbyrne, line 58: Can you add an event emit to the root when the user copies text?

PR #333, opened 2021-09-03
fix: Add Poetry installer

dustinbyrne:
This also fixes an issue with the `pip` installer, where `pip install` was never actually run.


apotterri:
Two things I think we should consider improving very soon:

* The behavior when errors are encountered isn't very user-friendly. Having to rerun the command just to see the error output seems unnecessary. We could, instead, hide normal output and show error output.
* We should be able to do a better job of choosing default options. For example, Java, Ruby, and Python projects all use different build-config files, so we should be able pick the correct default project type.

apotterri (approved):

PR #337, opened 2021-09-09
feat: Record and open an AppMap

kgilpin:
https://www.loom.com/share/45015cd8c0304781be7ac06098e61d1e


sashahonchar (approved):
I checked it and it works with both relative and absolute paths to appmap file

apotterri (approved):
I have reservations about approving this, because it doesn't have any tests. I know it's important, though, so I'll leave it up to as to when you add some.

PR #340, opened 2021-09-09
feat: Add telemetry API

dustinbyrne:


packages/cli/src/telemetry.ts
```diff
@@ -0,0 +1,209 @@
...
+  '00:00:00:00:00:00',
+  'ff:ff:ff:ff:ff:ff',
+  'ac:de:48:00:11:22',
+]);
+
+// This key is meant to be publically shared. However, I'm adding a simple
+// obfuscation to mitigate key scraping bots on GitHub. The key is split on
+// hypens and base64 encoded without padding.
+// key.split('-').map((x) => x.toString('base64').replace(/=*/, ''))
+const INSTRUMENTATION_KEY = [
```
apotterri, line 23: Re hiding this from bots, is there any advantage to using a name that doesn't contain `KEY`?

apotterri (approved):
This LGTM. It overlaps with https://github.com/applandinc/appmap-js/pull/324, so there'll be some integration required. I'll leave it up to you as to which order you think they should be merged.

packages/cli/src/telemetry.ts
dustinbyrne: Maybe, but I think this is already overkill. I've seen a number of repositories where they just paste the key in plain text, Microsoft included.

PR #361, opened 2021-09-17
feat: Auto-select installer

### LINKED ISSUE #356: `install-agent` is renamed to `install` and has a hidden alias for `install-agent`

### LINKED ISSUE #363: Auto-select installer if only one is available
Otherwise, if two or more installers are available, allow the user to pick one before presenting the confirmation prompt.

dustinbyrne:
- If only one installer is available, select it and prompt the user to confirm.
- If more than one installer is available, prompt the user to select one before confirming.
- If no installers are present, inform the user of the supported project types and exit.
- Overwrite `appmap.yml` now comes after the installation confirmation prompt.
- Remove broad language installer prompt and arg line options.
- Rename `install-agent` to `install`, with hidden `install-agent`, `i`, aliases.

Resolves #363
Resolves #356

Example videos:
- [Installing into a maven directory](https://asciinema.org/a/fT3AyS5ay8JnmsI1Psbub2xJ6)
- [Installing into a directory with more than one project type](https://asciinema.org/a/0vZ2l1NJ7xajMu7setdYJepQw)
- [Installing into a directory with no project](https://asciinema.org/a/3DIIPmT1mcvSxNAkErrsjdBk8)


packages/cli/tests/unit/agentInstall/command.spec.ts
```diff
@@ -18,7 +18,6 @@ tmp.setGracefulCleanup();
 
 const invokeCommand = (
   projectDir: string,
-  projectType: string,
```
dustinbyrne, line 21: We don't need this anymore because the installer will automatically identify the correct project type.

apotterri (approved):
There's quite a bit of new functionality here, but no tests for most of it. You should consider adding some.

kgilpin:
@dustinbyrne since we have fixtures for the installable projects (right?) it would probably be time well spent to take an hour or so to add some tests.

kgilpin:
@apotterri could you suggest how some tests might work and take advantage of the fixtures we have?

dustinbyrne:
It was my feeling that this change did not need much additional testing, as majority of the changes focus on changing the order and content of the prompts instead of the install logic. That's pretty well covered in [`/tests/unit/agentInstall/command.spec.ts`](https://github.com/applandinc/appmap-js/blob/main/packages/cli/tests/unit/agentInstall/command.spec.ts). Those specs do rely now on auto-selection of the installer.

apotterri:
These two seem like changes to the install logic, no?
> If more than one installer is available, prompt the user to select one before confirming.
If no installers are present, inform the user of the supported project types and exit.

A new fixture directory with both a `requirements.txt` and a `poetry.lock` could be used to test the first one. An empty directory could be used to test the second.

PR #373, opened 2021-09-22
fix: dump log to stdout when command failed with error

### LINKED ISSUE #353: If an error occurs while running a command, the output log is dumped to stderr
Instead of requiring the user to re-run with `-v`, dump the log out at the end of the process if it encountered an error during `exec`.

sashahonchar:
Fixes #353

### DIFF
--- a/packages/cli/src/cmds/runCommand.ts
+++ b/packages/cli/src/cmds/runCommand.ts
@@ -14,16 +14,8 @@ export default async function runCommand(fn: () => Promise<any>) {
       return Yargs.exit(2, err);
     }
 
-    if (verbose()) {
-      UI.error(err);
-    } else {
-      UI.error(`An error occurred: ${err.toString().split('\n')[0]}`);
-      UI.error(
-        `Try re-running the command with the ${chalk.red(
-          'verbose'
-        )} flag (${chalk.red('-v')}).`
-      );
-    }
+    UI.error(err);
+
     Yargs.exit(3, err);
   });
 }


apotterri (changes requested):
This doesn't work for me: https://asciinema.org/a/lQMBulEetUJ38VpH9y76vlyEe . The message
```
ERROR: Could not find an activated virtualenv (required).                                                               
```
doesn't show up.

I think it's also important to split this into mulitple steps, the way @kgilpin recommended. Show the user a message, including the first few lines of the command output, and ask them if they'd like to see the rest of the message. Only if they say yes should the whole error get dumped to the screen.

A test to show that this works would be great, too.

dustinbyrne:
Updated output: https://asciinema.org/a/E5Vd3Of44zUtmi3U2PhWBRuPh

packages/cli/tests/unit/agentInstall/commandRunner.spec.ts
```diff
@@ -0,0 +1,18 @@
...
+  describe('run', () => {
+    it('raises an error when a command is not found', async () => {
+      const unknownCmd = randomBytes(8).toString('hex');
+      try {
+        await run(new CommandStruct(unknownCmd, [], process.cwd()));
+      } catch (err) {
+        expect(err.message).toMatch(`${unknownCmd} was not found`);
+        return;
+      }
+      fail('Expected an error to be thrown');
```
apotterri, line 15: You can clean up the end of this test a bit by putting an `expect.assertions(1)` at the beginning of the function: https://jestjs.io/docs/expect#expectassertionsnumber and getting rid of the `return` and `fail`.

Personally, I prefer Jest's `toThrow` or `rejects` assertions for testing error behavior, but that's mostly a style thing.

packages/cli/src/utils.js
```diff
@@ -168,6 +168,16 @@ function exists(path) {
...
 }
 
+/**
+ * Append a prefix to each line in a string
+ * @param {string} str the string to be prefixed
+ * @param {string} prefix a string to prefix each line with
+ * @returns {string} the resulting string which starts each line with a prefix
+ */
+function prefixLines(str, prefix) {
+  return str.replaceAll(/^/gm, prefix);
```
apotterri, line 178: `replaceAll` was added in Node 15. Probably just want to use `replace` here.

apotterri (changes requested):
This seems ok if the error message is only a couple of lines long. If it's more than that, though, it looks like this: https://asciinema.org/a/fgqXId7S5QbtS59JG4Kf6J4NY .

I think it would be better to let the user know that an error has occurred, and then ask them if they want to see the error output.

dustinbyrne:
Thanks for the feedback. I've addressed the above and updated the flow: https://asciinema.org/a/Rxv6KfgSfyKZQCm3l1JBB64Vo

apotterri (approved):
That looks great. Thanks for making the changes.

PR #385, opened 2021-09-28
Capture signals on exit

### LINKED ISSUE #378: Capture exit cases not originating from an exception

dustinbyrne:
Capture information when the user closes the process via `SIGINT`/`SIGTERM`/`SIGBREAK`.

Resolves #378


dustinbyrne:
There are more technical hurdles than this feature is worth at the moment. We can revisit this at a later date.

PR #390, opened 2021-09-30
feat: apply 'Set as root' context menu action globally

sashahonchar:
image

image

### DIFF
--- a/packages/components/src/pages/VsCodeExtension.vue
+++ b/packages/components/src/pages/VsCodeExtension.vue
@@ -232,6 +232,7 @@ export default {
       versionText: '',
       VIEW_COMPONENT,
       VIEW_FLOW,
+      rootCodeObject: null,
       filters: {
         limitRootEvents: {
           on: true,
@@ -307,6 +308,17 @@ export default {
         rootEvent.traverse((e) => callTree.push(e));
         return callTree;
       }, []);
+
+      if (this.rootCodeObject) {
+        const eventBranches = this.rootCodeObject.allEvents.map((e) => [
+          e.id,
+          e.linkedEvent.id,
+        ]);
+
+        events = events.filter((e) =>
+          eventBranches.some((branch) => e.id >= branch[0] && e.id <= branch[1])
+        );
+      }
 
       events = events.filter(
         (e) =>
@@ -521,6 +533,7 @@ export default {
     },
 
     resetDiagram() {
+      this.rootCodeObject = null;
       this.clearSelection();
 
       this.renderKey += 1;
@@ -563,6 +576,12 @@ export default {
     setUserFiltered() {
       this.isUserFiltered = true;
     },
+  },
+
+  mounted() {
+    this.$root.$on('makeRoot', (codeObject) => {
+      this.rootCodeObject = codeObject;
+    });
   },
 };
 </script>
--- a/packages/components/src/pages/VsCodeExtension.vue
+++ b/packages/components/src/pages/VsCodeExtension.vue
@@ -318,6 +318,8 @@ export default {
         events = events.filter((e) =>
           eventBranches.some((branch) => e.id >= branch[0] && e.id <= branch[1])
         );
+
+        appMap.classMap.setTemporaryRoot(this.rootCodeObject.id);
       }
 
       events = events.filter(
@@ -535,6 +537,7 @@ export default {
     resetDiagram() {
       this.rootCodeObject = null;
       this.clearSelection();
+      this.$store.state.appMap.classMap.clearTemporaryRoot();
 
       this.renderKey += 1;
     },
@@ -580,7 +583,9 @@ export default {
 
   mounted() {
     this.$root.$on('makeRoot', (codeObject) => {
+      this.clearSelection();
       this.rootCodeObject = codeObject;
+      console.log(codeObject);
     });
   },
 };
--- a/packages/components/src/pages/VsCodeExtension.vue
+++ b/packages/components/src/pages/VsCodeExtension.vue
@@ -298,7 +298,7 @@ export default {
     },
     filteredAppMap() {
       const { appMap } = this.$store.state;
-      let classMap = appMap.classMap;
+      let { classMap } = appMap;
       let rootEvents = appMap.rootEvents();
 
       if (this.filters.limitRootEvents.on) {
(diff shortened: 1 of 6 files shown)


dustinbyrne:
The events in the left panel don't appear to be reactive to the root changing. In the video below after changing the root node the left panel still lists events which have been filtered out. Only upon deselecting the node and reselecting it does it show the correct data.

https://user-images.githubusercontent.com/8737782/135514289-9dd365e7-04c8-48cb-bf05-90d8557bcdcb.mp4

dustinbyrne:
It looks like classes within a package aren't being filtered out entirely. This is demonstrable when expanding a package. In the picture below, `Spree::Admin::OrdersController` has no relation to the root node.
image

dustinbyrne (approved):
This looks good and I'm approving, though I left some comments above. Would you mind looking into these issues? If they're non-trivial bugs, please feel free to move them into separate issues and this can be merged as-is.

packages/models/src/classMap.js
```diff
@@ -89,6 +90,14 @@ export default class ClassMap {
...
   }
 
+  setTemporaryRoot(id) {
+    this.roots = [this.codeObjectsById[id]];
+  }
+
+  clearTemporaryRoot() {
+    this.roots = [...this.originalRoots];
+  }
+
```
dustinbyrne, line 100: I don't think `models` is the right place for this. This feels like render state to me, and we should limit that as best we can to the `components` package. Is there a better place for this?

packages/models/src/classMap.js
```diff
@@ -89,6 +90,14 @@ export default class ClassMap {
     return this.root(CodeObjectType.DATABASE);
   }
 
+  setTemporaryRoot(id) {
+    this.roots = [this.codeObjectsById[id]];
```
dustinbyrne, line 94: Consider the possibility of multiple root node filters.

dustinbyrne: This is slightly different from the original implementation, where the package is not included in the root.
For example, if I expand this node:
image

The name of the node changes in the resulting graph as a result of excluding the package:
image

While it's a somewhat minor cosmetic bug, my concern is that it is unexpected ("why did the name of the node change?") and may confuse the user.

packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -580,7 +583,9 @@ export default {
 
   mounted() {
     this.$root.$on('makeRoot', (codeObject) => {
+      this.clearSelection();
```
dustinbyrne, line 586: De-selecting the active node seems like an acceptable solution to the reactivity in the left sidebar. Is it a difficult fix to make the sidebar reactive?

dustinbyrne (changes requested):
Hi @sashahonchar, thanks for addressing the feedback. I've made a few more comments below on the updated build.

First, packages are disappearing once setting a node as root. They're disappearing from both the diagram and the left side panel. When I select `Set as root`, I expect only that existing nodes are hidden, not that the remaining nodes may be changed (i.e. packages hidden and expanded).

https://user-images.githubusercontent.com/8737782/135661184-067362bf-4df3-4696-88fd-4e7593390df1.mp4

packages/components/src/pages/VsCodeExtension.vue
sashahonchar: To make it reactive `selectedObject` should be changed, but after setting different root `classMap` is changed and it's not possible to re-select active `codeObject`, because ID of object has changed.

For example, if you select `ActiveSupport::SecurityUtils` node, ID of selected object is:
`class:active_support/ActiveSupport::SecurityUtils`
but when user set that object as root ID of that object becomes `SecurityUtils`

dustinbyrne: The issue seems that the panel takes an `AppMap` object as a prop. So that prop will be reactive (i.e., if the AppMap changes the panel will update), however changes to its children will not update the panel. The fix would be to switch the prop to a `CodeObject[]`. Sounds like this could be its own issue.

sashahonchar: Selected object info (inbound/outbound connections, events, etc.) is taken from `selectedObject`, but it doesn't change when `AppMap` changes.
https://github.com/applandinc/appmap-js/blob/feat/global-set-as-root/packages/components/src/components/DetailsPanel.vue#L28-L32

sashahonchar: And we can't update `selectedObject` by hand after appmap changes, because object's IDs were changed. @dustinbyrne

packages/models/src/classMap.js
sashahonchar: You're right, I removed that code modifications from `models` package.

packages/models/src/classMap.js
sashahonchar: I think we are going to implement multiple root node filters with new `Filters` panel.
I assume in this PR we focus on context menu button "Set as root" which sets new root object every time it gets clicked.
(Bug with name change was fixed)

dustinbyrne (changes requested):
Thanks @sashahonchar, everything looks to be working great. At this point, it's functionally ready to merge, but a couple of tests to assert this behavior would be ideal. It doesn't look like our existing tests cover set as root at all.

There's one bug still outstanding, but we can address it as a follow-on item at a later date: https://github.com/applandinc/appmap-js/issues/396

packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -308,6 +310,28 @@ export default {
...
+          this.rootCodeObject.type === 'class' &&
+          this.rootCodeObject.packageObject
+        ) {
+          classMap = new ClassMap([
+            classMap.codeObjects.filter(
+              (co) => co.id !== this.rootCodeObject.packageObject.id
+            ),
+          ]);
+        }
+      }
```
dustinbyrne, line 333: Would you mind adding a comment to this block explaining it's purpose, similar to what you described in Slack?
```
It's necessary to remove package wrap in Component Diagram, without this code package will be rendered in collapsed state
```

sashahonchar: Comment was added https://github.com/applandinc/appmap-js/pull/390/commits/80ad5d783b816686954a494d3fba42f28eebb5af

PR #393, opened 2021-10-05
fix: Support UTF-16LE encoded build files

apotterri:
For some locales, the build config files may be UTFi-16LE encoded.
Detect this, and make sure the config files are updated appropriately.

Also, use the platform's line endings when updating build config files.


packages/cli/src/encodedFile.ts
```diff
@@ -0,0 +1,77 @@
...
+    buf: Buffer,
+    bom0: number | undefined,
+    bom1: number | undefined,
+    encoding: TranscodeEncoding | undefined
+  } {
+    const buf = fs.readFileSync(path);
+    let bom0: number | undefined = buf.readUInt8(0);
+    let bom1: number | undefined = buf.readUInt8(1);
+    let encoding: TranscodeEncoding | undefined;
+    if (bom0 == 0xff && bom1 == 0xfe) {
```
kgilpin, line 57: Did you want `===` here?

apotterri: Yes. 👍

packages/cli/src/cmds/agentInstaller/gradleInstaller.ts
```diff
@@ -111,7 +111,7 @@ export default class GradleInstaller
     const lines = parseResult.plugins
       ? buildFileSource
           .substring(parseResult.plugins.lbrace + 1, parseResult.plugins.rbrace)
-          .split('\n')
+          .split(os.EOL)
```
dividedmind, line 114: I don't think that's right. Config files can have Unix line breaks even on Windows. We should handle this case and make sure we leave the existing line breaks as we found them, if only to limit diff noise. (I think it's acceptable, if a bit fugly, to insert any new ones as `os.EOL`.)

packages/cli/src/encodedFile.ts
```diff
@@ -0,0 +1,77 @@
...
+    bom0: number | undefined,
+    bom1: number | undefined,
+    encoding: TranscodeEncoding | undefined
+  } {
+    const buf = fs.readFileSync(path);
+    let bom0: number | undefined = buf.readUInt8(0);
+    let bom1: number | undefined = buf.readUInt8(1);
+    let encoding: TranscodeEncoding | undefined;
+    if (bom0 === 0xff && bom1 === 0xfe) {
+      encoding = 'utf16le';
```
dividedmind, line 58: What about other BOMs? I suppose they'll be rarely seen in the wild, so it's probably OK to throw an error, EXCEPT for the UTF-8 one, which Microsoft really likes to use. Note UTF-8 BOM is three bytes long, so it probably makes more sense to store it as an array.

packages/cli/src/encodedFile.ts
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  // Encode the given string with our encoding and write it to path
+  // @param {string} str the string to write
+  // @param {PathLike} [path] the path to write the string to, defaults to this.path
+  write(str: string, path?: fs.PathLike) {
+    let buf = Buffer.from(str);
+
+    if (this.encoding) {
+      buf = buffer.transcode(buf, 'utf8', this.encoding);
```
dividedmind, line 37: Wouldn't it make more sense to provide the encoding to `Buffer.from` instead of immediately transcoding it?

packages/cli/tests/unit/encodedFile.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+  
+  it('writes a 7-bit ASCII file', () => {
+    const expectedPath = join(fixtureDir, 'ascii.txt');
+    const ef = new EncodedFile(expectedPath);
+    const actualPath = join(tmpDir, 'ascii.actual.txt');
+    ef.write(ef.toString(), actualPath);
+    const expected = fs.readFileSync(expectedPath);
+    const actual = fs.readFileSync(actualPath);
+    expect(actual).toEqual(expected);
+  });
```
dividedmind, line 59: You need to add some more cases here:
- UTF-8 with BOM,
- UTF-8 with Windows line endings,
- UTF-16 with Unix line endings.

dividedmind (changes requested):
Plain text can be surprisingly complicated :) This needs some more work if we don't want a fix for one bug to introduce three more.

packages/cli/src/cmds/agentInstaller/gradleInstaller.ts
apotterri: Yeah, this is wrong. Thanks.

packages/cli/src/encodedFile.ts
apotterri: So far, we've seen 11 failures occur because we've inserted text before the BOM. All them have been UTF-16LE files. I'll add a test that expects to fail for UTF-8, but it's not clear to me that it makes sense to invest beyond that.

packages/cli/src/encodedFile.ts
apotterri: > My reading of the doc for Buffer.from is that the encoding option describes what's in the source string, not the desired encoding. So, the transcoding is necessary to get to the target encoding.

I've determined that I was mistaken, and you were right: the `encoding` passed to `Buffer.from` describes the encoding for the bytes that wind up in `buf`. I'll get rid of the call to `transcode`.

UTF-16 characters in the test files come through correctly, so I don't think there are any other changes that need to be made.

packages/cli/src/encodedFile.ts
dividedmind: UTF-8 BOM wouldn't have caused a failure before -- the file would still be readable as UTF-8, except by some Microsoft tools (notably notepad) which would have assumed local 8-bit encoding and garbled non-ASCII characters (if any); it would have a BOM in the middle, but most tools would just silently ignore it (notepad in 8-bit mode would display it as mojibake). But now it WILL cause a hard failure, so I don't think that's good enough.

apotterri: I believe it would have caused a failure before, analogous to the UTF-16 failures.

Given a requirements.txt that looks like this:
```
ajp@Alans-MacBook-Pro windows % od -c -tx1 requirements-utf-8.txt
0000000  357 273 277   #       U   T   F   -   8       r   e   q   u   i
           ef  bb  bf  23  20  55  54  46  2d  38  20  72  65  71  75  69
0000020    r   e   m   e   n   t   s   .   t   x   t
           72  65  6d  65  6e  74  73  2e  74  78  74
0000033
```
we would have updated it to
```
ajp@Alans-MacBook-Pro windows % od -c -tx1 updated-requirements-utf-8.txt
0000000    a   p   p   m   a   p   >   =   1   .   1   .   0   .   d   e
           61  70  70  6d  61  70  3e  3d  31  2e  31  2e  30  2e  64  65
0000020    v   0  \n 357 273 277   #       U   T   F   -   8       r   e
           76  30  0a  ef  bb  bf  23  20  55  54  46  2d  38  20  72  65
0000040    q   u   i   r   e   m   e   n   t   s   .   t   x   t
           71  75  69  72  65  6d  65  6e  74  73  2e  74  78  74
0000056
```
with the BOM in the middle.

Other tools may have been ok with that, but pip would have failed:
```
(.venv) PS Z:\src\applandinc\appmap-js\packages\cli\tests\unit\fixtures\encodedFile\windows> pip install -r .\updated-requirements-utf-8.txt
ERROR: Invalid requirement: 'ï»¿# UTF-8 requirements.txt' (from line 2 of .\updated-requirements-utf-8.txt)
WARNING: You are using pip version 21.0.1; however, version 21.2.4 is available.
You should consider upgrading via the 'C:\Users\ajp\tmp\chipy.org\.venv\Scripts\python.exe -m pip install --upgrade pip' command.
(.venv) PS Z:\src\applandinc\appmap-js\packages\cli\tests\unit\fixtures\encodedFile\windows>
```

dividedmind (approved):

PR #417, opened 2021-10-14
fix: Correct search for first gem in a Gemfile

apotterri:
Fix the regex the Ruby installers uses to find the first gem, group, or
require in a Gemfile. Now, it only matches those statements if they're
at the beginning of the line.

Also, adds a test to check that the Gemfile is properly updated.

### DIFF
--- a/packages/cli/src/cmds/agentInstaller/rubyAgentInstaller.ts
+++ b/packages/cli/src/cmds/agentInstaller/rubyAgentInstaller.ts
@@ -11,7 +11,8 @@ import { getOutput } from './commandUtil';
 import CommandStruct from './commandStruct';
 import EncodedFile from '../../encodedFile';
 
-const REGEX_GEM_DECLARATION = /(?!\s)(?:gem|group|require)\s/m;
+const REGEX_GEM_DECLARATION = /^(?:gem|group|require)\s/m;
+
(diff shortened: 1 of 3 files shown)


packages/cli/src/cmds/agentInstaller/rubyAgentInstaller.ts
```diff
@@ -11,7 +11,8 @@ import { getOutput } from './commandUtil';
 import CommandStruct from './commandStruct';
 import EncodedFile from '../../encodedFile';
 
-const REGEX_GEM_DECLARATION = /(?!\s)(?:gem|group|require)\s/m;
+const REGEX_GEM_DECLARATION = /^(?:gem|group|require)\s/m;
```
apotterri, line 14: @dustinbyrne @kgilpin What was the intent of this regex? To match the group when not preceded by a space, it should have been `(?:(?<!\s)|^)(?:gem|group|require)`. But, it seems like only matching when the group is at the beginning of the line is what we want.

Is there a case that this will miss?

dustinbyrne: I don't think the `(?!\s)` is important. At least, I can't think of a case where it makes a difference. The new pattern is an definitely an improvement worth getting in now.

PR #429, opened 2021-10-19
New global filters

### LINKED ISSUE #396: Selection is retained in left sidebar after `Set as root`
The left sidebar should be cleared if the current selection exists outside of the new graph.

https://user-images.githubusercontent.com/8737782/136071902-cc1572b4-8d0d-4d43-91d8-9bb50040b4f6.mp4

### LINKED ISSUE #399: `Set as root` can be performed from the left sidebar
A button exists which adds the current code object as a root code object in the class map.

__WIP__ (from [Figma](https://www.figma.com/file/T72cGO1asVVPg40ZioWwXa/custom-views?node-id=32%3A622))
image

### LINKED ISSUE #418: Map UI controls are updated
- Dependency & Trace map tabs matches designs
- Left-column search bar is moved into Dependency Map tab. This will search for code objects 
- Trace tab search bar for events in Trace view.
- Filter icon (top right of tab) toggles filters
- searchbar removed from left column (moving to dependency tab)

[Figma ->](https://www.figma.com/file/T72cGO1asVVPg40ZioWwXa/custom-views?node-id=242%3A5554)

custom_views_trace

### LINKED ISSUE #419: Users can cycle through search results
When users perform a search in either the Trace, or Dependency maps, their results will be highlighted in the map. 

If there are multiple results, users are able to cycle through these... the map will change the highlighted item, a count badge will be displayed, and the left column will display the details for the highlighted item
 
[Figma ->](https://www.figma.com/proto/T72cGO1asVVPg40ZioWwXa/custom-views?page-id=144%3A4603&node-id=144%3A7831&viewport=363%2C48%2C0.74&scaling=min-zoom&starting-point-node-id=144%3A5102)

search-results
.

### LINKED ISSUE #420: Code object displays controls for filtering
Each code object will have a filters component in it's details panel (left column). 

- Filters: 
-- Set as root
-- Hide
- Only one filter can be applied at a time. The other filter will become inactive when one is selected.
- Resetting filters deselect all options, making both available to the user

Screen Shot 2021-10-14 at 4 20 11 PM

### LINKED ISSUE #421: Clicking filter icon displays filters
Filter popup/modal is displayed when user clicks the filter icon. The filters are broken up into 2 sections: Root & Declutter. 

Users are able to add new items to these sections by using the *add new* field. New filter items are displayed below the field after the user clicks the + icon (or enter). Clicking the X in the displayed term removes this from the filters.

Clicking *Reset all* removes all filters, and the map is reset to reflect this.

The input fields have autocorrect for objects found within the AppMap.

[Figma ->](https://www.figma.com/file/T72cGO1asVVPg40ZioWwXa/custom-views?node-id=242%3A9386)

Screen Shot 2021-10-14 at 5 07 15 PM

Screen Shot 2021-10-14 at 5 14 43 PM

### LINKED ISSUE #453: Diagrams filter out media requests for blacklisted file extensions
See the attached AppMap as an example, 44 out of 41 requests are for images, css, .js files etc.:
[test.appmap.json.txt](https://github.com/applandinc/appmap-js/files/7523588/test.appmap.json.txt)
 

When the "Hide HTTP media requests" filter is on, all requests for these file extensions should be hidden:
- bitmap images (.jpg, .png, .jpeg, .gif, .ico, ...)
- vector images (.svg, ...)
- .css files
- .js files
- font files (.woff, .otf, .tff, ...)
- video files (.mp4, .mov, .flv, .avi, ...)
- .html files

sashahonchar:
image

Fixes #396 
Fixes #399 
Fixes #418 
Fixes #419 
Fixes #420 
Fixes #421 
Fixes #453


packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -72,40 +98,160 @@
...
+                        placeholder="add new root..."
+                      />
+                      <PlusIcon class="filters__form-plus" />
+                    </form>
+                  </div>
+                  <div class="filters__block-body filters__block-body--flex">
+                    <div
+                      class="filters__root"
+                      v-for="co in filters.rootObjects"
+                      :key="co.id"
```
dustinbyrne, line 131: The filters applied won't always map to a `CodeObject` `id`. For example, `package:app/*`. So there needs to be a separate list of filters from which the `rootObjects` array is populated.

packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -370,6 +536,43 @@ export default {
...
+        this.filters.declutter.hideName.names.forEach((fqid) => {
+          const codeObject = classMap.codeObjects.filter(
+            (co) => co.fqid === fqid
+          )[0];
+          if (codeObject) {
+            events = events.filter((e) => !codeObject.allEvents.includes(e));
+          }
+          classMap.codeObjects = classMap.codeObjects.filter(
+            (co) => co !== codeObject
+          );
```
dustinbyrne, line 572: You should be able to always filter `codeObjects` at the end of applying event filters. It may be easiest to create a new ClassMap by iterating through the events and merging `[event.codeObject, event.codeObject.ancestors()]` into the code object tree.

packages/models/src/appMapBuilder/eventStack.js
```diff
@@ -45,10 +45,6 @@ export default class EventStack {
           parent.children.push(call);
           call.parent = parent;
         }
-      } else {
-        throw new Error(
-          `return #${event.id} is missing call #${event.parent_id}`
-        );
```
dustinbyrne, line 51: Is this occurring?

dustinbyrne:
Thanks @sashahonchar. This is looking good. I'm going to add some issues below.

dustinbyrne:
Submitting an empty root clears everything and the only way to recover is by reloading the AppMap.

https://user-images.githubusercontent.com/8737782/141496708-4e37aaeb-46ff-48fc-a2dd-2d5cc0fcf348.mp4

dustinbyrne:
Similarly, the `Hide names` input allows empty submissions.

https://user-images.githubusercontent.com/8737782/141497043-9fa840e2-adde-429f-b96e-a390dac8451d.mp4

dustinbyrne:
I'm not getting any autocomplete recommendations from the inputs in the filters box.

https://user-images.githubusercontent.com/8737782/141497211-4cd18524-f923-4a8e-bfc9-4e3443dbbc5a.mp4

packages/models/src/appMapBuilder/eventStack.js
sashahonchar: @dustinbyrne  We need to remove this exception because when we filter our events that exception blocks the whole script runtime.
We filter out only calls from events:
https://github.com/applandinc/appmap-js/blob/25ab8e0e451f93e469850a06eafb6a256620fce6/packages/components/src/pages/VsCodeExtension.vue#L538

dustinbyrne: That would indicate an issue in the filtering though, wouldn't it? This exception is thrown when a return event has no matching call. If the call is being filtered out, the return should be filtered out as well.

packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -572,8 +580,28 @@ export default {
         });
       }
 
+      let eventIds = [];
+      if (this.traceFilterValue) {
+        const filterValue = this.traceFilterValue
+          .replaceAll(/\s/g, '')
```
dustinbyrne, line 586: Prefer `replace(/\s+/g, '')` over replaceAll here - it has higher browser support coverage.

sashahonchar:
> Do you have a source on where this list came from? It would be good to know so we can updated it from time to time.

I copied some extensions from list of popular mime types from [MDN article](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types).

packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -52,15 +53,42 @@
...
+                <div class="trace-view__search-arrow">
+                  <ArrowSearchLeftIcon />
+                </div>
+                <div class="trace-view__search-arrows-text">
+                  <b>1</b>/3 results
+                </div>
+                <div class="trace-view__search-arrow">
+                  <ArrowSearchRightIcon />
+                </div>
+              </div>
```
dustinbyrne, line 80: This is not hooked up

packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -312,66 +497,88 @@ export default {
...
+      if (this.traceFilterValue) {
+        const filterValue = this.traceFilterValue.trim().split(' ');
+        if (filterValue.length) {
+          filterValue.forEach((item) => {
+            if (item.startsWith('event:')) {
+              eventIds.push(parseInt(item.replace('event:', ''), 10));
+            }
+          });
+        }
+      } else {
```
dustinbyrne, line 570: This should be highlighting events in the trace, not filtering out non-matching events. See the animation here: https://github.com/applandinc/appmap-js/issues/419

dustinbyrne:
Looks good. Two remaining items:
- [ ] Tests for the event search box and cycling through events
- [ ] The event search box can be prepopulated by an outside caller. Maybe this is possible from `VsCodeExtension.setState`?

packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -592,11 +834,160 @@ export default {
...
+        'webp',
+        'woff',
+        'woff2',
+        'xhtml',
+        '3gp',
+        '3g2',
+      ];
+
+      events.forEach((e) => {
+        const eventName = e.codeObject.prettyName;
```
dustinbyrne, line 944: It may be more clear to get the request information directly from the event instead of parsing the name.
```ts
const { requestMethod, requestPath } = event;
if (requestMethod === 'GET' && mediaFileExtensions.some((ext) => requestPath.endsWith(ext)) {

}
```

packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -312,71 +489,122 @@ export default {
...
       }
 
+      const eventIds = events.filter((e) => e.isCall()).map((e) => e.id);
+
       return buildAppMap({
-        events,
+        events: events.filter(
+          (e) =>
+            (e.isCall() && eventIds.includes(e.id)) ||
+            eventIds.includes(e.parent_id)
```
dustinbyrne, line 558: Worst case complexity using an array for `eventIds` here is O(2n^2).

`eventIds` should be a `Set`.

packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -312,71 +489,120 @@ export default {
...
+        ),
         classMap: classMap.roots.map((c) => ({ ...c.data })),
```
dustinbyrne, line 560: How is this working if we're not filtering the class map?

packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -312,71 +489,120 @@ export default {
...
+          const filterRegExp = new RegExp(id, 'ig');
+          classMap.codeObjects.forEach((codeObject) => {
+            if (!filterRegExp.test(codeObject.fqid)) {
+              return;
+            }
+
+            const eventBranches = codeObject.allEvents.map((e) => [
+              e.id,
+              e.linkedEvent.id,
+            ]);
```
dustinbyrne, line 503: Make sure `CodeObject.allEvents` is capturing everything we need. Will a package contain all the events of its ancestors? What about a class that contains nested classes (Such as `app/controllers/Spree` in the demo data). If `allEvents` is empty, it seems like we should skip the filter, because we don't want to end up with no data displayed.

packages/components/src/pages/VsCodeExtension.vue
sashahonchar: Class map objects without linked events won't be rendered in Dependency Map.

packages/components/src/pages/VsCodeExtension.vue
sashahonchar: Yes, `CodeObject.allEvents` contains all object events and all object children events.
https://github.com/applandinc/appmap-js/blob/main/packages/models/src/codeObject.js#L135-L137

packages/components/src/pages/VsCodeExtension.vue
sashahonchar: Class map objects without linked events won't be rendered in Dependency Map.

packages/components/src/pages/VsCodeExtension.vue
dustinbyrne: We're not checking `mime_type` in this block

packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -592,11 +834,160 @@ export default {
...
+        const eventName = e.codeObject.prettyName;
+
+        if (
+          eventName.startsWith('GET') &&
+          mediaFileExtensions.some((ext) => eventName.endsWith(ext))
+        ) {
+          excludedEvents.push(e.id);
+        } else if (
+          e.http_server_response &&
+          e.http_server_response.mime_type &&
```
dustinbyrne, line 946: Also check the `Content-Type` header if `mime_type` is undefined. `httpServerResponse.headers` may be undefined and the keys may vary in casing. Something like this should work (but, beware, I've not actually run this code):

```js
const [, contentType] = Object.entries(e.httpServerResponse.headers || {}).find(([k,v]) => k.toLowerCase() === 'content-type')
```

dustinbyrne:
Copying the content from the post I accidentally put in another PR:

The filters overflow if the viewport is not vertically big enough:
image

Some root node filters may result in no data:

https://user-images.githubusercontent.com/8737782/142919635-334779e1-ede9-471f-8a72-7fda18c87682.mp4

Some root node filters appear to break the diagram and display no events in the trace:

https://user-images.githubusercontent.com/8737782/142919689-d9841571-7f04-4b7d-88db-41a15a5feca6.mp4

I suggest adding tests for the following edge cases. Once they're fixed, try and find a few more and see if you can break it.
- A root node filter added for `package` renders as expected and contains the expected code objects and events
- A root node filter added for `class:app/controllers/Spree` renders as expected and contains the expected code objects and trace events
- A root node filter added for a route renders as expected and contains the expected code objects and events, with `Limit root events to HTTP` both on and off

dustinbyrne:
I'd really like to see some additional test cases for what's been described above. Adding a route as a root node still does not work as expected, nothing is filtered out of the diagram or the trace.

dustinbyrne:
Changing filters under the 'declutter' section does not update the sidebar, however it works properly when changing the root node filter. This should behave in a consistent manner, otherwise it's quite confusing. I'd like to see a test case for this as well.

packages/components/src/pages/VsCodeExtension.vue
sashahonchar: @dustinbyrne  According to AppMap specification `httpServerResponse` doesn't have `headers` attribute:
https://github.com/applandinc/appmap#http-server-request-return-attributes

packages/components/src/pages/VsCodeExtension.vue
sashahonchar: Finally I get it, fixed https://github.com/applandinc/appmap-js/pull/429/commits/857887c6b2744c42bf9d5402576d28837df4a8f3

packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -934,11 +934,9 @@ export default {
       ];
 
       events.forEach((e) => {
-        const eventName = e.codeObject.prettyName;
-
         if (
-          eventName.startsWith('GET') &&
-          mediaFileExtensions.some((ext) => eventName.endsWith(ext))
+          e.requestMethod === 'GET' &&
+          mediaFileExtensions.some((ext) => e.requestPath.endsWith(ext))
```
dustinbyrne: `requestPath` can be destructured to avoid invoking the accessor for each media file extension.

packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -487,10 +492,14 @@ export default {
...
-          classMap.codeObjects.forEach((codeObject) => {
-            if (!filterRegExp.test(codeObject.fqid)) {
+        classMap.codeObjects.forEach((codeObject) => {
+          this.filters.rootObjects.forEach((id) => {
+            if (id.includes('*')) {
+              const filterRegExp = new RegExp(id, 'ig');
+              if (!filterRegExp.test(codeObject.fqid)) {
+                return;
+              }
+            } else if (id !== codeObject.fqid) {
```
dustinbyrne, line 502: I don't think this is what we want. The regex star operator matches zero or more of the preceding token and doesn't act as a glob. If we intend for this to be a glob, the regex should be generated from the input (e.g. `package:*/controllers` -> `^package:.*/controllers$`)

As written now, if you're looking to match `package:app/controllers`, the root filter `package:*/controllers` will yield no results. There should be a test case that validates this behaves as intended.

packages/components/src/pages/VsCodeExtension.vue
sashahonchar: Done https://github.com/applandinc/appmap-js/pull/429/commits/41369ecb8ac787c473019980f31cdadfb920fde5

packages/components/src/pages/VsCodeExtension.vue
sashahonchar: Looks like I forgot to replace `*` with `.*`.
I've made changes and added new e2e test for `package:*/controllers` filter query.
https://github.com/applandinc/appmap-js/pull/429/commits/656103591f61226dc3b1c04881df6eb14dcccf1e

packages/components/src/pages/VsCodeExtension.vue
dustinbyrne: It looks like that needs to be updated. `headers` is available as an optional parameter, preferred over `mime_type`.

kgilpin:
Dustin's right, according to Mozilla the header name is case-insensitive

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers

(although something like `content_type` would not be valid)

packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -936,14 +936,16 @@ export default {
...
-          e.http_server_response.mime_type &&
-          mediaRegex.some(
-            (regex) => regex.test(e.http_server_response.mime_type) // 'mime_type' is no longer supported in the AppMap data standard, but we should keep this code for backward compatibility
-          )
-        ) {
-          excludedEvents.push(e.parent_id);
+        } else if (e.http_server_response) {
+          const mimeType =
+            e.http_server_response.headers &&
+            e.http_server_response.headers['content-type']
```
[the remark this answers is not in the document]
sashahonchar: Done https://github.com/applandinc/appmap-js/pull/429/commits/256ed0e41988d78e92240aae74fae5e27f8a4f5b

dustinbyrne (approved):

PR #446, opened 2021-11-04
Use IntelliJ JDK and build tools

### LINKED ISSUE #413: Investigate running within IntelliJ context (binaries, environment)

apotterri:
Use the IntelliJ embedded JDK and build tools, if available and not overriden, when installing and validating the Java agent. The user can override the use of the embedded components by setting the `PATH` and `JAVA_HOME` environment variables appropriately.

Fixes #413.

### DIFF
--- a/packages/cli/src/cmds/agentInstaller/agentInstallerProcedure.ts
+++ b/packages/cli/src/cmds/agentInstaller/agentInstallerProcedure.ts
@@ -62,7 +62,7 @@ export default class AgentInstallerProcedure extends AgentProcedure {
 
     await this.installer.installAgent();
 
-    this.verifyProject();
+    await this.verifyProject();
 
     const appMapYml = this.configPath;
     if (!useExistingAppMapYml) {
--- a/packages/cli/src/cmds/agentInstaller/javaBuildToolInstaller.ts
+++ b/packages/cli/src/cmds/agentInstaller/javaBuildToolInstaller.ts
@@ -1,7 +1,30 @@
+import glob from 'glob';
+import os from 'os';
+import path from 'path';
 import chalk from 'chalk';
 import CommandStruct from './commandStruct';
 import { run } from './commandRunner';
 import { getOutput } from './commandUtil';
+import { findIntelliJHome } from './jetBrainsSupport';
+
+export function addJetBrainsEnv() {
+  const jbHome = findIntelliJHome();
+  if (!jbHome) {
+    return;
+  }
+
+  const javaHome = path.join(jbHome, 'jbr');
+  const gradleBin = path.join(jbHome, 'lib');
+  const mvnBin = path.join(jbHome, 'plugins/maven/maven3/bin');
+
+  // Make sure we don't override the user's settings: append to path, use the
+  // existing JAVA_HOME if it's set.
+  if (!process.env['JAVA_HOME']) {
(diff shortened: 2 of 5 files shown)


packages/cli/src/cmds/agentInstaller/jetBrainsSupport.ts
```diff
@@ -0,0 +1,17 @@
...
+  if (os.platform() === 'win32') {
+    defaultDir = 'C:/Program Files/JetBrains/*';
+  } else if (os.platform() === 'darwin') {
+    defaultDir = '/Applications/IntelliJ*.app/Contents';
+  } else {
+    return;
+  }
+
+  let jbHome = glob.sync(path.join(defaultDir));
+  return jbHome.length !== 0 ? jbHome[0] : undefined;
```
dustinbyrne, line 16: Maybe we should prompt the user and allow them to chose an installation.

packages/cli/src/cmds/agentInstaller/javaBuildToolInstaller.ts
```diff
@@ -1,7 +1,30 @@
...
+  const javaHome = path.join(jbHome, 'jbr');
+  const gradleBin = path.join(jbHome, 'lib');
+  const mvnBin = path.join(jbHome, 'plugins/maven/maven3/bin');
+
+  // Make sure we don't override the user's settings: append to path, use the
+  // existing JAVA_HOME if it's set.
+  if (!process.env['JAVA_HOME']) {
+      process.env['JAVA_HOME'] = javaHome;
+  };
+  process.env['PATH'] = [process.env['PATH'], path.join(javaHome, 'bin'), gradleBin, mvnBin].join(path.delimiter);
```
dustinbyrne, line 25: If an `.idea` directory is present, I think we'd want to reverse the order here and add the IntelliJ paths to the beginning of `PATH` so they're guaranteed to be picked up. Maybe we ask the user?

packages/cli/src/cmds/agentInstaller/agentInstallerProcedure.ts
```diff
@@ -62,7 +62,7 @@ export default class AgentInstallerProcedure extends AgentProcedure {
 
     await this.installer.installAgent();
 
-    this.verifyProject();
+    await this.verifyProject();
```
dustinbyrne, line 65: Good catch!

dustinbyrne (approved):
👍 Looks good.

I've added some comments below, but I don't know that they need to be addressed immediately. It's hard to say without more data. I think it's worth shipping this as-is and observing the success rate.

packages/cli/src/cmds/agentInstaller/javaBuildToolInstaller.ts
apotterri: This doesn't seem right to me. If the user has added a JDK to their `PATH`, they probably did it for a reason (e.g. they explicitly don't want to use the embedded JDK). 

What would be a case where we want to override what the user has done?

packages/cli/src/cmds/agentInstaller/agentInstallerProcedure.ts
apotterri: This problem was hard to see on macOS, but really obvious on Windows. Running two instances of `mvn` in parallel caused errors when one tried to use a file that the other was downloading.

PR #449, opened 2021-11-10
fix: Map language types to OpenAPI types

### LINKED ISSUE #445: openapi: openapi does not generate valid parameter types
The docs are generated with language-native parameter type names instead of openapi-specced names. Examples:

generated from Python AppMaps:
```yaml
        - name: secret
          in: path
          schema:
            type: builtins.str
          required: true
        - name: thumbnail
          in: query
          schema:
            type: builtins.bool
```

generated from Java AppMaps:
```yaml
        - name: id
          in: path
          schema:
            type: java.lang.string
          required: true
```

Valid types for these two examples are `boolean`, `string`.

apotterri:
Add type mappings for Python and Java. This ensures that all types that
can appear in AppMaps will get mapped to the corresponding OpenAPI type.

Fixes #445.

### DIFF
--- a/packages/cli/src/openapi/util.js
+++ b/packages/cli/src/openapi/util.js
@@ -1,24 +1,69 @@
 const { buildAppMap } = require('@appland/models');
 
+const unrecognizedTypes = new Set();
+
 function classNameToOpenAPIType(className) {
-  let typeName = className;
-  if (!typeName || typeName === '') {
+  if (!className || className === '') {
     return 'unknown';
   }
-  typeName = typeName.toLowerCase();
 
-  switch (typeName) {
-    case 'hash':
-    case 'activesupport::hashwithindifferentaccess':
-      return 'object';
-    case 'nilclass':
-      return 'string';
-    case 'trueclass':
-    case 'falseclass':
-      return 'boolean';
-    default:
-      return typeName;
+  const mapRubyType = (t) => {
+    switch (t) {
+      case 'array':
+        return 'array';
+      case 'hash':
+      case 'activesupport::hashwithindifferentaccess':
+        return 'object';
+      case 'nilclass':
+        return 'string';
+      case 'trueclass':
+      case 'falseclass':
+        return 'boolean';
+      default:
+        return undefined;
+    }
+  };
+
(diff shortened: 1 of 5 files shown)


packages/cli/src/openapi/util.js
```diff
@@ -1,24 +1,60 @@
...
+      case 'java.lang.string':
+        return 'string';
+      default:
+        return undefined;
+    }
+  };
+
+  const mapper = (t) => mapRubyType(t) || mapPythonType(t) || mapJavaType(t);
+
+  return mapper(className.toLowerCase()) || className;
```
apotterri, line 57: @kgilpin Is there a reason to ignore case when comparing types here?

apotterri, line 57: @kgilpin @ptrdvrk The previous version of this code also just returns the class name if no other mapping is found. But, is this really the behavior we want? At the very least, it seems like we should also log a warning that we're generating an invalid spec.

ptrdvrk (approved):
@apotterri from a user's perspective this look good to me, thank you. swagger.io still complains about one issue in the Misago docs - but that is a different problem.

https://app.swaggerhub.com/apis/ptrdvrk8/misago/v1

apotterri:
Thanks, @ptrdvrk. Do you have an opinion about what the `type` of a parameter should be set to for unmapped types, or if we should log a message if we see one? For example, what should we have done when we encountered `java.lang.String`?

ptrdvrk:
@apotterri I would log a warning but default to the class name, with a comment in the generated doc file. I think it is more valuable to have an openapi document that requires a few edits vs not having one at all.

kgilpin:
Me too.

apotterri:
@ptrdvrk Logging the warning is easy. Getting a comment into the generated doc is going to take a fair bit of work -- the point where we detect the problem is pretty far from where the doc gets created.

Do you want to wait for the comment to be added, or should we do that as a follow on change?

kgilpin:
We don't need the comment at this time. It's going to be fairly clear to
the user what needs to be done.

ptrdvrk:
+1, let's release without comments now.

PR #450, opened 2021-11-10
fix: Wait for debugging information on exit

dustinbyrne:
`Yargs.exit` results in a call to `process.exit`, immediately exiting the process without waiting for promises to resolve. This was resulting in us losing some data. As a result, all telemetry events are now sent without `await`, and if the queue is not clear by the time the command is finishing, we'll wait for the queue to flush. This won't capture exit via `SIGINT` or similar, but that's a more complicated effort.

### DIFF
--- a/packages/cli/src/cmds/agentInstaller/install-agent.ts
+++ b/packages/cli/src/cmds/agentInstaller/install-agent.ts
@@ -57,7 +57,7 @@ class InstallerError {
       .join(', ');
 
     if (this.error instanceof AbortError) {
-      await Telemetry.sendEvent({
+      Telemetry.sendEvent({
         name: 'install-agent:abort',
         properties: {
           installer: this.project?.selectedInstaller?.name,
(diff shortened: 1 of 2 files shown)


packages/cli/src/cmds/agentInstaller/install-agent.ts
```diff
@@ -231,6 +231,8 @@ export default {
     }
 
     if (errors.length) {
+      // Make sure there's nothing left in the queue before we exit.
+      await Telemetry.flush();
       Yargs.exit(1, new Error(errors.map((e) => e.message).join('\n')));
     }
```
apotterri, line 237:
```suggestion
    // Make sure there's nothing left in the queue before we exit.
    await Telemetry.flush();
    if (errors.length) {
      Yargs.exit(1, new Error(errors.map((e) => e.message).join('\n')));
    }
```
We send telemetry events when things succeed, too. Shouldn't this flush happen unconditionally?

apotterri (changes requested):
Seems like we should always be flushing telemetry events before exiting.

packages/cli/src/cmds/agentInstaller/install-agent.ts
dustinbyrne: Whoops. Yes. I changed this last minute. Thanks for catching that.

apotterri (approved):

PR #456, opened 2021-11-12
fix: Rebuild packages before publish

dustinbyrne:
In case the published package copies package.json, rebuild the package before publishing artifacts.

### DIFF
--- a/.releaserc.js
+++ b/.releaserc.js
@@ -10,10 +10,10 @@ module.exports = {
         releaseRules: [
           {
             type: 'feature',
-            release: 'minor'
-          }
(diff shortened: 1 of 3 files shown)


apotterri (changes requested):
This certainly seems like it's worth a shot. Since it's a completely different approach from the one I took, though, will you take the `prepack` hooks out of `package.json`, and `packages/cli/package.json`, please? As far as I could tell, they didn't help, and leaving them in may lead to confusion. Thanks!

apotterri (approved):
Thanks, LGTM.

PR #458, opened 2021-11-17
Fix hash generation for events without parameters/labels

torchello:
FIxes `TypeError: Cannot read property 'forEach' of undefined` error.

### DIFF
--- a/packages/models/src/util.js
+++ b/packages/models/src/util.js
@@ -717,10 +717,11 @@ export function hashEvent(e) {
 
   const content = [];
   getStaticPropValues(e).forEach((v) => content.push(v));
-  e.parameters.forEach((p) =>
-    getStaticPropValues(p).forEach((v) => content.push(v))
-  );
-  [...e.labels].forEach((l) => content.push(l));
+  if (e.parameters) {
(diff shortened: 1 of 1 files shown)


packages/models/src/util.js
```diff
@@ -717,10 +717,14 @@ export function hashEvent(e) {
...
-  );
-  [...e.labels].forEach((l) => content.push(l));
+  if (e.parameters) {
+    e.parameters.forEach((p) =>
+      getStaticPropValues(p).forEach((v) => content.push(v))
+    );
+  }
+  if (e.labels) {
+    [...e.labels].forEach((l) => content.push(l));
+  }
```
dustinbyrne, line 727: Was this occurring? `labels` should always exist.

dustinbyrne: The labels need not be copied into an array either, the `forEach` method will exist on the `Set`.

kgilpin:
From what I’ve seen, labels is always Set, never null

torchello:
As discussed with @dustinbyrne removing labels for now.

dustinbyrne (approved):

PR #463, opened 2021-11-22
feat: show event exceptions in details panel

### LINKED ISSUE #462: Exceptions not listed in left side panel
When selecting an event that threw an exception the details of the exception should be available in the left side panel.

sashahonchar:
image

Fixes #462

### DIFF
--- a/packages/components/src/components/DetailsPanelEvent.vue
+++ b/packages/components/src/components/DetailsPanelEvent.vue
@@ -72,6 +72,16 @@
       </ul>
     </div>
 
+    <div class="event-params" v-if="object.exceptions.length">
+      <h5>Exceptions</h5>
+      <ul class="table-01">
+        <li v-for="exception in object.exceptions" :key="exception.object_id">
+          <strong>{{ exception.class }}</strong>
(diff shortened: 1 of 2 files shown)


dustinbyrne:
The filters overflow if the viewport is not vertically big enough:
image

Some root node filters may result in no data:

https://user-images.githubusercontent.com/8737782/142919635-334779e1-ede9-471f-8a72-7fda18c87682.mp4

Some root node filters appear to break the diagram and display no events in the trace:

https://user-images.githubusercontent.com/8737782/142919689-d9841571-7f04-4b7d-88db-41a15a5feca6.mp4

I suggest adding tests for the following edge cases. Once they're fixed, try and find a few more and see if you can break it.
- A root node filter added for `package` renders as expected and contains the expected code objects and events
- A root node filter added for `class:app/controllers/Spree` renders as expected and contains the expected code objects and trace events
- A root node filter added for a route renders as expected and contains the expected code objects and events, with `Limit root events to HTTP` both on and off

kgilpin:
@dustinbyrne this PR is for showing Exception details.

dustinbyrne (approved):

PR #466, opened 2021-11-28
feat: Add JavaScript project installers

apotterri:
Add installers for NPM and Yarn projects.

Also, update config validation so it can use either a single schema or
an array of schemas.


dustinbyrne (approved):
💯

PR #468, opened 2021-11-30
fix: Use correct path to JB maven plugin

apotterri:
Use the correct path to the IntelliJ maven plugin. IntelliJ doesn't have
an embedded gradle executable.


ptrdvrk (approved):
LGTM @apotterri.

PR #475, opened 2021-12-01
feat: filter events in Trace by label

### LINKED ISSUE #464: Event search in trace matches labels
The search bar in the trace view allows `label` search terms and highlights every event with the given label(s). If more than one label is provided, all labels must match.

image

sashahonchar:
image

Fixes #464

### DIFF
--- a/packages/components/src/pages/VsCodeExtension.vue
+++ b/packages/components/src/pages/VsCodeExtension.vue
@@ -590,6 +590,13 @@ export default {
               if (!Number.isNaN(eventId) && knownEventIds.includes(eventId)) {
                 nodes.push(eventId);
               }
+            } else if (item.startsWith('label:')) {
+              const labelName = item.replace('label:', '');
+              this.filteredAppMap.events.forEach((event) => {
+                if (event.isCall() && event.labels.has(labelName)) {
+                  nodes.push(event.id);
+                }
+              });
             }
           });
         }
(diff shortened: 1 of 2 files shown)


packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -590,6 +590,13 @@ export default {
               if (!Number.isNaN(eventId) && knownEventIds.includes(eventId)) {
                 nodes.push(eventId);
               }
+            } else if (item.startsWith('label:')) {
+              const labelName = item.replace('label:', '');
+              this.filteredAppMap.events.forEach((event) => {
+                if (event.isCall() && event.labels.has(labelName)) {
+                  nodes.push(event.id);
```
kgilpin, line 597: Suggest using a Set to deal with events that match more than one filter expression.

packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -590,6 +590,13 @@ export default {
               if (!Number.isNaN(eventId) && knownEventIds.includes(eventId)) {
```
kgilpin, line 590: Looking up an id in an Array is inefficient - use a Set instead.

kgilpin (changes requested):

dustinbyrne (approved):

PR #496, opened 2021-12-13
feat: highlight events while looping through filtered events

### LINKED ISSUE #489: Event search highlights should modify the selected object, displaying details of the highlighted object in the left side bar

### LINKED ISSUE #490: Events have a secondary highlight when they match the trace view search criteria, but are not the currently selected event
image

### LINKED ISSUE #491: Pan to event stops working predictably when cycling through event search results and an event is selected
https://user-images.githubusercontent.com/8737782/145274694-6597a913-0678-478b-b75b-95e2d03a7dd8.mp4

sashahonchar:
image

Fixes #489 
Fixes #490 
Fixes #491

### DIFF
--- a/packages/components/src/components/trace/Trace.vue
+++ b/packages/components/src/components/trace/Trace.vue
@@ -50,8 +50,8 @@ export default {
       default: null,
     },
     highlightedEvents: {
-      type: Array,
-      default: () => [],
+      type: Set,
+      default: new Set(),
     },
     highlightedEventId: {
       type: Number,
--- a/packages/components/src/components/trace/Trace.vue
+++ b/packages/components/src/components/trace/Trace.vue
@@ -6,6 +6,7 @@
       :event="event"
       :selected-events="selectedEvents"
       :focused-event="focusedEvent"
+      :highlighted-events="highlightedEvents"
       :highlighted-event-id="highlightedEventId"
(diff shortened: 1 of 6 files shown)


packages/components/src/components/trace/Trace.vue
```diff
@@ -48,6 +49,10 @@ export default {
       type: Object,
       default: null,
     },
+    highlightedEvents: {
+      type: Array,
+      default: () => [],
```
dustinbyrne, line 54: This should be a `Set`, otherwise this may lead to an `O(n^2)` operation as each event has to iterate the array to check if it must render a highlight. Everywhere we're drilling this prop should use a `Set` as well.

dustinbyrne (changes requested):
Definite improvement. I've requested one change below, and I think this is significant enough that it warrants a test.

packages/components/src/components/trace/Trace.vue
sashahonchar: Done https://github.com/applandinc/appmap-js/pull/496/commits/7faf9ed463eb9aae887b92f8fb2fa2db1d7f62e8

dustinbyrne:
This still needs a test that verifies the expected events receive the correct styling.

dustinbyrne (approved):

PR #509, opened 2021-12-28
feat: get/set global filters state with state API

### LINKED ISSUE #505: Default filters can be modified via `setState` API
Use case:
The user is linked to an AppMap with a selected event that is not visible under the default filters. For example, a query that occurs between HTTP web requests can not be selected due to the default filter to restrict root events to HTTP.

sashahonchar:
Fixes #505

### DIFF
--- a/packages/components/tests/unit/VsCodeExtension.spec.js
+++ b/packages/components/tests/unit/VsCodeExtension.spec.js
@@ -27,6 +27,16 @@ describe('VsCodeExtension.vue', () => {
 
     wrapper.vm.setState('{"selectedObject":"class:app/models/User"}');
     expect(wrapper.vm.selectedObject.id).toMatch('app/models/User');
+
+    wrapper.vm.setState(
+      '{"currentView":"viewFlow","filters":{"rootObjects":[],"limitRootEvents":false,"hideMediaRequests":true,"hideUnlabeled":false,"hideElapsedTimeUnder":100,"hideName":false}}'
+    );
+    expect(wrapper.vm.filters.declutter.limitRootEvents.on).toBe(false);
+    expect(wrapper.vm.filters.declutter.hideMediaRequests.on).toBe(true);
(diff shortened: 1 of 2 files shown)


dustinbyrne (changes requested):
The change looks good but this needs test cases.

sashahonchar:
@dustinbyrne  Found a bug in code and added unit test. https://github.com/applandinc/appmap-js/pull/509/commits/b5a1835fa8ac26ccaad3d22f7dfa106e45aeff92

packages/components/tests/unit/VsCodeExtension.spec.js
```diff
@@ -27,6 +27,16 @@ describe('VsCodeExtension.vue', () => {
...
+
+    wrapper.vm.setState(
+      '{"currentView":"viewFlow","filters":{"rootObjects":[],"limitRootEvents":false,"hideMediaRequests":true,"hideUnlabeled":false,"hideElapsedTimeUnder":100,"hideName":false}}'
+    );
+    expect(wrapper.vm.filters.declutter.limitRootEvents.on).toBe(false);
+    expect(wrapper.vm.filters.declutter.hideMediaRequests.on).toBe(true);
+    expect(wrapper.vm.filters.declutter.hideUnlabeled.on).toBe(false);
+    expect(wrapper.vm.filters.declutter.hideElapsedTimeUnder.on).toBe(true);
+    expect(wrapper.vm.filters.declutter.hideElapsedTimeUnder.time).toBe(100);
+    expect(wrapper.vm.filters.declutter.hideName.on).toBe(false);
```
dustinbyrne, line 39: `limitRootEvents` / `hideUnlabeled` / `hideName` properties have no test coverage in this case.

dustinbyrne (changes requested):
Looking better. We'll also need an updated test for `getState`, where the filters contain non-default values. Plus a comment below.

dustinbyrne (approved):

PR #519, opened 2022-01-11
fix: select events outside filtered with event filter

### LINKED ISSUE #517: Events can be selected when matches are present using the event filter
See https://www.loom.com/share/3ae20cb71d3642f2989467b086693dbe

The user should be able to click into other events at any time.

Acceptance tests:
```
GIVEN a valid query (e.g. `label:json`) is entered into the event filter input
WHEN a non-zero amount of events are matched
THEN another event can be selected via click
AND the events matched by the event filter query all receive the unselected highlight styling
```

```
GIVEN a valid query (e.g. `label:json`) is entered into the event filter input
WHEN greater than one event is matched
THEN another event can be selected via click
AND clicking the next button or previous results button once re-selects the nearest matching event
```

sashahonchar:
Fixes #517

### DIFF
--- a/packages/components/src/pages/VsCodeExtension.vue
+++ b/packages/components/src/pages/VsCodeExtension.vue
@@ -69,14 +69,17 @@
               </div>
               <div
                 class="trace-view__search-arrows"
-                v-if="highlightedNodes.size"
+                v-if="highlightedNodes.length"
               >
                 <div class="trace-view__search-arrow" @click="prevTraceFilter">
                   <ArrowSearchLeftIcon />
                 </div>
                 <div class="trace-view__search-arrows-text">
-                  <b>{{ currentTraceFilterIndex + 1 }}</b
-                  >/{{ highlightedNodes.size }} results
+                  <span v-if="Number.isInteger(currentTraceFilterIndex)">
+                    <b>{{ currentTraceFilterIndex + 1 }}</b>
+                    /
+                  </span>
+                  {{ highlightedNodes.length }} results
                 </div>
                 <div class="trace-view__search-arrow" @click="nextTraceFilter">
                   <ArrowSearchRightIcon />
@@ -88,8 +91,8 @@
               :events="filteredAppMap.rootEvents()"
               :selected-events="selectedEvent"
               :focused-event="focusedEvent"
-              :highlighted-events="highlightedNodes"
-              :highlighted-event-id="highlightedEventId"
+              :highlighted-events="new Set(highlightedNodes)"
+              :highlighted-event="highlightedEvent"
               :highlighted-event-index="currentTraceFilterIndex + 1"
               :name="VIEW_FLOW"
               :zoom-controls="true"
@@ -438,8 +441,21 @@ export default {
     },
     '$store.getters.selectedObject': {
       handler(selectedObject) {
-        if (selectedObject && !(selectedObject instanceof Event)) {
-          this.setView(VIEW_COMPONENT);
+        if (selectedObject) {
+          if (!(selectedObject instanceof Event)) {
+            this.setView(VIEW_COMPONENT);
+          }
+
+          if (selectedObject instanceof Event) {
+            const highlightedIndex = this.highlightedNodes.findIndex(
+              (e) => e === selectedObject
+            );
+
+            this.currentTraceFilterIndex =
+              highlightedIndex >= 0 ? highlightedIndex : undefined;
+          }
+        } else {
+          this.currentTraceFilterIndex = undefined;
         }
 
         this.$root.$emit('stateChanged', 'selectedObject');
@@ -463,6 +479,12 @@ export default {
     highlightedNodes: {
       handler() {
         this.currentTraceFilterIndex = 0;
+        this.selectCurrentHighlightedEvent();
+      },
+    },
+    currentTraceFilterIndex: {
+      handler() {
+        this.selectCurrentHighlightedEvent();
       },
     },
     filteredAppMap: {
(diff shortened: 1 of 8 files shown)


packages/components/tests/e2e/specs/vsCodeExtension.spec.js
```diff
@@ -747,6 +747,35 @@ context('VS Code Extension', () => {
       cy.get('.trace-node[data-event-id="15"]')
         .should('be.visible')
         .should('have.class', 'highlight');
+
+      cy.get('.trace-node[data-event-id="13"]')
+        .click()
+        .should('have.class', 'highlight');
+      cy.get('.trace-node[data-event-id="1"]').should(
+        'not.have.class',
```
dustinbyrne, line 755: The filtered nodes should retain the `filtered` class despite the active selection not being part of the filtered set.

packages/components/tests/e2e/specs/vsCodeExtension.spec.js
```diff
@@ -747,6 +747,35 @@ context('VS Code Extension', () => {
...
+      cy.get('.trace-node[data-event-id="3"]').should(
+        'not.have.class',
+        'filtered'
+      );
+      cy.get('.trace-node[data-event-id="15"]').should(
+        'not.have.class',
+        'highlight'
+      );
+
+      cy.get('.details-panel__buttons').contains('Clear selection').click();
```
dustinbyrne, line 767: I'd expect that `Clear selection` removes the active selection entirely, i.e. a highlighted event is not then selected

packages/components/tests/e2e/specs/vsCodeExtension.spec.js
```diff
@@ -747,6 +747,35 @@ context('VS Code Extension', () => {
...
+      cy.get('.trace-node[data-event-id="13"]').should(
+        'not.have.class',
+        'highlight'
+      );
+      cy.get('.trace-node[data-event-id="1"]').should('have.class', 'filtered');
+      cy.get('.trace-node[data-event-id="3"]').should('have.class', 'filtered');
+      cy.get('.trace-node[data-event-id="15"]').should(
+        'have.class',
+        'highlight'
+      );
```
dustinbyrne, line 778: There should also be a test where the next (and previous) button is clicked, and the active selection changes to the _nearest_ filtered event in that direction.

In other words, clicking the left search arrow moves to the nearest previous filtered event. The right search arrow moves to the nearest next filtered event.

image

packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -438,8 +438,21 @@ export default {
...
+        if (selectedObject) {
+          if (!(selectedObject instanceof Event)) {
+            this.setView(VIEW_COMPONENT);
+          }
+
+          if (
+            selectedObject instanceof Event &&
+            this.highlightedNodes.has(selectedObject.id)
+          ) {
+            this.currentTraceFilterIndex = Array.from(
```
dustinbyrne, line 450: If the active selection is not a filtered event, can we hide this number from this display?

For example, this element:
image


Should become:
image

If the user does not have a filtered event selected.

dustinbyrne (changes requested):
Thanks @sashahonchar! This interaction is feeling much better. I've left a couple of comments below for adjustments. Also do you mind implementing the acceptance tests defined here in the issue (#517)?

packages/components/tests/e2e/specs/vsCodeExtension.spec.js
sashahonchar: That action on line 767 removes active selection from event id=13 (which is not present in filter input), but it was clicked by user on line 752.

sashahonchar: After that, previously selected event (15) became selected again and user can continue loop through filtered events.

packages/components/src/pages/VsCodeExtension.vue
sashahonchar: fixed https://github.com/applandinc/appmap-js/pull/519/commits/c9bd5fee3eba5272ed597d26dd4775d7a8294dcc

packages/components/tests/e2e/specs/vsCodeExtension.spec.js
sashahonchar: fixed https://github.com/applandinc/appmap-js/pull/519/commits/c9bd5fee3eba5272ed597d26dd4775d7a8294dcc

packages/components/tests/e2e/specs/vsCodeExtension.spec.js
sashahonchar: Done https://github.com/applandinc/appmap-js/pull/519/commits/bdd8356bfa37b513a18ea9862d9deb28a2720586

Now all filtered events are ordered by ID, and I changed test input filter to verify that behavior:
https://github.com/applandinc/appmap-js/pull/519/commits/bdd8356bfa37b513a18ea9862d9deb28a2720586#diff-f5e13ae703840c989eb60ca988f5d4e016a49c59a6e4366c0cdeead459043f88R710

packages/components/tests/e2e/specs/vsCodeExtension.spec.js
dustinbyrne: `Clear selection` should remove the active selection entirely. It's unexpected that it selects another event and following that the button then does nothing.

packages/components/tests/e2e/specs/vsCodeExtension.spec.js
dustinbyrne: This is working much nicer than before. I think an ideal UX would be to select the nearest filtered events to the current selection, regardless of what the last selected filtered event index was.

As an example:
image

`ActiveSupport::SecurityUtils.secure_compare` is not part of the filtered set. If I click the previous or next filtered event button, I'd expect to always select one of the two `JSON::Ext::Parser#parse` events shown here as appropriate. This would be regardless of any previous filtered event index I had selected.

Let's save this for a future story, because the current behavior is still a much needed improvement over what we currently have.

sashahonchar: @dustinbyrne would you mind to create new issue for the future?

packages/components/tests/e2e/specs/vsCodeExtension.spec.js
sashahonchar: Got it. Now `Clear selection` button will make empty Trace view filter input.
https://github.com/applandinc/appmap-js/pull/519/commits/b346b187ba21a6d25084edf2372d22168ea3a58c

PR #520, opened 2022-01-14
fix: Drop node protocol import

dustinbyrne:
These imports restrict the node version to v16+. We could potentially remove them in a post-process build step, but I don't think it's worth the effort.


kgilpin (approved):
Ugh. The linter made me add this in the first place. How annoying.

dustinbyrne:
Yeah I had no idea either, I believe it was added within the last month.

PR #527, opened 2022-01-26
fix: Apply fixes to the record command

kgilpin:
See Git commit messages for details.


apotterri (approved):
I wish the subject lines for your commit messages were less than 72 characters. ;)
image

Changes look good otherwise.

PR #530, opened 2022-01-27
fix: Handle spaces in CommandStruct.args

apotterri:
When creating a CommandStruct, add quotes around any element of args that
contains a space.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,8 +27,11 @@ jobs:
       os: windows
       env:
         - YARN_GPG=no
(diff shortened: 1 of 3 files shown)


kgilpin:
Can we use spawn here? It doesn’t require escaping.

apotterri:
We are using spawn. But, we're also using `shell: true`, so that we can find non-.exe commands on Windows: https://github.com/applandinc/appmap-js/commit/12986a5bbaa97a8e9b671ab1308636813f06ef2d . When you tell spawn to use a shell, an argument containing spaces will get split into mulitple arguments, causing failures like https://github.com/applandinc/appmap-agent-js/issues/88 .

kgilpin (approved):

PR #531, opened 2022-01-28
feat: Improved event hash

dividedmind:
Hashing a call event now only considers the fields which are truly specific to the call generalically and not specific arguments that     were used. This also includes literals and variables in SQL queries.

Note that all events will now have different hashes than previously calculated, so the transition can confuse change tracking. Also the new algorithm refuses to hash non-call events (as they don't really have any data that is generic enough to make sense).

### DIFF
--- a/packages/models/src/event/hash.js
+++ b/packages/models/src/event/hash.js
@@ -1,0 +1,41 @@
+import sha256 from 'crypto-js/sha256';
+import { buildQueryAST } from '../util';
+
+// returns a JSON of SQL query AST
+// with all literals replaced by variables
+// and all variable names removed
+function abstractSqlAstJSON(query) {
+  const ast = buildQueryAST(query);
+  return JSON.stringify(ast, (_, value) => {
+    switch (value.type) {
+      case 'variable':
+      case 'literal':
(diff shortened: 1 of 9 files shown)


packages/models/src/event/hash.js
```diff
@@ -0,0 +1,40 @@
...
+
+// Returns a string suitable for durable identification of a call event
+// that's independent of parameters and return values.
+// For SQL queries, it's a JSON of abstract query AST.
+// For HTTP queries, it's the method plus normalized path info.
+// For function calls it's the qualified function id.
+function callEventToString(event) {
+  const { sqlQuery } = event;
+  if (sqlQuery) return abstractSqlAstJSON(sqlQuery);
+  return event.toString();
```
kgilpin, line 28: It seems like relying on `event.toString()` might be risky for the purposes of a stable hash. `toString` is probably thought of as more of a debugging / informational method and someone might change `toString` for HTTP events - and break hashing.

dividedmind: You might be right. I've been a bit on the fence about it.

kgilpin (approved):

PR #534, opened 2022-01-31
feat: auto-complete suggestions list for Trace view filter

sashahonchar:
image

### DIFF
--- a/packages/components/src/components/trace/TraceFilter.vue
+++ b/packages/components/src/components/trace/TraceFilter.vue
@@ -127,9 +127,25 @@ export default {
       return classList;
     },
     suggestionsList() {
-      return this.suggestions
+      let suggestions = this.suggestions
         .map((e) => e.toString())
         .filter((e, i, arr) => arr.indexOf(e) === i);
+
+      const terms = this.filterValue.match(/(?:[^\s"]+|"[^"]*")+/g);
+      const lastTerm = terms ? terms[terms.length - 1] : null;
+      if (
+        lastTerm &&
+        !/^".+"$/g.test(lastTerm) &&
+        !/^%.+%$/g.test(lastTerm) &&
+        !/^id:.+$/g.test(lastTerm) &&
+        !/^label:.+$/g.test(lastTerm)
+      ) {
+        suggestions = suggestions.filter((item) =>
+          new RegExp(lastTerm, 'ig').test(item)
+        );
+      }
+
+      return suggestions;
     },
   },
 
@@ -161,15 +177,28 @@ export default {
       if (
         this.$refs.input === window.document.activeElement &&
         this.showSuggestions &&
-        this.suggestions.length
+        this.suggestionsList.length
       ) {
         this.makeSelection(
-          this.suggestions[this.selectedSuggestion].toString()
+          this.suggestionsList[this.selectedSuggestion].toString()
         );
       }
     },
     makeSelection(eventName) {
-      this.filterValue = `${this.filterValue} "${eventName}"`.trim();
+      const terms = this.filterValue.match(/(?:[^\s"]+|"[^"]*")+/g);
+      const lastTerm = terms ? terms[terms.length - 1] : null;
+      if (
+        lastTerm &&
+        !/^".+"$/g.test(lastTerm) &&
+        !/^%.+%$/g.test(lastTerm) &&
+        !/^id:.+$/g.test(lastTerm) &&
+        !/^label:.+$/g.test(lastTerm)
+      ) {
+        terms.pop();
+      }
+      this.filterValue = `${
+        terms ? terms.join(' ') : ''
+      } "${eventName}"`.trim();
     },
     onWindowClick(event) {
       if (!getEventTarget(event.target, this.$refs.form)) {
@@ -187,7 +216,7 @@ export default {
           this.$refs.input.blur();
           break;
         case 'ArrowDown':
-          if (this.selectedSuggestion !== this.suggestions.length - 1) {
+          if (this.selectedSuggestion !== this.suggestionsList.length - 1) {
             this.selectedSuggestion += 1;
           }
           break;
--- a/packages/components/src/components/trace/TraceFilter.vue
+++ b/packages/components/src/components/trace/TraceFilter.vue
@@ -23,11 +23,11 @@
       >
         <li
           :class="getSuggestionClasses(index)"
-          v-for="(item, index) in suggestions"
-          :key="item.id"
-          @click.stop="makeSelection(item.id)"
+          v-for="(item, index) in suggestionsList"
+          :key="item"
+          @click.stop="makeSelection(item)"
         >
-          {{ item.toString() }}
+          {{ item }}
         </li>
       </ul>
     </form>
@@ -126,6 +126,11 @@ export default {
 
       return classList;
     },
+    suggestionsList() {
+      return this.suggestions
+        .map((e) => e.toString())
+        .filter((e, i, arr) => arr.indexOf(e) === i);
+    },
   },
 
   methods: {
@@ -158,12 +163,13 @@ export default {
         this.showSuggestions &&
         this.suggestions.length
       ) {
-        this.makeSelection(this.suggestions[this.selectedSuggestion].id);
+        this.makeSelection(
+          this.suggestions[this.selectedSuggestion].toString()
+        );
       }
     },
-    makeSelection(eventId) {
-      this.filterValue = `${this.filterValue} id:${eventId}`.trim();
-      this.showSuggestions = false;
+    makeSelection(eventName) {
+      this.filterValue = `${this.filterValue} "${eventName}"`.trim();
     },
     onWindowClick(event) {
       if (!getEventTarget(event.target, this.$refs.form)) {
--- a/packages/components/src/components/trace/TraceFilter.vue
+++ b/packages/components/src/components/trace/TraceFilter.vue
@@ -12,6 +12,7 @@
         autocomplete="off"
         placeholder="search events..."
         @focus="onInputFocus"
+        @blur="onInputBlur"
       />
       <span v-if="filterValue" class="trace-filter__suffix" @click="clearValue">
         <CloseThinIcon />
@@ -131,23 +132,21 @@ export default {
       return classList;
     },
     suggestionsList() {
-      let suggestions = this.suggestions
-        .map((e) => e.toString())
-        .filter((e, i, arr) => arr.indexOf(e) === i);
+      let { suggestions } = this;
+      const lastTermRegex = this.filterValue.includes('"')
+        ? /([^"]*)$/
+        : /([^\s]*)$/;
+      const lastTerm = this.filterValue.match(lastTermRegex).pop().trim();
 
-      if (!this.filterValue.endsWith(' ')) {
-        const terms = this.filterValue.match(/(?:[^\s"]+|"[^"]*")+/g);
-        const lastTerm = terms ? terms[terms.length - 1] : null;
-        if (
-          lastTerm &&
-          !/^".+"$/g.test(lastTerm) &&
-          !/^id:.+$/g.test(lastTerm) &&
-          !/^label:.+$/g.test(lastTerm)
-        ) {
-          suggestions = suggestions.filter((item) =>
-            new RegExp(lastTerm, 'ig').test(item)
-          );
-        }
+      if (
+        lastTerm &&
+        !/^".+"$/g.test(lastTerm) &&
+        !/^id:.+$/g.test(lastTerm) &&
+        !/^label:.+$/g.test(lastTerm)
+      ) {
+        suggestions = suggestions.filter((item) =>
+          new RegExp(lastTerm, 'ig').test(item)
+        );
       }
 
       return suggestions;
@@ -178,6 +177,11 @@ export default {
       this.showSuggestions = true;
       this.selectedSuggestion = false;
     },
+    onInputBlur() {
+      if (!this.filterValue.endsWith(' ')) {
+        this.filterValue += ' ';
+      }
+    },
     onFormSubmit() {
       if (
         this.$refs.input === window.document.activeElement &&
@@ -188,8 +192,6 @@ export default {
         this.makeSelection(
           this.suggestionsList[this.selectedSuggestion].toString()
         );
-      } else {
-        this.filterValue += ' ';
       }
 
       this.$refs.input.blur();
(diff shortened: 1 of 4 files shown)


packages/components/src/components/trace/TraceFilter.vue
```diff
@@ -0,0 +1,354 @@
...
+        if (selected.offsetTop < this.$refs.suggestionsList.scrollTop) {
+          this.$refs.suggestionsList.scrollTop = selected.offsetTop;
+        } else if (
+          selected.offsetTop >=
+          this.$refs.suggestionsList.scrollTop +
+            this.$refs.suggestionsList.offsetHeight
+        ) {
+          this.$refs.suggestionsList.scrollTop =
+            selected.offsetTop - selected.offsetHeight * 2;
+        }
```
dustinbyrne, line 113: There's something off about this. When hitting the down arrow key, the first item that is below the visible area is not shown. This video is recorded on Chrome 97.

https://user-images.githubusercontent.com/8737782/152190699-f9b51098-0047-4721-b2ad-4c4b2de200e8.mp4

packages/components/src/components/trace/TraceFilter.vue
```diff
@@ -0,0 +1,354 @@
...
+      this.showSuggestions = true;
+      this.selectedSuggestion = 0;
+    },
+    onFormSubmit() {
+      if (
+        this.$refs.input === window.document.activeElement &&
+        this.showSuggestions &&
+        this.suggestions.length
+      ) {
+        this.makeSelection(this.suggestions[this.selectedSuggestion].id);
```
dustinbyrne, line 160: I'd expect this to input the same text that's suggested, meaning we're searching by name and not an identifier. The suggestions would need to be de-duplicated as a result.

image


https://user-images.githubusercontent.com/8737782/152193373-dfe9980c-43cb-40ce-a3d7-944747d25087.mp4

packages/components/src/components/trace/TraceFilter.vue
```diff
@@ -0,0 +1,354 @@
...
+        <CloseThinIcon />
+      </span>
+      <ul
+        ref="suggestionsList"
+        v-if="suggestions.length"
+        class="trace-filter__suggestions"
+      >
+        <li
+          :class="getSuggestionClasses(index)"
+          v-for="(item, index) in suggestions"
```
dustinbyrne, line 26: `suggestions` should update as the user inputs text into the search bar. The selected list item should be the best match, similar to the root object filters.

https://user-images.githubusercontent.com/8737782/152194221-c856ac03-88d0-439a-8abd-f1e225d4fe59.mp4

packages/components/src/components/trace/TraceFilter.vue
dustinbyrne: This is somewhat complicated as a suggestion may contain spaces, and we have no way of filtering a literal that contains whitespace because we split the search term into queries delimited by whitespace. `POST /users` will search for `POST && /users`. Ideally we'd allow the query to be quoted (`"POST /users"`) and not process any whitespace contained within.

packages/components/src/components/trace/TraceFilter.vue
```diff
@@ -0,0 +1,354 @@
...
+        this.$refs.input === window.document.activeElement &&
+        this.showSuggestions &&
+        this.suggestions.length
+      ) {
+        this.makeSelection(this.suggestions[this.selectedSuggestion].id);
+      }
+    },
+    makeSelection(eventId) {
+      this.filterValue = `${this.filterValue} id:${eventId}`.trim();
+      this.showSuggestions = false;
```
dustinbyrne, line 165: The text input is still focused after submitting, so the user is free to type another query, but they'll have no suggestions for the second query

https://user-images.githubusercontent.com/8737782/152193174-0147524c-3d2d-4900-8802-7b0c1dfbaa7b.mp4

.

packages/components/src/components/trace/TraceFilter.vue
```diff
@@ -0,0 +1,354 @@
...
+      }
+    },
+    onWindowKeyUp(event) {
+      if (this.$refs.input !== window.document.activeElement) {
+        return;
+      }
+
+      switch (event.key) {
+        case 'Escape':
+          this.showSuggestions = false;
```
dustinbyrne, line 179: We should clear the focus on the input as well. Otherwise the user needs to drop and regain focus to bring the suggestion list back.

packages/components/tests/e2e/specs/vsCodeExtension.spec.js
```diff
@@ -706,33 +706,30 @@ context('VS Code Extension', () => {
...
-      );
-      cy.get('.trace-view__search-input-suffix').click();
-      cy.get('.trace-view__search-input-element').should('have.value', '');
+      cy.get('.trace-filter__suffix').should('not.exist');
+      cy.get('.trace-filter__input').type('example');
+      cy.get('.trace-filter__input').should('have.value', 'example');
+      cy.get('.trace-filter__suffix').click();
+      cy.get('.trace-filter__input').should('have.value', '');
 
       let eventQuery = 'id:1 label:json id:3 id:15 id:99999 #link_to_edit_url';
```
dustinbyrne, line 715: Also worth testing:
- Hitting ESC clears focus from the text input and the suggestion list is hidden
- After submitting the form, the suggestion list is still visible
- Two suggestions are selected and the text input contains the expected string
- If we allow quoted literals containing whitespace, verify they return the expected results in a case where they'd otherwise return unexpected results if each word was treated as its own query (`"SELECT COUNT"` vs `SELECT COUNT`)
- Search terms are case-insensitive

packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -54,45 +54,19 @@
...
-              </div>
-            </div>
+            <v-trace-filter
+              ref="traceFilter"
+              :nodesLength="highlightedNodes.length"
+              :currentIndex="currentTraceFilterIndex"
+              :suggestions="eventsSuggestions"
+              @onChange="
+                (value) => {
+                  this.traceFilterValue = value;
```
dustinbyrne, line 64: Seems unrelated to this change, but there are some casing issues with the terms being searched for.

Searching for `SELECT` results in nothing. It should have many results.
Searching for `select` seems to yield the intended results, despite the actual text being `SELECT`.

https://user-images.githubusercontent.com/8737782/152196173-e30ce147-e2b5-42db-9f78-7ee392a93164.mp4

dustinbyrne (changes requested):
This suggestion list is great. The interaction is actually way better than I expected. I left some comments below.

packages/components/src/components/trace/TraceFilter.vue
sashahonchar: Fixed https://github.com/applandinc/appmap-js/pull/534/commits/17162e5790996cae1eac7ad34a92674287cc186f

packages/components/src/pages/VsCodeExtension.vue
sashahonchar: Fixed, full-text search is case-insensitive now https://github.com/applandinc/appmap-js/pull/534/commits/b591f836495be962b38cc25908767581ca2926d1

packages/components/src/components/trace/TraceFilter.vue
sashahonchar: Now picked event name will be present in filter input in double quotes https://github.com/applandinc/appmap-js/pull/534/commits/16d5e5d02a8ac984eb15a64001f3398079c10f74

packages/components/src/components/trace/TraceFilter.vue
sashahonchar: Fixed https://github.com/applandinc/appmap-js/pull/534/commits/16d5e5d02a8ac984eb15a64001f3398079c10f74

packages/components/src/components/trace/TraceFilter.vue
sashahonchar: @dustinbyrne it's already implemented.
If you type "POST /users", filtering algorithm will look up for event with that name, here is the code which responsible for that:
https://github.com/applandinc/appmap-js/blob/feat/trace-filter-auto-complete/packages/components/src/pages/VsCodeExtension.vue#L616-L626

packages/components/src/components/trace/TraceFilter.vue
sashahonchar: Done https://github.com/applandinc/appmap-js/pull/534/commits/ab513bf07e98b30335eb29f80bcc490490ab1d94

sashahonchar:
After commit https://github.com/applandinc/appmap-js/pull/534/commits/ab513bf07e98b30335eb29f80bcc490490ab1d94 there are some changes in filter input logic:
`id:1` or `label:json` input text will filter as it was
`"GET /admin"` will filter all events with name "GET /admin"
`%admin%` will filter out all events with word "admin" in it (it works as full-text search)

When user types regular string in filter input (ex. "admin") none events are filtered out, but suggestions list will be updated and contains only events with word "admin".

packages/components/tests/e2e/specs/vsCodeExtension.spec.js
sashahonchar: Added new e2e test for filter suggestions https://github.com/applandinc/appmap-js/pull/534/commits/ab513bf07e98b30335eb29f80bcc490490ab1d94

dustinbyrne:
> `%admin%` will filter out all events with word "admin" in it (it works as full-text search)

Does this string need to be wrapped with `%`? I'd expect `admin` to match anything containing the phrase without requiring additional syntax.

A quoted string should only be necessary if the search term contains a space. The quotes indicate to us that we should not consider whitespace as a delimiter of search terms.

kgilpin:
Yes I agree. the search term:

admin


should be a substring match.

Exact match:

"admin"

sashahonchar:
I made that change (with search term wrapped with `%`) to separate full-text search in events list and search in suggestions list.

For example, we have a list of 9000 events:
```
1. admin
2. admin1
3. admin2
....
9000. foobar
```

When user focuses into filter input - we show all suggestions.
But our user knows there is an event with name "foobar" somewhere in the end of the list and types string "foobar" in the input.

And here comes 2 cases:
1) old behavior
We make a full-text search, set all events with substring "foobar" in the name filtered and highlight them. But user expected to filter out only suggestions list and pick needed event by click. And in this case he will never see that event in suggestions list, because it's already selected and we exclude selected events from suggestions.
2) new behavior
We filter out suggestions list with substring "foobar", user clicks on needed event, we make event selected and show in filter input full name of that event in double quotes. To make full-text search through all events user can type `%foobar%`.

packages/components/src/components/trace/TraceFilter.vue
```diff
@@ -0,0 +1,391 @@
...
+      ) {
+        terms.pop();
+      }
+      this.filterValue = `${
+        terms ? terms.join(' ') : ''
+      } "${eventName}"`.trim();
+    },
+    onWindowClick(event) {
+      if (!getEventTarget(event.target, this.$refs.form)) {
+        this.showSuggestions = false;
```
dustinbyrne, line 205: I think this would best be done upon losing keyboard focus. Otherwise, a drag selection is likely to close the suggestion dropdown.

https://user-images.githubusercontent.com/8737782/153258684-1dc0c73c-b810-4e22-beb9-91df773fa078.mp4

sashahonchar: Seems there is a bug in Chrome with drag selection of input contents.
(In Firefox `click` event is not triggered when I finish selection outside input)
Fixed https://github.com/applandinc/appmap-js/pull/534/commits/1ee196b1ebf066702e08604d77911effb3f55dda

dustinbyrne:
Adding whitespace should indicate that the previous search term is complete. After adding whitespace to a non-quoted string I'd expect that the suggestion list returns back its default state, listing the full set of items. However, it doesn't update until another (non-whitespace) character is input.

https://user-images.githubusercontent.com/8737782/154325623-3f6b6e4d-d2ed-4a0d-9ddd-ce343c94a41a.mp4

dustinbyrne:
Regular search terms are not getting any results

https://user-images.githubusercontent.com/8737782/154327004-6778cbd8-3727-4a50-946c-f2c7aa48b1e0.mp4

packages/components/src/components/trace/TraceFilter.vue
```diff
@@ -0,0 +1,416 @@
...
+      }
+
+      return classList;
+    },
+    suggestionsList() {
+      let suggestions = this.suggestions
+        .map((e) => e.toString())
+        .filter((e, i, arr) => arr.indexOf(e) === i);
+
+      if (!this.filterValue.endsWith(' ')) {
```
dustinbyrne, line 138: This should also work when searching a quoted string, e.g. `"GET /admin"`. The space after `"GET` shouldn't trigger the list to display every item.

dustinbyrne:
There's still an issue with the case sensitivity when searching for partial text matches in SQL. Searching for `SELECT` yields no results, while `select` does. There should be a test case for this (https://github.com/applandinc/appmap-js/pull/534#discussion_r797795284).

dustinbyrne:
Typing a double quote throws an error. This should also be covered by a test case (https://github.com/applandinc/appmap-js/pull/534#discussion_r797795284).

image

sashahonchar:
Fixed. @dustinbyrne 
Now when user makes search through suggestions by event name (ex. `"GET /admin`), suggestions list will be filtered out.
https://github.com/applandinc/appmap-js/pull/534/commits/e31624e6664beba69ec91d0fa714fd7095328f44

packages/components/tests/e2e/specs/vsCodeExtension.spec.js
```diff
@@ -841,26 +841,28 @@ context('VS Code Extension', () => {
...
+      cy.get('.trace-filter__input').type('{esc}');
       cy.get('.trace-filter__suggestions').should('not.be.visible');
       cy.get('.trace-filter__input').should('not.be.focused');
 
-      cy.get('.trace-filter__input').focus().type('select');
+      cy.get('.trace-filter__input').type('select');
       cy.get('.trace-filter__suggestions').should('be.visible');
       cy.get('.trace-filter__suggestions-item')
         .first()
         .should('contain.text', 'SELECT');
```
dustinbyrne, line 860: This was previously passing even though a search for `SELECT` would fail. Can we have a test case which validates a search for `select` and `SELECT` return the same results?

dustinbyrne (approved):

PR #536, opened 2022-02-02
feat: Refactor SQL utilities into @appland/models

kgilpin:
Move SQL parsing and warnings into `@appland/models`, so that the code can be shared by packages such as `@appland/appmap` (CLI commands including `inspect`), and `@appland/scanner`.

Write SQL parse warnings into a file so that:

- They are easy to collect
- They don't pollute console output

Tests are passing in this repo, and also in `scanner` using `yarn link`.


packages/models/src/sql/log.js
```diff
@@ -0,0 +1,29 @@
...
+  if (!messages) return;
+
+  [...new Set(messages)].forEach((msg) => console.warn(msg));
+});
+
+// eslint-disable-next-line import/prefer-default-export
+export function warn(/** @type {string} */ message) {
+  if (SqlWarningFileName) {
+    open(SqlWarningFileName, 'w', (err, fd) => {
+      if (err || !fd) return;
```
dividedmind, line 17: This will silently drop messages. Either show a warning or fallback to console.

packages/models/src/sql/log.js
```diff
@@ -0,0 +1,29 @@
+import { open, write } from 'fs';
+
+let SqlWarningFileName = 'sql_warning.txt';
+let messages /** @type {string[]} */ = [];
+let writeMessage = (msg) => messages.push(msg);
+
+process.on('exit', () => {
```
dividedmind, line 7: Why not print messages as they come?

packages/models/src/sql/log.js
kgilpin: It shouldn't drop messages, because line 28 will add the message to the `messages` array, and they will all be dumped out on exit.

packages/models/src/sql/log.js
kgilpin: Well, this is what we were doing before, and it was very messy with all the messages getting printed to the console and inter-mingling with the console output. In any case, exit hook only takes effect if we couldn't open the sql_warning.txt file.

packages/models/src/sql/log.js
kgilpin: Turns out `@appland/models` is not allowed to have dependencies on Node.js, so I removed it.

kgilpin: This package will simply report errors to a function that the client can optionally provide.

dividedmind: > It shouldn't drop messages, because line 28 will add the message to the `messages` array, and they will all be dumped out on exit.

Ah, you're right I missed the lambda :)

packages/models/src/sql/log.js
dividedmind: I think it's preferable to print them as they come (even if it's messy), but if not, then I'd add some header to say what the dump is about and that it's been collected over the whole run. Otherwise it can be confusing if you don't already know what it is.

dividedmind (approved):
Definitely an improvement :)

dividedmind (changes requested):
Definitely an improvement re cleanup. I'm not sure about the error handling. 

It seems to try to sidestep normal exception handling, which I find quite weird. I'd much prefer this left as is and the redirecting of error messages handled in the client code.

kgilpin:
Honestly I think the client having to anticipate and catch exceptions is more weird. There’s not even a way in typescript to indicate what exceptions a function may throw (right?). At least this way, the contract is explicit.

dividedmind:
> Honestly I think the client having to anticipate and catch exceptions is more weird. There’s not even a way in typescript to indicate what exceptions a function may throw (right?). At least this way, the contract is explicit.

Could have at least thrown by default. As it stands, there's a default noop error callback which will silently discard errors and return null, which is a bit unexpected.

PR #544, opened 2022-02-10
fix: OpenAPI types

kgilpin:


apotterri (approved):
LGTM

packages/models/types/index.d.ts
```diff
@@ -156,14 +156,16 @@ declare module '@appland/models' {
...
+    readonly requestMethod?: string;
+    readonly requestContentType?: string;
+    readonly responseContentType?: string;
+    readonly route?: string;
+    readonly sqlQuery?: string;
     readonly fqid: string;
-    readonly previousSibling?: Event | null;
-    readonly nextSibling?: Event | null;
+    readonly previousSibling?: Event;
+    readonly nextSibling?: Event;
```
dustinbyrne, line 168: These were correctly typed. The value returned is either `string | undefined | null` or `Event | undefined | null`.

kgilpin:
What’s the difference between undefined and null in this context?

dustinbyrne:
> What’s the difference between undefined and null in this context?

Here's an example:
https://github.com/applandinc/appmap-js/blob/main/packages/models/src/event.js#L195-L212

We could always change the return type to `Event | undefined` by dropping the `return null`, but this isn't how it's implemented currently. When we were pure JavaScript the difference didn't matter much, you'd always check for a falsey value via `!event`. In TypeScript it's much more common to explicitly check `event === undefined` if you're guaranteed to have a type of `Event | undefined`. In this case if you were to do this, a `null` value can slip through.

packages/models/src/event.js
```diff
@@ -171,6 +182,14 @@ export default class Event {
     return null;
   }
 
+  get requestContentType() {
+    return Event.contentType(this.httpServerRequest, this.httpClientRequest);
+  }
+
+  get responseContentType() {
+    return Event.contentType(this.httpServerResponse, this.httpClientResponse);
+  }
```
dustinbyrne, line 191: These are also `string | null | undefined` given the definition of `Event.contentType`

PR #545, opened 2022-02-11
Don't replace SQL comments with a placeholder

dividedmind:
This leads to incorrect SQL, for example:
```sql
SELECT * FROM products 
   WHERE in_stock > 0
   -- 10 - toys
   AND category_id <> 10;
```
was normalized to
```sql
SELECT * FROM products 
   WHERE in_stock > ?
   ?
   AND category_id <> ?;
```


packages/models/tests/unit/sql.spec.js
```diff
@@ -84,6 +84,14 @@ describe('normalize SQL', () => {
       normalize(`SELECT users.*, a.* FROM users WHERE a.user_id = 'alice'`)
     ).toEqual(`SELECT users.*, a.* FROM users WHERE a.user_id = ?`);
   });
+
+  test('removes comments correctly', () => {
+    expect(
+      normalize(
+        `SELECT 4 AS random; -- guaranteed random, generated by fair dice roll`
```
kgilpin, line 91: Looks good! Do you also want to check that it does the right thing when the comment is embedded in the SQL string? And/or make sure it plays nicely with any other more complex SQL features?

dividedmind: > Looks good! Do you also want to check that it does the right thing when the comment is embedded in the SQL string? 

Good point, it didn't. I've fixed it.

> And/or make sure it plays nicely with any other more complex SQL features?

Not sure what else to test. Note this normalization isn't meant to be fool-proof, nor can possibly be without being a full parser.

As our SQL parser becomes more complete it might make sense to remove this heuristic and instead normalize the AST and regenerate the SQL query from that.

kgilpin: Ok. Also, I assume you tested this on the SQL that was provided by the user?

packages/models/tests/unit/sql.spec.js
```diff
@@ -84,6 +84,31 @@ describe('normalize SQL', () => {
...
+
+  // TODO: I would prefer if the whitespace was collapsed, but it's OK for now.
+  test('handles comments embedded in multiline SQL', () => {
+    expect(
+      normalize(`SELECT *
+-- this table contains users
+FROM users`)
+    ).toEqual(`SELECT *
+
+FROM users`);
```
kgilpin, line 110: @dividedmind the way it leaves this whitespace isn't super ideal but it's OK for now. Just thought I'd point this out. I am not sure if normalizing whitespace is part of what this code is supposed to do anyway. Like you said, parse and emit by traversing the AST will be a better approach.

kgilpin (approved):

packages/models/tests/unit/sql.spec.js
dividedmind: Not really, it isn't, as far as I can tell. We're not normalizing spaces elsewhere either, I think.

PR #547, opened 2022-02-13
fix: OpenAPI gen

kgilpin:
* Require minimum necessary version of `@appland/models`.
* Handle cases where `analyzeSQL` returns null.

Does this surprise you? It did me:

```
> typeof(null)
'object'
```


apotterri (approved):
Approved, though it would be better if this included a test.

PR #548, opened 2022-02-14
feat: Client retry

kgilpin:
Hooks to retry on client error and on HTTP 503.

The `retryOnError` and `retryOn503` functions are inserted into the request processing; they will re-invoke the service function if a retry-able error is detected.

## Example

https://github.com/applandinc/scanner/pull/109/commits/6eab58665521686b3b47c1fc8a0a260f388218b8


packages/client/test/retry.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+import { retryOn503, retryOnError } from '../src';
+import { RejectFunction, ResolveFunction } from '../src/retryHandler';
+
+const Message = { ok: true };
+const Handler = (resolve: ResolveFunction) => {
+  resolve(Message as any as IncomingMessage);
+};
+
+describe('retryOnError', () => {
+  it('handles EPIPE', async () => {
```
apotterri, line 12: There should be another example that checks `ECONNRESET.

packages/client/src/retryOnError.ts
```diff
@@ -0,0 +1,18 @@
+import { RejectFunction, ResolveFunction, RetryHandler } from './retryHandler';
+
+export default function retryOnError(
+  retryHandler: RetryHandler,
+  resolve: ResolveFunction,
+  reject: RejectFunction
+): (error: Error) => void {
+  return (error: Error): void => {
+    const { code } = error as never;
```
apotterri, line 9: What does `as never` do here?

apotterri (changes requested):
Needs one more test, I think, then should be good to go.

packages/client/src/retryOnError.ts
```diff
@@ -1,14 +1,17 @@
...
+};
+
 export default function retryOnError(
   retryHandler: RetryHandler,
   resolve: ResolveFunction,
   reject: RejectFunction
 ): (error: Error) => void {
   return (error: Error): void => {
-    const { code } = error as never;
+    const { code } = error as unknown as SystemError;
```
apotterri, line 13: Thanks. That's easier to understand.

apotterri (approved):

PR #550, opened 2022-02-16
feat: new design for Details panel

sashahonchar:
image
image


packages/components/src/components/DetailsPanelList.vue
```diff
@@ -96,44 +96,45 @@ export default {
...
 <style scoped lang="scss">
 .v-details-panel-list {
-  list-style-type: none;
-  padding: 0;
-  margin: 0 -1rem 1.5rem;
+  margin-bottom: 1rem;
+  border-radius: $border-radius;
+  padding: 0.5rem;
+  color: $base03;
+  background-color: rgba(0, 0, 0, 0.1);
```
dustinbyrne, line 103: This style is duplicated in many different components. Can it be moved into it's own component? The Figma contains guidelines for what should be componentized.

dustinbyrne (changes requested):
This is looking good. Regarding the style, there's a small border around these containers that's missing in the implementation:
image
image

The container around the list and the parameter value are the same.

packages/components/src/components/DetailsPanelList.vue
sashahonchar: I moved that unit test to `VsCodeExtension.spec.js` https://github.com/applandinc/appmap-js/pull/550/commits/e5bd8d3fdb624e9b23e5a67afd6d4970b55952f5

packages/components/tests/unit/DetailsPanelEvent.spec.js
```diff
@@ -6,24 +6,6 @@ import { store, SET_APPMAP_DATA } from '@/store/vsCode';
...
-      store,
-    });
-
-    wrapper.findComponent({ ref: 'viewSource' }).trigger('click');
-
-    const rootWrapper = createWrapper(wrapper.vm.$root);
-    const [[location]] = rootWrapper.emitted().viewSource;
-    expect(location).toBe(event.codeObject.location);
-  });
-
```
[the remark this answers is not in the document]
sashahonchar: I moved that unit test to `VsCodeExtension.spec.js` https://github.com/applandinc/appmap-js/pull/550/commits/e5bd8d3fdb624e9b23e5a67afd6d4970b55952f5

PR #553, opened 2022-02-25
fix: add dummy return events for filtered events

### LINKED ISSUE #552: Improve error handling in partial AppMap
When limiting root events to HTTP, nothing is displayed in the trace view:
image

Also while root events are limited to HTTP, `HelloWorldServer$HelloWorldHandler` is missing from the dependency map:
image

Toggling `Limit root events to HTTP` off displays the desired results in the dependency map:
image

However the HTTP server request is still missing from the trace view:
image

Below is the AppMap described. It's worth noting that two of the three events in the AppMap do not return:
- http_server_request - no return
- HelloWorldHandler#handle - no return
- HelloWorldHandler#sayHello - returns

It's likely that the missing returns here are causing issues. The AppMap client should be able to gracefully handle cases where such data is missing.
[hw.appmap.json.txt](https://github.com/applandinc/appmap-js/files/8110304/hw.appmap.json.txt)

sashahonchar:
Fixes #552

### DIFF
--- a/packages/components/src/pages/VsCodeExtension.vue
+++ b/packages/components/src/pages/VsCodeExtension.vue
@@ -560,7 +560,9 @@ export default {
         ),
         classMap: classMap.roots.map((c) => ({ ...c.data })),
         metadata: appMap.metadata,
-      }).build();
+      })
+        .addDummyReturns()
+        .build();
     },
 
     rootObjectsSuggestions() {
--- a/packages/components/src/stories/VsCodeExtensionNoReturnEvents.stories.js
+++ b/packages/components/src/stories/VsCodeExtensionNoReturnEvents.stories.js
@@ -1,0 +1,16 @@
+import VExtension from '@/pages/VsCodeExtension.vue';
+import hwScenario from './data/hw.json';
+
+export default {
+  title: 'Pages/VS Code/Extension No Return Events',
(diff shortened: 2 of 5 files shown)


dustinbyrne (changes requested):
It looks like the actual issue here was that `filteredEvents` is not calling `normalize` while building the AppMap. We don't want to remove the stack balancing logic from `normalize` because it'll affect downstream dependencies.

dustinbyrne:
This would be a good opportunity to add a test case. Using the AppMap from the issue, we can ensure that the proper events are rendered.

dustinbyrne:
What I mean to point out is that here we're not calling `normalize`:
https://github.com/applandinc/appmap-js/blob/4d503713019147636496f6d622b594a30ea3a45c/packages/components/src/pages/VsCodeExtension.vue#L557-L565

dustinbyrne:
You're right, I was overlooking the fact that we were copying events which have already been normalized.

Shouldn't this `filter` capture the return events? The events are seemingly present without filters, so they _do_ exist. Are we losing data here by adding return events? 
https://github.com/applandinc/appmap-js/blob/4d503713019147636496f6d622b594a30ea3a45c/packages/components/src/pages/VsCodeExtension.vue#L558-L560

PR #556, opened 2022-03-07
fix: Show orig_id

apotterri:


packages/models/src/event.js
```diff
@@ -412,8 +412,12 @@ export default class Event {
 
   get qualifiedMethodId() {
     const { definedClass, isStatic, methodId } = this;
-    if (!definedClass) return undefined;
-    return `${definedClass}${isStatic ? '.' : '#'}${methodId}`;
+    let ret = `[${this.orig_id}] `;
```
kgilpin, line 415: Does this only affect the UI or might it affect other code like event fingerprints as well?

packages/models/src/event.js
```diff
@@ -412,8 +412,12 @@ export default class Event {
...
+    ret += `${definedClass}${isStatic ? '.' : '#'}${methodId}`;
+    if (!isStatic) {
+      ret += `(${this.receiver.value.split('.').slice(-1)})`;
```
kgilpin, line 418: What’s this generally going to be?

packages/models/src/event.js
apotterri: I don't really know. I suspect it could have an impact on other code.

I'm not really proposing these changes as-is. I mostly wanted to put them someplace others could use them to reproduce what I put in my doc.

PR #558, opened 2022-03-28
fix: empty "normalized_path_info" of HTTP server request is present as empty array in AppMap

sashahonchar:
For some reason in some AppMaps `normalized_path_info` is preset as empty array.
According to [AppMap specification ](https://github.com/applandinc/appmap#http-server-request-call-attributes) it's not expected behavior, but to make plugin work with such AppMaps I've added type cast to `string`.

### DIFF
--- a/packages/models/src/event.js
+++ b/packages/models/src/event.js
@@ -161,14 +161,10 @@ export default class Event {
 
   get requestPath() {
     if (this.httpServerRequest) {
-      const normalizedPathInfo = this.httpServerRequest.normalized_path_info
-        ? this.httpServerRequest.normalized_path_info.toString()
-        : '';
-      const pathInfo = this.httpServerRequest.path_info
-        ? this.httpServerRequest.path_info.toString()
-        : '';
-
-      return normalizedPathInfo || pathInfo;
+      return (
+        this.httpServerRequest.normalized_path_info ||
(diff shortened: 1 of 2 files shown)


packages/models/src/event.js
```diff
@@ -161,10 +161,14 @@ export default class Event {
 
   get requestPath() {
     if (this.httpServerRequest) {
-      return (
-        this.httpServerRequest.normalized_path_info ||
-        this.httpServerRequest.path_info
-      );
+      const normalizedPathInfo = this.httpServerRequest.normalized_path_info
+        ? this.httpServerRequest.normalized_path_info.toString()
+        : '';
```
kgilpin, line 166: @dustinbyrne Sasha says this is an empty array sometimes. Is there a way to normalize this even farther upstream?

dustinbyrne: Events are mutated during normalization here: https://github.com/applandinc/appmap-js/blob/main/packages/models/src/appMapBuilder/index.js#L90-L114

This would be the best place to fix up any standardization issues.

sashahonchar: Fair enough. I moved that code to AppMap builder's `normalize()` method.  https://github.com/applandinc/appmap-js/pull/558/commits/7a74df0093d1279613b45096451b2012a6477c93

PR #559, opened 2022-03-31
feat: OpenAPI response types

kgilpin:
## Emit openapi `response` content schema

With this change, instead of just emitting MIME types for each status code, we also emit [response schema](https://spec.openapis.org/oas/v3.1.0#response-object) information when `return_value` is available on HTTP server response (this is a new optional field proposed in   https://github.com/applandinc/appmap/pull/29 and implemented in https://github.com/applandinc/appmap-ruby/pull/236).

## Schema incorporates multiple examples

Builds a merged schema when multiple examples are available for a route + status code.

Previously, the schema reflected just one example (which ever was seen first).

## Refactor / merge of code

Migrates the code to TypeScript and adds support for `http_client_request` (this is copied over from the `scanner` project). 

## Example

*Before*

```yaml
  "/api/api_keys":
    post:
      responses:
        '201':
          content:
            application/json: {}
```

*After*

```yaml
  "/api/api_keys":
    post:
      responses:
        '201':
          content:
            application/json:
              schema:
                type: object
                properties:
                  api_key:
                    type: string
```

## TODO

- [x] Replace `NilClass` with concrete type, when able.
- [x] Expand array and hash types into sub-properties

```
        "properties": [
          { "name": "mapset_id", "class": "NilClass" },
          { "name": "scenario_uuid", "class": "String" },
          { "name": "code_objects", "class": "Array" },
          { "name": "metadata", "class": "Sequel::Postgres::JSONBHash" },
          { "name": "labels", "class": "Sequel::Postgres::PGArray" }
        ]
```


packages/cli/src/openapi/response.ts
```diff
@@ -15,11 +16,31 @@ export default class Response {
...
     };
+
+    const schemata: Record<string, OpenAPIV3.SchemaObject> = {};
+
+    this.rpcRequests.forEach((rpcRequest) => {
+      const returnValue = rpcRequest.returnValue;
+      if (!returnValue) return;
+
+      if (rpcRequest.responseContentType) {
+        const mimeType = rpcRequest.responseContentType!.split(';')[0];
```
dustinbyrne, line 28: Just an FYI: because of the null/undefined check on the previous line, the type of `rpcRequest.responseContentType` should have changed to not include `null | undefined`. The `!` shouldn't be necessary.

dustinbyrne (approved):

PR #565, opened 2022-04-12
style: update background color

typaulhus:
The background of the IDE extensions was intended to match the colors used in VSCode. As it has expanded to other IDEs and to appland itself, the colors used look out of place. COlors should follow AppLand brand guidelines.

Styles the background color for the extension to use AppLand branding colors. 
Closes SCAN-241.

Before:
Screen Shot 2022-04-12 at 12 09 05 PM

After:
Screen Shot 2022-04-20 at 5 34 02 PM


Daniel-Warner-X (changes requested):
The filter text is nearly invisible. Please make sure that all text has an accessible level of contrast.

image

Daniel-Warner-X (approved):

PR #569, opened 2022-04-21
feat: "View source" link in details panel

sashahonchar:
Before: 
image

After:
image

Fixes MAP-253

### DIFF
--- a/packages/components/src/components/DetailsPanel.vue
+++ b/packages/components/src/components/DetailsPanel.vue
@@ -3,6 +3,13 @@
     <h3 class="details-panel__title">
       <AppMapLogo width="70" />
     </h3>
+    <div
+      class="details-panel__source"
+      v-if="!selectedObject && !selectedLabel && sourceLocationObject"
+    >
+      <h4 class="details-panel__source-title">Test case</h4>
+      <v-source-code-link :object="sourceLocationObject" :static="true" />
+    </div>
     <div class="details-panel__content">
       <div class="details-panel__notification">
         <slot name="notification" />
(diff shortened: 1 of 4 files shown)


packages/components/src/components/DetailsPanel.vue
```diff
@@ -109,6 +118,14 @@ export default {
...
       );
     },
+    sourceLocationObject() {
+      if (this.appMap.metadata.source_location) {
+        return {
+          location: this.appMap.metadata.source_location,
+        };
+      }
+      return null;
+    },
```
dustinbyrne, line 128: The source location cannot be static because we need the ability to modify the location based off the environment. For example, in an IDE the location may be a local file whereas the browser would require a link to the file in a remote repository, e.g. GitHub, GitLab, etc. Providing this as a static value does not allow us to specify this behavior.

dustinbyrne (changes requested):
See below comment. Please add a test to make sure this component is present when expected.

packages/components/src/components/DetailsPanel.vue
sashahonchar: @dustinbyrne Link to CodeObject location is always a link to file which will be opened in IDE, are we need another behavior for AppMap source location?
image

dustinbyrne: Only in some cases will the link need to be opened in an IDE. In other cases, we'd need to transform the file path to a location in GitHub. 

See below for an example of how we're using this event protocol to resolve locations in the AppMap data into something else like a URL:
https://github.com/applandinc/appmap-js/blob/main/packages/components/src/stories/support/resolvePath.js

sashahonchar: Thanks for explanation, it's done now. https://github.com/applandinc/appmap-js/pull/569/commits/7f462f1d4a4ba54450f3619c8d466d1ab55d0b12

dustinbyrne (approved):

PR #572, opened 2022-04-27
feat: Telemetry for 'record'

kgilpin:


packages/cli/src/cmds/runCommand.ts
```diff
@@ -2,28 +2,78 @@ import { AbortError, ValidationError } from './errors';
...
-export default async function runCommand(fn: () => Promise<any>) {
-  return fn().catch((err) => {
+async function invokeCommand() {}
+
+function flushTelemetry(exitCode: ExitCode, err?: Error) {
+  Telemetry.flush(() => {
+    if (err) {
+      Yargs.exit(exitCode, err);
+    } else {
+      exit(0);
```
apotterri, line 17: I believe this should be calling `Yargs.exit`, rather than `process.exit`. Yargs is kind of particular/weird about exiting the process.

This is the source for the `Yargs.exit`: https://github.com/yargs/yargs/blob/2d1136d303ea805685a973ded62f52efd49b78b9/lib/yargs-factory.ts#L724 , which expects that `err` can be undefined. IIRC, I've seen problems when a bad declaration for `Yargs.exit` gets picked up from somewhere else (maybe `@types`?).

apotterri (changes requested):

dustinbyrne (approved):

packages/cli/src/cmds/runCommand.ts
kgilpin: The problem for me is that `err` must be an Error, but I don't have one:

*@types/yargs*
```
        /** Manually indicate that the program should exit, and provide context about why we wanted to exit. Follows the behavior set by `.exitProcess().` */
        exit(code: number, err: Error): void;
```

apotterri: Yeah, that's the declaration I was thinking of. As you can see in the yargs source I linked to, it's wrong.

Can you pass a `new Error()`?

kgilpin:
But then it might think there's actually an error, when there isn't.

apotterri:
That's what I was wondering. Should be easy to try it out, no?

I'm pushing on this because the issues I've seen have been test failures that have been really hard to diagnose. Yargs expects you to use yargs.exit, and if you don't, weird stuff happens.

kgilpin:
I mean, I ran it both ways and nothing bad happened either time for me.

apotterri (approved):

PR #582, opened 2022-05-09
feat: scan --watch option

kgilpin:
`watch` mode watches the `appmap_dir` for changes to `mtime` files. These files are written by the AppMap indexer as AppMap JSON files are generated. So, the operation of this command is dependent on having the AppMap indexer running in watch mode on the workspace.

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -526,7 +526,7 @@ yargs(process.argv.slice(2))
         const cmd = new FingerprintWatchCommand(argv.appmapDir);
         await cmd.execute();
 
-        if (!argv.verbose) {
+        if (!argv.verbose && process.stdout.isTTY) {
           process.stdout.write('\x1B[?25l');
           const consoleLabel = 'AppMaps processed: 0';
           process.stdout.write(consoleLabel);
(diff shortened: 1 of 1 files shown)


packages/scanner/src/cli/scan/watchScan.ts
```diff
@@ -0,0 +1,59 @@
...
+    const pathTokens = fileName.split('/');
+    const appmapDir = pathTokens.slice(0, pathTokens.length - 1).join('/');
+    const appmapFile = [appmapDir, 'appmap.json'].join('.');
+    const reportFile = [appmapDir, 'appmap-findings.json'].join('/');
+
+    if (!(await promisify(exists)(appmapFile))) return;
+
+    const scanner = await buildScanner(true, this.options.configData, [appmapFile]).catch(
+      (error: Error) => {
+        throw new ValidationError(error.message + '\nUse --all to perform an offline scan.');
```
dustinbyrne, line 46: This conflicts with an error above stating not to use `--all` with `--watch`. As it stands now, there doesn't seem to be a way to run with `--watch` completely offline.
```
$ yarn start scan --watch -d ~/dev/applandinc/appland/tmp/appmap        
Unexpected status code: 401
```

dustinbyrne (approved):

packages/scanner/src/cli/scan/watchScan.ts
kgilpin: Because `reportAllFindings` is hard-coded to `true` here, it will always be offline. I'm not sure why it didn't work for you. I've removed the catch clause since it should never be needed.

PR #583, opened 2022-05-10
feat: Change filename defaults

kgilpin:
Default scanner config file becomes `appmap-scanner.yml` (was: no default besides the default sampleConfig). If `appmap-scanner.yml` is not available, the default sampleConfig is used, as before.

Default findings file becomes `appmap-findings.json` (was: `appland-findings.json`).

If `appmap.yml` contains *appmap_dir*, it will be used as the default dir for the `scan` and `ci` commands.


apotterri (approved):
LGTM

PR #584, opened 2022-05-10
feat: Extended finding info

kgilpin:
Add some new optional fields to Finding:

* `volatileData` Hash of named data that is associated with the finding, but is not stable across AppMaps.
* `participatingEvents` Hash of named events that are associated with the finding. The hash keys should be stable across AppMaps.
* `description` an extended Markdown description of the finding, with substitution of `participatingEvents` and `volatileData`.

Events in `participatingEvents` will be automatically added to `relatedEvents`.

`participatingEvents` is designed to support an upgrade to the finding hash algorithm, by specifically listing each significant event and its role in the finding.

`volatileData` is designed to be used in the description (along with other event data), to present a comprehensive description of the problem to the user in the UI, e.g. a code editor popup. 

## Example

`secret-in-log`

```ruby
      const participatingEvents: Record<string, Event> = { logEvent: event };
      if (match.generatorEvent) {
        participatingEvents.generatorEvent = match.generatorEvent;
      }
      const volatileData = { logMessage: value } as any;
      let description = `A [log event]({logEvent as url}) contains the message {value}.`;
      if (pattern instanceof RegExp) {
        volatileData.pattern = pattern;
        description += ` This message matches the regular expression {pattern as regexp}, which is likely to be a secret value.`;
      } else {
        volatileData.secret = pattern;
        description += ` This message contains the secret value {pattern}.`;
        if (match.generatorEvent) {
          description += ` This secret is generated by the function [{generatorEvent as prettyName}]({generatorEvent as url}).`;
        }
        description += ` Later in request processing, the secret is writen to the log.`;
      }
```

## TODO

- [x] Clone the participatingEvents to enable AppMaps to be GC-ed
- [ ] Add `stableData` field, including the command string from OS command exec.


packages/scanner/src/rules/lib/util.ts
```diff
@@ -15,6 +17,17 @@ export async function appmapDirFromConfig(): Promise<string | undefined> {
...
 
+export async function writeFileAtomically(fileName: string, data: any): Promise<void> {
+  const tmpDir = await mkdtemp(tmpdir() + '/');
+  const tempFileName = join(tmpDir, basename(fileName));
+  try {
+    await writeFile(tempFileName, data);
+    await rename(tempFileName, fileName);
+  } finally {
+    await rm(tmpDir, { recursive: true }).catch((err) => console.warn(err));
+  }
```
dividedmind, line 28: This is incorrect — `tmp` is likely on a different volume than the target file and the rename will fail with `EXDEV` (cross-device link not permitted). The temp file needs to be in the target directory. (You could put it in temp then move, but this won't be atomic and it will needlessly copy data around.)

Consider using a library like [write-file-atomic](https://www.npmjs.com/package/write-file-atomic) instead of rolling your own, this stuff is notoriously finicky.

packages/scanner/src/types.d.ts
```diff
@@ -111,6 +114,20 @@ interface Finding {
...
+  // Examples:
+  // {value} Locate the indicated key (value) in volatileData, and format as text.
+  // {value as regexp} Locate the indicated key (value) in volatileData, and format as a regexp.
+  // {logEvent as url} Locate the indicated key (logEvent) in participatingEvents, and format as a URL.
+  // {logEvent as prettyName} Locate the indicated key (logEvent) in participatingEvents, and format
+  // as a prettified event or code object name.
+  description?: string;
+  // Map of events by functional role name; for example, logEvent, secret, scope, etc.
+  participatingEvents?: Record<string, Event>;
+  volatileData?: Record<string, string | number>;
```
dividedmind, line 130: Perhaps these and other properties common with `MatchResult` should be refactored to a common base interface. It could maybe also be used for `buildFinding` argument (see the comment above).

dividedmind (changes requested):

kgilpin:
Are you interested in taking this over? I’m deep in VSCode.

dividedmind:
> Are you interested in taking this over? I’m deep in VSCode.

Sure.

kgilpin:
`participatingEvents` has been split out into its own PR. We can revive volatileData if we want to, at some later time. It doesn't feel critical for now.

kgilpin:
That would be great because I’m still worried that our hash system isn’t as
good as it needs to be.

For example if I defer a finding of a secret in log, I think it will ignore
all the future occurrences even if the secret comes from a different
function.

We need to think deeply about this. Or at least, you do :-) I’m super
absorbed in the extension right now.

PR #587, opened 2022-05-16
fix: Have scanner prune AppMaps before uploading

apotterri:
Limit the size of each uploaded AppMap to 40MB.


packages/scanner/src/cli/upload/pruneAppMap.ts
```diff
@@ -0,0 +1,13 @@
+import { AppMap as AppMapStruct, buildAppMap } from '@appland/models';
+
+const APPLAND_MAX_SIZE = parseInt(process.env.APPLAND_MAX_SIZE || '40960') * 1024;
```
kgilpin, line 3: All our env vars start with `APPMAP_` so let's go with something like `APPMAP_UPLOAD_MAX_SIZE`.

kgilpin (approved):

packages/scanner/src/cli/upload/pruneAppMap.ts
apotterri: Done.

PR #588, opened 2022-05-16
fix: Add missing openapi dependency on models

kgilpin:


dustinbyrne (approved):

dustinbyrne:
The `yarn.lock` will need to be committed to get the build passing.

kgilpin:
I'm sending that up now.

dividedmind:
Good catch, thanks!

PR #592, opened 2022-05-19
Rafal/map 288 scanner has telemetry

dividedmind:
Currently only one kind of event:
```json
{
  "name": "@appland/scanner/scan:completed",
  "measurements": {
    "appland.scanner.duration": 0.328,
    "appland.scanner.numChecks": 11,
    "appland.scanner.numAppMaps": 1,
    "appland.scanner.numFindings": 0
  },
  "properties": {
    "common.source": "@appland/scanner",
    "common.os": "linux",
    "common.platformversion": "4.14.103-1.pvops.qubes.x86_64",
    "common.arch": "x64",
    "appmap.cli.machineId": "cc0ce078a0cd9b82e778221c4cd3c574c17498b635d8cb41b5f9946616b1f1d9",
    "appmap.cli.sessionId": "7ab3b9c4376b1cea9858fac89517dc0f3ee18779423993ade7b2ba56edf69e03",
    "appland.scanner.version": "1.56.1",
    "appland.scanner.args": "/home/user/projects/appmap-js/packages/scanner/built/cli.js scan --all --app appland -f ../../../appland/Application_page_filter_options_can_be_selected.appmap.json",
    "appland.scanner.checks": "[authz-before-authn], [deserialization-of-untrusted-data], [exec-of-untrusted-command], [http-500], [logout-without-session-reset], [missing-content-type], [n-plus-one-query], [secret-in-log], [too-many-joins], [too-many-updates], [update-in-get-request]"
  }
}
```


kgilpin:
Nice! I'm sure a big part of this was just making the telemetry package
exist.

packages/scanner/src/cli/scan/singleScan.ts
```diff
@@ -70,4 +71,19 @@ export default async function singleScan(options: SingleScanOptions): Promise<vo
       numChecks / (elapsed / 1000.0)
     )} checks/sec)`
   );
+
+  const { checks } = scanResults;
+
+  Telemetry.sendEvent({
+    name: 'scan:completed',
+    properties: {
+      checks: checks.join(', '),
```
kgilpin, line 80: I think we may just want unique rule names here. Checks can include properties, which when represented as JSON and then joined into a comma-delimited string is going to be pretty nasty. Really we just want to know which rules the user has enabled, anyway.

kgilpin (commented):
This looks like a great start. Can we just add a telemetry event when there is an error? This is something that we do in the CLI and it's very useful. You should be able to find the handler function and put it there, similar to the CLI. Hopefully.

dividedmind:
> This looks like a great start. Can we just add a telemetry event when there is an error? This is something that we do in the CLI and it's very useful. You should be able to find the handler function and put it there, similar to the CLI. Hopefully.

Yeah, I started working on this, I just wanted to quickly get something going first :)

kgilpin:
OK, I think it would be best to ship it with the errors as well.

packages/scanner/src/cli/scan/singleScan.ts
dividedmind: `Check.toString()` doesn't represent it as JSON (see example above), but ok.

dividedmind:
@kgilpin I added error telemetry and changed the scan telemetry to only count and report unique rules.

kgilpin:
Approved. There’s a build error though.

kgilpin (approved):

PR #595, opened 2022-05-26
feat: install command accepts -d for target directory

kgilpin:
See: https://linear.app/appland/issue/MAP-318/install-command-accepts-d-argument


apotterri (approved):
LGTM.

PR #596, opened 2022-05-26
feat: Save recordings to appmap dir remote

kgilpin:
https://linear.app/appland/issue/MAP-323/save-remote-recordings-to-dollarappmap-dirremote


apotterri (approved):
Looks good, though it should probably include a test case, too.

PR #609, opened 2022-06-06
fix: User can configure extended RR URL

kgilpin:
Even if the server process can't be found in the process table, the user can elect to continue with remote recording and specify extended options for the URL connection.

### DIFF
--- a/packages/cli/src/cmds/record/prompt/continueWithRequestOptionConfiguration.ts
+++ b/packages/cli/src/cmds/record/prompt/continueWithRequestOptionConfiguration.ts
@@ -16,7 +16,7 @@ export default async function continueWithRequestOptionConfiguration(): Promise<
   UI.progress(
     `If you're sure it's running on that port, we can continue with extra configuration ` +
       `options that may enable me to connect to the agent. ` +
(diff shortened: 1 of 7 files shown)


packages/cli/src/cmds/record/prompt/continueWithRequestOptionConfiguration.ts
```diff
@@ -0,0 +1,33 @@
...
+  UI.progress(
+    `I can't find your AppMap server process on port ${
+      (await requestOptions()).port
+    }.`
+  );
+  UI.progress('');
+  UI.progress(
+    `If you're sure it's running on that port, we can continue with extra configuration ` +
+      `options that may enable me to connect to the agent. ` +
+      `Otherwise you start over by reconfiguring the host and port.`
```
apotterri, line 19: Seems like a word is missing here? "Otherwise you can", "should", "must"?

apotterri (changes requested):
One minor wording change, looks good otherwise. Thanks for doing this.

apotterri (approved):

kgilpin:
Yes, right. Should be “can”

PR #610, opened 2022-06-06
feat: Watch the scan config file

kgilpin:
In watch mode, scan command watches the config file and reloads the checks config when the file is modified.

### DIFF
--- a/packages/scanner/src/cli/scan/watchScan.ts
+++ b/packages/scanner/src/cli/scan/watchScan.ts
@@ -7,32 +7,54 @@ import { formatReport } from './formatReport';
 import { default as buildScanner } from './scanner';
 import { exists } from 'fs';
 import { promisify } from 'util';
+import { parseConfigFile } from '../../configuration/configurationProvider';
+import assert from 'assert';
 
 type WatchScanOptions = {
   appmapDir: string;
-  configData: Configuration;
+  configFile: string;
 };
 
 export class Watcher {
-  watcher?: chokidar.FSWatcher;
+  config?: Configuration;
+  appmapWatcher?: chokidar.FSWatcher;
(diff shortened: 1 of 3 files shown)


packages/scanner/src/cli/scan/watchScan.ts
```diff
@@ -7,46 +7,69 @@ import { formatReport } from './formatReport';
...
+    this.appmapWatcher.on('add', this.scan.bind(this)).on('change', this.scan.bind(this));
   }
 
   abort(): void {
-    if (!this.watcher) return;
+    if (!this.appmapWatcher) return;
 
-    this.watcher.close();
-    this.watcher = undefined;
+    assert(this.configWatcher);
```
apotterri, line 44: Why are you using an assertion here and in `scan`, rather than assuring that these properties are set?

It seems especially strange that a method called `abort` is asserting anything....

kgilpin: Isn’t an assertion a way to ensure that the properties are set?

kgilpin: close would be a better name

apotterri: > Isn’t an assertion a way to ensure that the properties are set?
Yeah. But, under what circumstances would these properties not be set? Would you be better off ensuring that they never occurred, or is that simply not possible? If you're going to keep the `assert`, it would be clearer if you also passed it a message: https://nodejs.org/api/assert.html#assertvalue-message

> Both watchers should either be defined or undefined.
One's getting tested with an `if`, but the other is tested with an `assert`. Why are they different?

> close would be a better name
But, even if it's named `close`, it seems like it should be more defensive, and just ignore the undefined property, rather than triggering an assertion.

PR #613, opened 2022-06-07
fix: Record command

kgilpin:
Fix two bugs in the record command:

* listAppMapFiles blows up with a `stat` error when the indexer is writing temp files into the index directory
* maxTime may be non-numeric, in which case it's not properly enforced

### DIFF
--- a/packages/cli/src/utils.js
+++ b/packages/cli/src/utils.js
@@ -2,8 +2,8 @@
 const { constants: fsConstants, promises: fsp } = require('fs');
 const { queue } = require('async');
 const glob = require('glob');
-const { join: joinPath } = require('path');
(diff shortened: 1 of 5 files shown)


packages/cli/src/utils.js
```diff
@@ -88,24 +88,8 @@ async function listAppMapFiles(directory, fn) {
...
-          await listAppMapFiles(path, fn);
-        }
-
-        if (file.endsWith('.appmap.json')) {
-          await fn(path);
-        }
-
-        return null;
-      })
+    (await promisify(glob)(`${directory}/**/*.appmap.json`)).map(fn)
```
kgilpin, line 92: @dividedmind is this OK? (simplified by using glob)

PR #622, opened 2022-06-14
Adding License

brikelly:


kgilpin (changes requested):
This seems questionable because not all the code in this repo is covered exactly by the MIT license. Scanners for example is MIT + Commons Clause. 

We may be better off telling the user to consult the package dirs for license terms.

Individual packages that use the MIT license (as indicated in the package.json) could have this file added to their package directory.

PR #631, opened 2022-06-18
fix: Indexing now works correctly even with long appmap filenames

dividedmind:


packages/cli/src/fingerprint/fingerprinter.js
```diff
@@ -125,7 +125,7 @@ class Fingerprinter {
 
     const tempAppMapFileName = joinPath(
       index.indexDir,
-      [basename(appMapFileName), 'tmp'].join('.')
+      [basename(appMapFileName, '.json'), 'tmp'].join('.')
```
kgilpin, line 128: Maybe we should just always name this temp file `appmap.tmp` or something.

kgilpin:
@dividedmind the test failures are all like this: 

```
➤ YN0000: [@appland/appmap]: Summary of all failing tests
➤ YN0000: [@appland/appmap]: FAIL tests/unit/inspect.spec.js
➤ YN0000: [@appland/appmap]:   ● Inspect › finds a named function
➤ YN0000: [@appland/appmap]: 
➤ YN0000: [@appland/appmap]:     ENOENT: no such file or directory, rename '/tmp/tmp-6089-zC5nq6OgOPQk/revoke_api_key/.15032385535000' -> '/tmp/tmp-6089-zC5nq6OgOPQk/revoke_api_key/canonical.classes.json'
➤ YN0000: [@appland/appmap]: 
```

PR #633, opened 2022-06-20
project picker updates

typaulhus:
Fixes applandinc/board#32


packages/components/src/pages/install-guide/EmptyState.vue
```diff
@@ -0,0 +1,182 @@
+<template>
+  <QuickstartLayout>
+    <section>
+      <header>
+        <h1>No projects found</h1>
+      </header>
+      <main>
+        <article class="empty-state" v-if="projects.length == 0">
```
dustinbyrne, line 8: There's a lot of double equality operators (`==`) in here. We typically use triple equality (`===`) across the board as it's more accurate and doesn't suffer from weird JavaScript-y behaviors. Can we make sure to move to triple equals before this gets merged?

ahtrotta: Sure thing!

packages/components/src/components/QualityIcon.vue
```diff
@@ -0,0 +1,28 @@
...
+  components: {
+    GoodIcon,
+    OKIcon,
+    BadIcon,
+  },
+
+  props: {
+    quality: {
+      type: String,
+      default: '1',
```
dustinbyrne, line 24: Is this the correct default? `'bad'`?

ahtrotta: yup, fixed

packages/components/src/components/quickstart/QuickstartLayout.vue
```diff
@@ -122,7 +122,7 @@ body {
 
 .qs-content {
   margin-bottom: 12px;
-  background: $color-tile-background;
+  background: $gray2;
```
dustinbyrne, line 125: The `$color-*` variables [defined here](https://github.com/applandinc/appmap-js/blob/main/packages/components/src/scss/_variables.scss#L114-L128) were being used to allow the user's VSCode theme to influence the styling used in these components. It's okay if we're moving away from that, but I want to make sure this change is made purposefully.

packages/components/src/pages/SingleProject.vue
```diff
@@ -0,0 +1,107 @@
...
+    />
+    <v-open-app-maps id="open-appmaps" :app-maps="appMaps" />
+    <v-investigate-findings
+      id="investigate-findings"
+      :scanned="hasFindings"
+      :num-findings="numFindings"
+      :project-path="path"
+    />
+  </v-multi-page>
+</template>
```
dustinbyrne, line 22: Do we need this file? We should try to limit ourselves to a single entrypoint for the install guide (`InstallGuide.vue`).

packages/components/src/pages/install-guide/EmptyState.vue
```diff
@@ -0,0 +1,182 @@
...
+            <div class="content">
+              <p>No projects were found in this workspace.</p>
+              <p>Open a project to see if it's ready to create AppMaps.</p>
+            </div>
+          </div>
+        </article>
+      </main>
+    </section>
+  </QuickstartLayout>
+</template>
```
dustinbyrne, line 22: It looks like this isn't in use?

packages/components/src/pages/install-guide/ProjectPicker.vue
```diff
@@ -186,4 +167,46 @@ p.note {
...
+      align-items: center;
+      gap: 0.5rem;
+      &.requirement-good {
+        svg {
+          color: $alert-success;
+        }
+      }
+    }
+  }
+}
```
dustinbyrne, line 187: It looks like this class has been refactored out.

packages/components/src/pages/install-guide/SingleProject.vue
```diff
@@ -0,0 +1,325 @@
...
+                <BadIcon />
+                <span class="bad">Supported Rails versions: 5.x, 6.x, 7.x</span>
+              </li>
+            </ul>
+          </article>
+        </template>
+      </main>
+    </section>
+  </QuickstartLayout>
+</template>
```
dustinbyrne, line 174: Can this file be deleted as well?

packages/components/src/components/install-guide/ProjectPickerRequirements.vue
```diff
@@ -0,0 +1,275 @@
...
+    },
+    testFrameworkQuality() {
+      if (
+        !(
+          this.project &&
+          this.project.testFramework &&
+          this.project.testFramework.score
+        )
+      )
+        return 'bad';
```
dustinbyrne, line 176: These guard conditionals are kind of gnarly, especially with Prettier formatting applied. My only advice would be to apply defaults to the potentially undefined values somewhere early on so you're guaranteed to always get a sane value.

Turning on optional chaining would also be beneficial, then we could boil this down to:
```js
if (!this.project?.testFramework?.score) return 'bad';
```

Something to consider for the future. I'm OK with leaving this as-is for now, it's pretty clear what's going on.

packages/components/src/components/install-guide/ProjectPickerRequirements.vue
```diff
@@ -0,0 +1,275 @@
+<template>
+  <div v-if="project">
```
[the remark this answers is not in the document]
ahtrotta: It seems like the default constructor for the prop isn't working, because `projects` is `undefined` when a project isn't selected in the table. But I'll change this so that we're not relying on the default constructor failing to work as expected.

packages/components/src/components/install-guide/ProjectPickerRequirements.vue
ahtrotta: Yup, optional chaining is much better, I just refactored it 😄

dustinbyrne (approved):

Issue #637, opened 2022-06-22
AppMap Labels get removed when running index command


petecheslock:
Using this project: https://github.com/land-of-apps/sample_app_6th_ed/tree/codespaces-devcontainer

Loom of issue: https://www.loom.com/share/29ac7e1efa4e4cf9b153bef0dfdfe07a

1) Install Appmap
2) Install latest code `yarn add --dev @appland/scanner @appland/appmap`
3) Record tests: `npx @appland/appmap record test -d /Users/petecheslock/repos/sample_app_6th_ed`
4) Open AppMap - See labels in an appmap
5) run an index `yarn run appmap index --appmap-dir .` - labels get removed. 
6) Install older appmap `yarn add --dev @appland/appmap@3.29.0`
7) Rerun tests to regenerate appmaps `npx @appland/appmap record test -d /Users/petecheslock/repos/sample_app_6th_ed`
8) Run index with version 3.29.0 `yarn run appmap index --appmap-dir .` Labels remain.  

[Diff between 3.29.0 and latest](https://github.com/applandinc/appmap-js/compare/@appland/appmap-v3.28.1...@appland/appmap-v3.29.0)

PR #643, opened 2022-06-23
fix: Emit missing data fields to JSON

kgilpin:
`CodeObject#toJSON` was omitting some fields from the source data. This PR ensures that all the expected fields are emitted. It does the same for `Event` as well, although this problem is not known to be occurring with Event (not to say it wasn't...). 

As a result, printing an AppMap object to JSON will result in a valid and complete AppMap. So, for example, raw AppMap data can be loaded, normalized, and then printed without loss of data.

Fixes https://github.com/applandinc/appmap-js/issues/637

### DIFF
--- a/packages/models/src/classMap.js
+++ b/packages/models/src/classMap.js
@@ -176,7 +176,6 @@ export default class ClassMap {
   }
 
   toJSON() {
-    // Don't write out code objects that were created during runtime
-    return this.roots.filter((obj) => !obj.dynamic);
+    return this.roots;
   }
 }
--- a/packages/models/src/util.js
+++ b/packages/models/src/util.js
@@ -425,3 +425,25 @@ export function resolveDifferences(arr1, arr2) {
 export function getRootEvents(eventArray) {
   return eventArray.filter((e) => e.isCall() && !e.parent);
 }
+
+export function transformToJSON(dataKeys, obj) {
+  const emptyLength = (value) => 'length' in value && value.length === 0;
+  const emptySize = (value) => 'size' in value && value.size === 0;
+  const empty = (value) =>
+    value === undefined ||
+    value === null ||
+    (typeof value === 'object' &&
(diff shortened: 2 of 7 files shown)


packages/models/src/classMap.js
```diff
@@ -176,7 +176,6 @@ export default class ClassMap {
   }
 
   toJSON() {
-    // Don't write out code objects that were created during runtime
-    return this.roots.filter((obj) => !obj.dynamic);
```
dividedmind, line 180: It's not entirely clear to me why this was removed.

packages/models/src/util.js
```diff
@@ -425,3 +425,25 @@ export function resolveDifferences(arr1, arr2) {
...
+    const value = obj[key];
+    if (empty(value)) {
+      // nop
+    } else if (value.constructor === Set) {
+      memo[key] = [...value];
+    } else {
+      memo[key] = value;
+    }
+    return memo;
+  }, {});
```
dividedmind, line 448: I feel this would be clearer as a `for ... of` loop.

packages/models/tests/unit/classMap.spec.js
```diff
@@ -4,6 +4,7 @@ import { CodeObjectType } from '../../src/codeObject';
 import scenario from './fixtures/user_page_scenario.appmap.json';
 import httpScenario from './fixtures/many_requests_scenario.json';
 import petClinicScenario from './fixtures/spring_petclinic.json';
+import assert from 'assert';
```
dividedmind, line 7: This import doesn't appear to be used.

dividedmind (approved):
Looks good, thanks!

packages/models/src/classMap.js
kgilpin: This filter was never working anyway because the it's actually `obj.data.dynamic`. In any case, when we write the `classMap.json`, we want to include all code objects, whether generated or not.

PR #649, opened 2022-06-28
Feat: Add Impact Domain to Finding

MaxDillon:
### Changes

- Changed finding to include a field impactDomain
    - Specifies 'Security' | 'Performance' | 'Maintainability' | 'Stability'
- Modified scanner code to pass this value down to finding creation
### Use-case

- To be displayed in VSCode Extension on investigate. findings page
- Replaces raw count of findings


apotterri (changes requested):
This needs at least one automated test before it gets merged. If you can replicate the manual test you did, that would be perfect.

MaxDillon:
- I added 4 tests, one for each of Security, Stability, Performance, Maintainability
    - All tests are passing
- Each test runs a scan with a rule that corresponds to its domain
    - Respectively: 
        - Decentralization of Untrusted Data
        - HTTP500
        - N+1 Query
        - Too Many Updates
- Question: Tests rely on appmaps that rule specific tests also rely on. Is this problematically overloading these rules / testcases? Will we want to change the domain of certain rules in the future?

kgilpin:
If we change the domain of a rule we will just update the test. Not a big
deal. Good question :-)

apotterri (approved):
Thanks for adding these tests, looks good now.

PR #654, opened 2022-06-29
node messaging for undetected version

typaulhus:
Fixes https://github.com/applandinc/board/issues/88
Add conditional messaging when node is not detected, with instructions to new page on docs (separate PR)

Mockup: 
image


apotterri (approved):
Looks good, thanks.

PR #668, opened 2022-07-10
fix: More reliable detection of appmap changes by scanner watch

dividedmind:
There was a race in the file watching code which could manifest on
Linux by scanner sometimes not noticing AppMap updates. The bug
has been fixed in chokidar, but it has not been merged upstream yet;
for this reason, use our fork in the meantime to fix this.

### DIFF
--- a/packages/scanner/package.json
+++ b/packages/scanner/package.json
@@ -65,7 +65,7 @@
     "applicationinsights": "^2.1.4",
     "async": "^3.2.3",
     "chalk": "^4.1.2",
-    "chokidar": "^3.5.3",
+    "chokidar": "applandinc/chokidar#fix/new-file-new-directory-race-on-linux",
     "cli-progress": "^3.11.0",
     "conf": "^10.0.2",
     "form-data": "^4.0.0",
--- a/packages/scanner/test/cli/scan.spec.ts
+++ b/packages/scanner/test/cli/scan.spec.ts
@@ -128,15 +128,7 @@ describe('scan', () => {
       });
     }
 
-    beforeEach(async () => {
-      tmpDir = await fsextra.mkdtemp(tmpdir() + '/');
-      secretInLogDir = join(
-        tmpDir,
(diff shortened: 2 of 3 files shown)


kgilpin:
What I don’t quite understand is: what’s in the fork we are using and will it be merged back to the upstream? Did you write it or was it someone else?

dividedmind:
Sorry, yes, I wrote it. I must have assumed the fact it's at applandinc/chokidar and the target branch git log would suggest as much.
It will hopefully get merged in and then we can go back to tracking the upstream: https://github.com/paulmillr/chokidar/pull/1228

kgilpin:
I mean, I figured I could track it down, but the best practice for review
is to provide all the information in the PR description that the reviewer
needs. At least links :-)

dividedmind:
> I mean, I figured I could track it down, but the best practice for review is to provide all the information in the PR description that the reviewer needs. At least links :-)

Fair point, thanks for calling me out :)
Here's the branch: https://github.com/applandinc/chokidar/tree/fix/new-file-new-directory-race-on-linux
And here's the diff to master: https://github.com/applandinc/chokidar/compare/master...applandinc:chokidar:fix/new-file-new-directory-race-on-linux

kgilpin (approved):

kgilpin:
@dividedmind is this going to address the case where we start a watch on a directory that doesn't exist yet? Probably not - but I'm just wondering, that's a big outstanding issue.

dividedmind:
> @dividedmind is this going to address the case where we start a watch on a directory that doesn't exist yet? Probably not - but I'm just wondering, that's a big outstanding issue.

No. Is that something you reckon I should work on?

kgilpin:
So far we are treating the project root dir as the appmap_dir and that's
working OK so far, though not optimal long term.

It's a good thing to do at some point but maybe not job 1. I'm going to be
away next week so if you took a crack at updating the deserialization rule
the way we described, that might be more useful.

PR #670, opened 2022-07-11
feat: Stable properties

kgilpin:
Add `Event.stableProperties`, which is meant to be a better foundation for an improved event/finding hash.

The concept is that all the data that identifies this event and is stable across AppMaps should be present here, and can be included in the hash. Volatile information such as SQL literals, parameter and return value literals, ... should be omitted.

### DIFF
--- a/packages/models/src/event.js
+++ b/packages/models/src/event.js
@@ -4,7 +4,7 @@ import {
   identityHashEvent,
   transformToJSON,
 } from './util';
-import hashEvent from './event/hash';
+import hashEvent, { abstractSqlAstJSON } from './event/hash';
 
 // This class supercedes `CallTree` and `CallNode`. Events are stored in a flat
 // array and can also be traversed like a tree via `parent` and `children`.
@@ -348,6 +348,13 @@ export default class Event {
     return this.$hidden.hash;
   }
 
+  get stableProperties() {
+    if (!this.$hidden.stableProperties) {
+      this.$hidden.stableProperties = this.gatherStableProperties();
+    }
+    return this.$hidden.stableProperties;
+  }
+
   get identityHash() {
     if (!this.$hidden.identityHash) {
       this.$hidden.identityHash = identityHashEvent(this);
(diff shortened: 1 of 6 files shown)


kgilpin:
@dividedmind can you take a look at this? I'm trying to build a foundation for a better finding hash function. Related:

https://github.com/applandinc/appmap-js/pull/669

packages/models/tests/unit/event.spec.js
```diff
@@ -93,6 +101,35 @@ describe('Event', () => {
...
+            '{"type":"statement","variant":"list","statement":[{"type":"statement","variant":"select","result":[{"type":"function","name":{"type":"identifier","variant":"function","name":"count"},"args":{"type":"identifier","variant":"star","name":"*"}}],"from":{"type":"identifier","variant":"table","name":"spree_stores"}}]}',
+        });
+      });
+      it('stableProperties with query parameters', () => {
+        const sql = {
+          database_type: 'sqlite',
+          sql: 'SELECT 1 FROM "spree_stores" WHERE id IN ( ?, ?, ?, 12 )',
+        };
+        const eventWithQueryParameters = new Event({
+          ...{},
```
dividedmind, line 117: Isn't this line a noop?

dividedmind (commented):
I had two specific comments. Also, I'm not sure if it makes sense merging this in without something that will actually use it -- I'm worried it'll end up being another piece of dead code. Should it stay in a branch until we're ready to actually integrate it with something?

kgilpin:
> I had two specific comments. Also, I'm not sure if it makes sense merging this in without something that will actually use it -- I'm worried it'll end up being another piece of dead code. Should it stay in a branch until we're ready to actually integrate it with something?

We definitely need something better than our current `hash` implementation, which, for example, hashes `http` for any HTTP request. It felt like breaking this down into steps would be easier than submitting a wholesale refactor of the hash algorithm.

packages/models/src/event.js
```diff
@@ -440,4 +447,59 @@ export default class Event {
...
+        route: this.route,
+        status_code: this.httpServerResponse.status,
+        content_type: this.responseContentType,
+      });
+
+    let properties;
+    if (sqlQuery) {
+      const sqlNormalized = abstractSqlAstJSON(sqlQuery, this.sql.database_type)
+        .split(/{"type":"variable"}(?:,{"type":"variable"})*/g)
+        .join(`{"type":"variable"}`);
```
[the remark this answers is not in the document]
kgilpin: It's common that a query with an IN clause will have different numbers of `?` across different invocations. But fundamentally, it's the same query and it should be normalized as such.

packages/models/tests/unit/event.spec.js
kgilpin: Yeah I guess so.

dividedmind:
> > I had two specific comments. Also, I'm not sure if it makes sense merging this in without something that will actually use it -- I'm worried it'll end up being another piece of dead code. Should it stay in a branch until we're ready to actually integrate it with something?
> 
> We definitely need something better than our current `hash` implementation, which, for example, hashes `http` for any HTTP request. 

We do :) But re http, my reading of https://github.com/applandinc/appmap-js/blob/7332a9fa49939c53c25e0849084935e2591cda73/packages/models/src/event/hash.js#L31 (and certainly the intent of the code) was that it should hash the route?

> It felt like breaking this down into steps would be easier than submitting a wholesale refactor of the hash algorithm.

I see what you mean, but I think the it should only get merged into the main branch once we have the whole thing down. I'd be more comfortable if this work would happen on a feature branch. We could then do part-wise pull requests (like this one) into the feature branch to keep review burden reasonable and then merge the whole thing once it's ready.

What do you think about it? If you're worried about the code becoming difficult to merge (or sync back with main), I'll be happy to take care of any gnarly issues that could happen in that vein :)

kgilpin:
I guess it's OK either way, but this has already been lingering on a branch
for a long time and so when that happens my inclination is to try and get
it merged (or discard it; but I don't think it deserves that...)

dividedmind:
> I guess it's OK either way, but this has already been lingering on a branch for a long time and so when that happens my inclination is to try and get it merged (or discard it; but I don't think it deserves that...)

I agree it shouldn't be discarded :) But we have enough dead code as it is, I don't think we should add to that. Can we move forward with the feature branch? Do you need any help or advice organizing this?

kgilpin:
I’m going to be on vacay until July 25.

PR #671, opened 2022-07-11
feat: Handle -d working directory arg

kgilpin:
`diff` and `inventory` commands will use default `appmap_dir` from *appmap.yml*.

This is a start on adding this behavior to all the commands.


apotterri (approved):
LGTM

PR #674, opened 2022-07-11
feat: Find by regexp

kgilpin:


packages/cli/src/search/types.d.ts
```diff
@@ -44,14 +44,13 @@ export interface SQLInfo {
...
 export interface CodeObject {
-  id: string;
   name: string;
   type: string;
-  database_type: string;
-  children: CodeObject[];
-  parent: CodeObject;
-  location: string; // Functions only
-  static: boolean; // Functions only
+  fqid: string;
```
kgilpin, line 49: Better to store `type` and `id` and then compute `fqid` on the fly by simply joining them together? Or drop `type` as a separate field?

PR #675, opened 2022-07-12
Write a pidfile on startup

apotterri:
**Needs  a rebase before merge!**

Used to coordinate with the vscode extension. Part of fixing applandinc/vscode-appland#399.


packages/cli/src/fingerprint/fingerprintWatchCommand.js
```diff
@@ -1,4 +1,6 @@
 const chokidar = require('chokidar');
+const fs = require('fs-extra');
+const path = require('node:path');
```
kgilpin, line 3: I think Dustin said we should use the old `path` instead of `node:path`? @dustinbyrne ?

packages/cli/src/fingerprint/fingerprintWatchCommand.js
```diff
@@ -10,26 +12,37 @@ class FingerprintWatchCommand {
...
+      this.fpQueue.process();
+      this.watcher = chokidar.watch(`${this.directory}/**/*.appmap.json`, {
+        ignoreInitial: true,
+      });
+      this.watcher
+        .on('add', this.added.bind(this))
+        .on('change', this.changed.bind(this))
+        .on('unlink', this.removed.bind(this))
+        .on('ready', () => {
+          const pidfile = path.join(this.directory, 'index.pid');
```
kgilpin, line 29: Should we be using os tmpdir instead? `this.directory` is often going to be the project dir.

When does the pid file get cleaned up?

apotterri: We probably don't want it to the be just tmpdir, because then we'll have collisions when multiple workspaces get opened. We could qualify it, but is there a reason not to write it to `this.directory`? And, isn't `this.directory` usually going to be `tmp` under the project dir, rather than the project dir itself? 

Also, I considered disabling creation of the pid file unless an env var is set. Do you think that would be better.

Thanks for pointing out the missing clean up. It was in some changes that got deleted.

kgilpin: We would need to pass the workspace name to the function, I think. `this.directory` is going to be the project dir, because we are not currently configuring the watch command to be constrained to `$appmap_dir`. One reason is that we don't reliably know 100% of the time where `$appmap_dir` is. And the other reason is that the watcher has undefined behavior if it's started up and `$appmap_dir` doesn't exist. 

We could use the system temp dir and then a pid file name that incorporates the VSCode process id and the workspace name/path (sanitized)? Or we could create a `tmp` dir in the project - that would be fine for many projects but kind of unexpected for Java projects that don't typically use `tmp`.

apotterri: Rather than complicate things, I'm just going to punt and only create the pid file if an env var is set. If it turns out that there's some other use than just testing, we can revisit.

kgilpin:
Right, makes sense.

packages/cli/src/fingerprint/fingerprintWatchCommand.js
```diff
@@ -7,29 +9,51 @@ class FingerprintWatchCommand {
...
+      const { pidfilePath, directory } = this;
+      const removePidfile = this.removePidfile.bind(this);
+      this.watcher
+        .on('add', this.added.bind(this))
+        .on('change', this.changed.bind(this))
+        .on('unlink', this.removed.bind(this))
+        .on('ready', () => {
+          if (pidfilePath) {
+            fs.outputFileSync(pidfilePath, `${process.pid}`);
+            process.on('exit', removePidfile);
```
kgilpin, line 42: Does this function need to be bound to this?

apotterri: You mean the arrow function being passed to `.on('ready'`, right? It does not need to be bound to `this`, because it's getting all the variables it needs from lines 33 & 34 above:
```
      const { pidfilePath, directory } = this;
      const removePidfile = this.removePidfile.bind(this);
```
I did this because of the way the functions are nested. But, looking at it again, it probably makes sense to factor the whole thing out into a separate method. I'll do that.

packages/cli/src/fingerprint/fingerprintWatchCommand.js
```diff
@@ -7,30 +9,41 @@ class FingerprintWatchCommand {
...
-      .on('unlink', this.removed.bind(this));
   }
 
-  close() {
-    this.watcher.close();
+  async close() {
+    await this.watcher.close();
+    this.removePidfile();
     this.watcher = null;
+    this.pidfilePath = null;
```
kgilpin, line 46: Would you want to do this in removePidFile?

apotterri: Oops, yeah.

kgilpin (approved):

PR #676, opened 2022-07-12
feat: trace-data command

kgilpin:
Add function `appmap trace-data [data-item]`.

* `data-item` a specifier of the form `code-object-type:code-object-id-or-pattern(param-name)`. Example: `function:psych/Psych.load(yaml)`.

data-item is parsed into a code object id or pattern, and a parameter name. So in the above example, we want to trace the `yaml` parameter to `Psych.load`. Example AppMap is attached.

This command finds all AppMaps that match the code object id or pattern. Then, for each AppMap, it finds the value of the specified parameter across all matching events. It also traces the value backwards through other events that have this value as a return value. And it finds events whose return value matches a parameter value of those events, and so on.

Once this set of "events of interest" is built, the command prints a new AppMap for each matching AppMap, which includes only the events of interest. 

## Example command invocation

```sh-session
$ yarn start trace-data -d ~/source/land-of-apps/sample_app_6th_ed "function:psych/Psych.load(yaml)"
```

## Example of the unfiltered AppMap

[Password_resets_password_reset_attack.appmap.json.zip](https://github.com/applandinc/appmap-js/files/9096567/Password_resets_password_reset_attack.appmap.json.zip)

## Example of a filtered AppMap showing "events of interest"

[Password_resets_password_reset_attack_data_trace.appmap.json.zip](https://github.com/applandinc/appmap-js/files/9096561/Password_resets_password_reset_attack_data_trace.appmap.json.zip)

### DIFF
--- a/packages/cli/smoketest.sh
+++ b/packages/cli/smoketest.sh
@@ -15,7 +15,7 @@ yarn add ./package.tgz
 yarn run appmap index --appmap-dir ruby
 yarn run appmap depends --appmap-dir ruby
 yarn run appmap inventory --appmap-dir ruby
-yarn run appmap openapi -d ruby -o /dev/null
+yarn run appmap openapi --appmap-dir ruby -o /dev/null
 
 
 cd "$PKGDIR"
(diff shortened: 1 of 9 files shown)


kgilpin:
@dividedmind can you update this to use your taint analysis instead?

dividedmind:
> @dividedmind can you update this to use your taint analysis instead?

Good idea, although the code for that is in scanner currently. Do you think I ought to move this command there?

PR #680, opened 2022-07-25
feat: Hash participating events and partial stack

kgilpin:
Adds a new `v2` hash algorithm based in part on the new `Event#stableProperties`.

Hash v2 first builds a canonical string of the finding, and then hashes it. Details of the canonical string algorithm are commented on the HashV2 class. Here's an example:

```
rule=n-plus-one-query
findingEvent.event_type=sql
findingEvent.sql_normalized={"type":"statement","variant":"list","statement":[{"type":"statement","variant":"select","result":[{"type":"identifier","variant":"star","name":"active_storage_attachments.*"}],"from":{"type":"identifier","variant":"table","name":"active_storage_attachments"},"where":[{"type":"expression","format":"binary","variant":"operation","operation":"and","left":{"type":"expression","format":"binary","variant":"operation","operation":"and","left":{"type":"expression","format":"binary","variant":"operation","operation":"=","left":{"type":"identifier","variant":"column","name":"active_storage_attachments.record_id"},"right":{"type":"variable"}},"right":{"type":"expression","format":"binary","variant":"operation","operation":"=","left":{"type":"identifier","variant":"column","name":"active_storage_attachments.record_type"},"right":{"type":"variable"}}},"right":{"type":"expression","format":"binary","variant":"operation","operation":"=","left":{"type":"identifier","variant":"column","name":"active_storage_attachments.name"},"right":{"type":"variable"}}}],"limit":{"type":"expression","variant":"limit","start":{"type":"variable"}}}]}
participatingEvent.commonAncestor.event_type=function
participatingEvent.commonAncestor.id=app_views_microposts__micropost_html_erb.render
participatingEvent.commonAncestor.raises_exception=false
stack[1].event_type=function
stack[1].id=app_views_users_show_html_erb.render
stack[1].raises_exception=false
stack[2].event_type=function
stack[2].id=ActionController::Instrumentation#process_action
stack[2].raises_exception=false
stack[3].event_type=http_server_request
stack[3].route=GET /users/{id}
stack[3].status_code=200
```

Note the following features:

1) Starts with the rule id
2) Followed with `commandEvent` - this is the HTTP server request, @command or @job that encloses the finding
3) Followed with `findingEvent` - this is the primary event of the finding
4) Followed by named `participatingEvent`s - these are other events that are instrumental to the finding, and should be invariant across duplicate findings. 

In this example, the `findingEvent` is the canonicalized form of the SQL query that's repeated N times. There is a participating event called `commonAncestor` which is the code function that contains all the N occurrences of the query.  

If the same query is found a different number M times within the same command and common ancestor, then that finding will have the same hash.

If the query is structurally different, or it's in a different command, or the queries have a different common ancestor, then the finding hash will be different. 

Hash V2 is generally more specific than Hash V1. For example, Hash V1 does not include the common ancestor of an N+1 query, nor does it include the command event. 

However, Hash V2 does do a better job of canonicalizing the query, because it collapses repeated query literals and query parameters.

Because Hash V2 includes more data about the finding, it will de-duplicate less aggressively.

### DIFF
--- a/packages/models/src/event.js
+++ b/packages/models/src/event.js
@@ -470,8 +470,11 @@ export default class Event {
     ) =>
       Object.assign(baseProperties, {
         route: this.route,
-        status_code: this.httpServerResponse.status,
-        content_type: this.responseContentType,
+        status_code:
+          this.httpServerResponse?.status ||
+          this.httpServerResponse?.status_code ||
+          this.httpClientResponse?.status ||
+          this.httpServerResponse?.status_code,
       });
 
     let properties;
@@ -493,12 +496,9 @@ export default class Event {
     } else {
       properties = {
         event_type: 'function',
-        name: this.qualifiedMethodId,
-        parameterNames: (this.parameters || [])
-          .map((p) => p.name)
-          .filter(Boolean)
-          .join(','),
-        returnValueType: this.returnValue?.class,
+        id: this.qualifiedMethodId,
+        raises_exception:
+          this.returnEvent.exceptions && this.returnEvent.exceptions.length > 0,
       };
     }
     return normalizeProperties(properties);
--- a/packages/models/tests/unit/event.spec.js
+++ b/packages/models/tests/unit/event.spec.js
@@ -37,9 +37,8 @@ describe('Event', () => {
     it('stableProperties', () => {
       verifyJSON(getTasksEvent.stableProperties, {
         event_type: 'function',
-        name: 'org/apache/zookeeper/book/recovery/RecoveredAssignments#getTasks',
-        parameterNames: '',
-        returnValueType: '',
+        raises_exception: false,
+        id: 'org/apache/zookeeper/book/recovery/RecoveredAssignments#getTasks',
       });
     });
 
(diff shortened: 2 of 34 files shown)


dustinbyrne:
Are we sure about including `commandEvent` every event hash where it's available? It makes sense in some cases such as the de-serialization of untrusted data, but for something like N+1 query it may be unnecessary. We've previously established that a finding should map 1:1 with a single defect in the code base, and this would seem to violate that.

My only concern is that we may no longer be able to say with certainty how many unique findings exist. If I introduce an N+1 query into a common code path executed by all my controllers, I'd generate N findings, though there's really only one defective point of code.

Everything else looks good!

kgilpin:
@dustinbyrne agreed - and this is the kind of feedback I'm looking for. I am also concerned about cases where we are overly aggressive about marking findings as duplicates when they really aren't. For example, the same normalized query could be an N+1 query in more than one place. The v1 hash will miss that and think that there is only one distinct finding. v2 will probably not miss it, because the command event is included in the hash - but it will be maybe overly reluctant to merge findings.

Philosophically, which is worse - a false positive or a false negative? False positives (findings considered unique, when they are actually duplicates) have the nice behavior that when the developer fixes the problem, all the affected findings will disappear. So at least, when you fix the problem, all the findings that should disappear will disappear. 

If a developer `Defers` a finding whose hash is too non-specific, then other distinct and separate findings will also be considered Deferred by the system, and essentially will never be surfaced.

dustinbyrne (approved):

PR #687, opened 2022-08-02
feat: add video link to runtime analysis instruction page

ahtrotta:
Fixes applandinc/board#121

image


packages/components/src/pages/install-guide/InvestigateFindings.vue
```diff
@@ -90,30 +90,27 @@
...
+              href="https://appland.com/docs/analysis#demo"
             >
-              <SlackLogo />
-              <v-button
-                label="Join our Slack for early access"
-                class="cta-button b-0"
+              <img
+                class="video-thumbnail"
+                alt="runtime analysis video thumbnail"
+                src="../../assets/runtime_analysis_video_thumbnail.png"
```
dustinbyrne, line 111: I believe this is the first `png` we're adding to the distribution. Can you confirm this will properly be added to the bundle that's published to npm?

dustinbyrne (changes requested):
There's an alignment issue with the video thumbnail in the Storybook, can you take a look?
https://5fe18baf66f2280021e634d0-ncxxxvnzhz.chromatic.com/?path=/story/pages-vs-code-install-guide-pages--investigate-findings-disabled

packages/components/src/pages/install-guide/InvestigateFindings.vue
ahtrotta: When I locally generate the `dist/index.js` file and use it to load the Extension Host VS Code instance the `png` shows up okay:

image

Should I do anything else to confirm?

ahtrotta:
I fixed the rendering issue in Storybook and tested the branch locally. Let me know if there's anything else!

dustinbyrne (approved):

PR #698, opened 2022-08-08
feat: Scan interactive

kgilpin:
Run the scan command with `-i` flag:

[asciicast](https://asciinema.org/a/xucu7IiuPe1bMdMl2TVpRhSc4)

A useful way to use this:

* Start a JavaScript debug console for the scanner project
* Run the scanner in interactive mode
* Run the scan once
* Set a scanner breakpoint at a particular code object or event
* When the scanner breaks at that code object or event, set a breakpoint in the scanner JavaScript code
* Continue the scanner
* Debug the scanner rule

### DIFF
--- a/packages/client/src/reportJson.ts
+++ b/packages/client/src/reportJson.ts
@@ -1,6 +1,6 @@
 import { IncomingMessage } from 'http';
 
-export default function <T>(response: IncomingMessage): Promise<T> {
+export default function reportJson<T>(response: IncomingMessage): Promise<T> {
   return new Promise((resolve, reject) => {
     let data = '';
     response.setEncoding('utf8');
--- a/packages/scanner/.gitignore
+++ b/packages/scanner/.gitignore
@@ -14,6 +14,7 @@ tmp
 
 # Misc
 .tmp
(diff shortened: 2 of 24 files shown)


dividedmind:
Neat! However I think it'd be more comfortable as a readline-based command line interface. Arrowing through menus can get inconvenient pretty quickly and makes repeating previous steps difficult.

BTW, is this meant to be useful for end users, or just for developing rules? If the latter I think it ought to be noted in the help text to avoid confusion. Perhaps even turn it off in the release?

kgilpin:
It's meant to be useful for developing rules.

Are you saying you just want to press a letter command instead of arrowing
through a menu?

dividedmind:
> It's meant to be useful for developing rules. Are you saying you just want to press a letter command instead of arrowing through a menu?

Not a letter command. Full command, for example, in the session from the movie, that'd be:
```
> run
> break ActionDispatch::Request::Session#[]
> continue
> continue
> continue
> eval event
> eval event.codeObject.fqid
> hints
> continue
> quit
> quit
```
with full readline editing and history.

PR #700, opened 2022-08-09
feat: Ignore schema info tables in too-many-joins rule

kgilpin:


dustinbyrne (approved):
[`src/rules/tooManyJoins.ts`](https://github.com/applandinc/appmap-js/blob/0995838ad64044e1bf7f461c77cfbb72f73d6bbf/packages/scanner/src/rules/tooManyJoins.ts) seems like it should be deleted now, but otherwise this looks good to me.

PR #703, opened 2022-08-10
Add scans for unverified signature and `none` algorithm JWT usage

### LINKED ISSUE #701: Scanner rules for safe JWT usage patterns

dustinbyrne:
Resolves #701


packages/scanner/package.json
```diff
@@ -56,7 +56,7 @@
   },
   "dependencies": {
     "@appland/client": "^1.3.0",
-    "@appland/models": "^1.16.1",
+    "@appland/models": "workspace:^1.18.1",
```
kgilpin, line 59: Is this right?

packages/scanner/src/rules/jwtAlgorithmNone.ts
```diff
@@ -0,0 +1,49 @@
...
+  where(event: Event): boolean {
+    return event.labels.has('jwt.encode');
+  }
+}
+
+class JwtAlgoritmNone implements Rule {
+  public readonly id = 'jwt-algorithm-none';
+  public readonly title = "JWT 'none' algorithm";
+  public readonly impactDomain = 'Security';
+  public readonly enumerateScope = false;
```
kgilpin, line 40: I think you want this to be true, so that it will feed you every event in the scope.

packages/scanner/src/rules/jwtUnverifiedSignature.ts
```diff
@@ -0,0 +1,49 @@
...
+    const matches = new Array<MatchResult>();
+
+    const children = [...event.descendants()];
+    while (children.length) {
+      const child = children.pop() as Event;
+      if (child.labels.has('jwt.signature.verify')) {
+        verified = true;
+        break;
+      }
+      children.push(...child.descendants());
```
kgilpin, line 17: Descendants of the child should already be in the array you’ve build. Also you can avoid all this array building pushing and popping by using event navigator.

packages/scanner/src/rules/jwtUnverifiedSignature.ts
```diff
@@ -0,0 +1,49 @@
...
+
+class JwtUnverifiedSignatureLogic implements RuleLogic {
+  matcher(event: Event): MatchResult[] | undefined {
+    let verified = false;
+    const matches = new Array<MatchResult>();
+
+    const children = [...event.descendants()];
+    while (children.length) {
+      const child = children.pop() as Event;
+      if (child.labels.has('jwt.signature.verify')) {
```
kgilpin, line 13: Is this sure to occur only in a descendent or could it also occur in a preceding event on the same receiver object?

packages/scanner/src/rules/jwtUnverifiedSignature.ts
```diff
@@ -0,0 +1,49 @@
...
+  where(event: Event): boolean {
+    return event.labels.has('jwt.decode');
+  }
+}
+
+class JwtUnverifiedSignature implements Rule {
+  public readonly id = 'jwt-unverified-signature';
+  public readonly title = 'Unverified signature';
+  public readonly impactDomain = 'Security';
+  public readonly enumerateScope = false;
```
kgilpin, line 39: Same comment about enumerate scope

packages/scanner/test/scanner/jwtAlgorithmNone.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+import rule from '../../src/rules/jwtAlgorithmNone';
+import { scan } from '../util';
+
+describe('JWT Unverified Signature', () => {
+  const check = new Check(rule);
+
+  it('identifies cases where a JWT is encoded and the algorithm is none', async () => {
+    const { findings } = await scan(check, 'JWT.encode_alg_none.appmap.json');
+    expect(findings).toHaveLength(1);
+    expect(findings.map((f) => f.event.id)).toEqual([1]);
```
kgilpin, line 11: With enumerate scope as false I think this only works because the encode event is the first event in the AppMap.

packages/scanner/test/fixtures/appmaps/JWT.decode_no_validation.appmap.json
```diff
@@ -0,0 +1 @@
...ue":"{alg=\u003eHS256}","object_id":2560,"size":1}},{"id":22,"event":"call","thread_id":2000,"defined_class":"String","method_id":"unpack1","path":"\u003cinternal:pack\u003e","lineno":280,"static":false,"parameters":[{"name":"fmt","class":"String","object_id":2480,"value":"m0","kind":"req"}],"receiver":{"class":"String","object_id":2580,"value":"eyJrZXkiOiJ2YWx1ZSJ9"}},{"id":23,"event":"return","thread_id":2000,"parent_id":22,"elapsed":4.000961780548096e-06,"return_value":{"class":"String","value":"{\"key\":\"value\"}","object_id":2600}},{"id":24,"event":"call","thread_id":2000,"defined_class":"JSON::Ext::Parser","method_id":"parse","path":"JSON::Ext::Parser#parse","static":false,"parameters":[],"receiver":{"class":"JSON::Ext::Parser","object_id":2620,"value":"#\u003cJSON::Ext::Parser:0x00007feb30943b80\u003e"}},{"id":25,"event":"return","thread_id":2000,"parent_id":24,"elapsed":5.999580025672913e-06,"return_value":{"class":"Hash","value":"{key=\u003evalue}","object_id":2640,"size":1}},{"id":26,"event":"return","thread_id":2000,"parent_id":17,"elapsed":0.000700000673532486,"return_value":{"class":"Array","value":"[{key=\u003evalue}, {alg=\u003eHS256}]","object_id":2660,"size":2}}]}
```
kgilpin, line 1: I’m curious how you generated these fixture AppMaps?

kgilpin (changes requested):

packages/scanner/package.json
dustinbyrne: Yes, this will allow us to resolve to local workspace packages instead of the published `npm` package. Once it's published, it'll reference `^1.18.1`.

kgilpin:
Can you commit it with the other fixture code?

packages/scanner/src/rules/jwtUnverifiedSignature.ts
dustinbyrne: I've updated the conditions to identify a signature verification match. An event which is labelled `jwt.signature.verify` must either:
- Also contain the `jwt.decode` label
```ts
function decode(token: string): Jwt | undefined {
  const [header, payload, signature] = token.split('.').map(t => btoa(t));

  switch (header.alg.toLowerCase()) {
  case 'hs256'
    // The token signature is validated here inline during decoding.
    break;
  
  case 'none':
  default:
    return undefined;
  }
  
  return new Jwt(header, payload, signature);
}
```
- Have the same receiver (by looking at object ID)
```ts
class Jwt {
  constructor(protected readonly token: string) {
    this.decode();
    this.verifySignature();
  }

  decode() {
    // Parse token segments
  }
  
  verifySignature() {
    // Make sure the signature matches
  }
}
```
- Contain a parameter which matches the decoded token string
```ts
class Jwt {
  constructor(protected readonly token: string, key: string) {
    this.decode(token);
    Signature.verify(this.algorithm, key, this.signingInput, this.signature);
  }
}

class Signature {
  static verify(algorithm: Algorithm, key: string, signingInput: string, signature: string) {
    // Perform the signature verification
  }
}
```
- Contain a parameter which matches the receiver of the decode event
```ts
class Jwt implements Signed {
  constructor(protected readonly token: string) {
    this.decode(token);
    Signature.verify(this);
  }
  
  public get signingInput(): string { return `${this.header}.${this.payload}`; }
  public get algorithm(): Algorithm { return Algorithms[this.decodedHeader.alg]; }
}

class Signature {
  static verify(token: Signed, key: string) {
    // Perform the signature verification
  }
}
```

dustinbyrne: It won't look at preceding events, only events which follow the decode. The reason being the token header must first be decoded to retrieve the encryption algorithm, so it's unlikely that the validation can come first.

dustinbyrne:
Added 👍

kgilpin (approved):

PR #704, opened 2022-08-12
ci: Build and release pre-built binaries to GitHub Releases

### LINKED ISSUE #702: AppMap CLI and Scanner have stable URLs to download the latest binary

dustinbyrne:
Build and release `scanner` and `appmap` binaries to GitHub Releases during semantic release.

Resolves #702

### DIFF
--- a/.releaserc.js
+++ b/.releaserc.js
@@ -46,7 +46,12 @@ module.exports = {
     ],
     '@semantic-release/git',
     join(__dirname, 'ci/yarnPublish.js'),
-    '@semantic-release/github',
+    [
+      '@semantic-release/github',
+      {
+        assets: 'release/*.tar.gz',
+      },
+    ],
   ],
-  extends: 'semantic-release-monorepo'
+  extends: 'semantic-release-monorepo',
 };
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -118,7 +118,7 @@
     "assets": [
       "package.json",
       "**/openapi-template.yaml",
-      "appmap.html"
+      "built/appmap.html"
     ],
     "outputPath": "dist"
   }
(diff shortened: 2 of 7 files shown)


packages/cli/src/cmds/open/serveAndOpenAppMap.ts
```diff
@@ -13,7 +13,7 @@ export default async function serveAndOpenAppMap(appMapFile: string) {
   const builtDir = join(__dirname, '../../../built');
   const baseDir = (await exists(builtDir))
     ? builtDir
-    : join(__dirname, '../../..');
+    : join(__dirname, '..', '..');
```
dustinbyrne, line 16: I think this was an artifact left over from when the `built` directory contained `src`. The directory structure has since flattened by one level (i.e., the compiled contents of `src` are now directly added to `built`). When run via `npx`, the former directory is used (`'../../../built'`).

packages/cli/package.json
```diff
@@ -118,7 +118,7 @@
     "assets": [
       "package.json",
       "**/openapi-template.yaml",
-      "appmap.html"
+      "built/appmap.html"
```
dustinbyrne, line 121: I thought it was cleaner to retain the same directory structure that we publish to npm. The fact that we had to add this file to the root directory seemed more like a bug.

.releaserc.js
```diff
@@ -46,7 +46,12 @@ module.exports = {
     ],
     '@semantic-release/git',
     join(__dirname, 'ci/yarnPublish.js'),
-    '@semantic-release/github',
+    [
+      '@semantic-release/github',
+      {
+        assets: 'release/*.tar.gz',
+      },
```
dustinbyrne, line 53: This _should_ attempt to release any `.tar.gz` file under the `release` directory for any given package. I don't imagine it'll fail if `release` doesn't exist (or is empty), but I'm unsure.

PR #709, opened 2022-08-18
feat: Track specific untrusted data in unsafe deserialization rule

dividedmind:
Unsafe deserialization scanner rule now tracks untrusted data through the call tree instead of relying instead on trusted sources
being tagged and only checking the call sequence.

This means that common false positives are less likely (such as for example arising from unmarshaling trusted on-disk data, as is frequent in some frameworks).

Since specific pieces of data are now tracked some false negatives can also be avoided, for example when multiple tainted data inputs are used but only some of them are sanitized.

The disadvantage is that the reliability of the rule now depends on completeness of the AppMap. If there is a data transformation
that is not captured it's invisible to the rule and will result in failure to associate it with the tracked untrusted data.

Note also that currently the tracked tainted data is limited to HTTP request messages, although there is no reason this cannot
be expanded in the future.


packages/scanner/doc/rules/deserialization-of-untrusted-data.md
```diff
@@ -30,9 +47,12 @@ that executes code shortly after deserialization.
 
 ### Resolution
 
-If you can guarantee that you are using unsafe deserialization in a safe way, but it's not possible
-to obtain the raw data from a function labeled `deserialize.sanitize`, you can wrap the
-deserialization in a function labeled `deserialize.safe`.
+Consider if you the library you're using offers a safe deserialization function variant that you can
```
kgilpin, line 50: Typo here.

kgilpin:
Is it worth preserving the old behavior, but issuing the finding with a lower confidence level?

dividedmind:
> Is it worth preserving the old behavior, but issuing the finding with a lower confidence level?

Not sure if it's worth it. I'm on the fence. Wasn't it too prone to false positives to be practically useful?
Also, we don't have any confidence indicators yet.

kgilpin (approved):

PR #710, opened 2022-08-22
Create a ticket when "record test" fails

apotterri:
**Please don't merge this, it needs a rebase.**

Offer the user the chance to create a ticket when `record test` has test command failures, or fails to create AppMaps.

Fixes applandinc/board#128. Fixes applandinc/board#106.

### DIFF
--- a/packages/cli/src/cmds/record/recordContext.ts
+++ b/packages/cli/src/cmds/record/recordContext.ts
@@ -5,6 +5,22 @@ import {
   readSetting,
   TestCommand,
 } from './configuration';
+export class RecordProcessResult {
+  constructor(
+    private _env: Record<string, string>,
+    public command: string,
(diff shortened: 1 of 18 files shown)


packages/cli/src/cmds/record/recordContext.ts
```diff
@@ -7,7 +7,6 @@ import {
 } from './configuration';
 
 export type RecordProcessResult = {
-  env: Record<string, string>;
```
apotterri: This changes originally included the process environment in the telemetry events, but this seemed risky. Let me know if you think I should put it back.

kgilpin:
Understood.

Let’s send any APPMAP variables that don’t also include KEY

kgilpin (approved):

PR #718, opened 2022-09-03
fix: Skip bad files when running scanner on a directory

dividedmind:
When running scanner in a directory scan mode, an error processing
an appmap (eg. due to malformation or too large size) will no
longer crash the process preventing scanning other maps.
A warning will be printed instead and the process continues.

If all appmaps fail, the process still exits with an error status.
If at least one succeeds, the process exits with success.

Note this change doesn't affect the scenario where you explicitly
list files using --appmap-file argument. Any failure there will
still cause the process to fail.

### DIFF
--- a/packages/scanner/src/cli/scan.ts
+++ b/packages/scanner/src/cli/scan.ts
@@ -8,6 +8,7 @@ import RuleChecker from '../ruleChecker';
 import { Finding } from '../types';
 
 import AppMapIndex from '../appMapIndex';
+import assert from 'node:assert';
 
 type Result = {
   appMapMetadata: Record<string, Metadata>;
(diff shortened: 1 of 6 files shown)


packages/scanner/src/cli/scan.ts
```diff
@@ -8,6 +8,7 @@ import RuleChecker from '../ruleChecker';
 import { Finding } from '../types';
 
 import AppMapIndex from '../appMapIndex';
+import assert from 'node:assert';
```
kgilpin, line 11: I think this format `node:assert` doesn’t work with all the versions we want to support.

kgilpin (approved):

packages/scanner/src/cli/scan.ts
dividedmind: Seems to be supported since v14: https://nodejs.org/docs/latest-v14.x/api/esm.html#esm_node_imports

PR #730, opened 2022-09-13
Check project dependencies before adding appmap

ahtrotta:
Fixes applandinc/board#139

Check dependencies before adding AppMap. The check varies with the package manager:

* `npm install --dry-run`
* `yarn install --immutable`
* `bundle check --dry-run`
* `pip install -r requirements.txt --dry-run`
* `poetry install --dry-run`

Maven and Gradle are less straightforward, so those will be a separate batch.

### DIFF
--- a/packages/cli/src/cmds/agentInstaller/install-agent.ts
+++ b/packages/cli/src/cmds/agentInstaller/install-agent.ts
@@ -5,6 +5,7 @@ import {
   ChildProcessError,
   InvalidPathError,
   ValidationError,
+  UserConfigError,
 } from '../errors';
 import { endTime, prefixLines, verbose } from '../../utils';
 import AgentInstallerProcedure from './agentInstallerProcedure';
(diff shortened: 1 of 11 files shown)


packages/cli/src/cmds/agentInstaller/install-agent.ts
```diff
@@ -85,6 +86,26 @@ class InstallerError {
...
       return true;
+    } else if (this.error instanceof UserConfigError) {
+      UI.error(
+        `${chalk.red('!')} Error in project configuration:\n\n${
+          this.error?.message
+        }`
+      );
+
+      Telemetry.sendEvent({
+        name: 'install-agent:user_config_error',
```
apotterri, line 97: We seem to have several different conventions for naming telemetry event, but snake-case is currently the least-used. I'd vote for making this either `user-config-error` or `userConfigError`, with a preference for the former.

packages/cli/tests/unit/agentInstall/installCommand.spec.ts
```diff
@@ -4,7 +4,7 @@ import tmp from 'tmp';
 import sinon from 'sinon';
 import inquirer from 'inquirer';
 import Telemetry from '../../../src/telemetry';
-import yargs from 'yargs';
+import yargs, { check } from 'yargs';
```
apotterri, line 7: Doesn't look like `check` is used?

apotterri (changes requested):
A couple of very minor changes to make, looks good on the whole. Thanks!

apotterri:
It's easier to review changes based on comments if you push them as a separate commit: force pushing throws away the history (of course), so there's no way to navigate from the comment to the change.

This is a problem addressed by `fixup` commits: https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---fixupamendrewordltcommitgt

ahtrotta:
Yeah that makes sense. So I would add each commit after the review with the message `fixup!` and then squash all of the commits when it's time to merge?

apotterri:
Yeah. This helps with the squashing: https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---autosquash

apotterri (approved):

PR #721, opened 2022-09-07
feat: Report largest maps and slowest functions after record command runs

symwell:
Report the functions to which AppMap recording adds the most overhead - not the functions that take the most time.
Also report the largest AppMap files.

Depends on https://github.com/applandinc/appmap-js/pull/731

There's a corresponding [pr in appmap-ruby](https://github.com/applandinc/appmap-ruby/pull/277) that saves instrumentation data in AppMaps

- Added a `stats` command that reports the 5 functions with the most AppMap overhead and the 5 biggest AppMaps that are over 0.5MB.
- If the functions overhead information is missing, the stats omit the functions overhead report.
- Added a `--limit` parameter to select how many results to display.
- Added a `--json` parameter to format results as JSON, similar to https://github.com/applandinc/appland-cli#quickstart
- Function parameters aren't displayed because execution times are aggregated: a function could be called from multiple callers with different parameters.
- Changed the `record` command to call the `stats` command at the end.

Regarding the performance consideration with the index, this pr didn't attempt to use the index because I didn't understand what [codeObjectId](https://github.com/applandinc/appmap-js/blob/main/packages/cli/src/cmds/inspect/inspect.ts#L36) should refer to.

Based on https://docs.google.com/document/d/1Qa90Hf4LBiIubSIB8mlo1yhKe4uz2uSTuO8VNjNIfoA/edit#
Fixes https://github.com/applandinc/board/issues/140

(1) Called from `record` command:
```
$ rm -rf  /home/test/src/sample_app_6th_ed/tmp/appmap > /dev/null 2>&1; npm start record test -- -d /home/test/src/sample_app_6th_ed

> @appland/appmap@3.34.0 start
> ts-node src/cli.ts record test -d /home/test/src/sample_app_6th_ed

The following test commands were provided by your AppMap configuration (appmap.yml):

APPMAP=true DISABLE_SPRING=true bundle exec rails test

? Use these test commands? Yes
? Enter the maximum time (in seconds) to allow test cases to run (-1 to run forever): 30
Running test command: APPMAP=true DISABLE_SPRING=true bundle exec rails test
Started with run options --seed 15012


Progress: |====================================================================================================================================================================|

Finished in 3.27253s
67 tests, 282 assertions, 0 failures, 0 errors, 0 skips

   ╭───────────────────────────────────────────────────────────────────────────────────────╮
   │                                                                                       │
   │           Success! There are now 67 AppMap files in directory 'tmp/appmap'.           │
   │                                                                                       │
[...]

PR #731, opened 2022-09-13
feat: Add elapsedInstrumentationTime, path, threadId, parentId in Event

symwell:
Add fields needed to use the Event type properly when showing functions with highest AppMap overhead in https://github.com/applandinc/appmap-js/pull/721.

I think this change needs to be merged and deployed first so that importing `@appland/models` will provide these fields.

I attempted to add these fields directly in https://github.com/applandinc/appmap-js/pull/721 but importing from a local module with the following didn't work.
```
import buildAppMap from '../../../../models/src/appMapBuilder/index';
```
It gave an error that said to unmark `models` as a module, which is probably not what we want.  I applied this change locally so I can develop in the meantime.
```
--- a/packages/models/package.json
+++ b/packages/models/package.json
@@ -4,7 +4,6 @@
   "description": "",
   "module": "dist/index.js",
   "main": "dist/index.cjs",
-  "type": "module",
   "publishConfig": {
     "access": "public"
   },
```

Addresses https://github.com/applandinc/board/issues/140

### DIFF
--- a/packages/models/src/event.js
+++ b/packages/models/src/event.js
@@ -318,6 +318,10 @@ export default class Event {
     this.$hidden.parent = value;
   }
 
+  set id(value) {
+    this.$hidden.id = value;
+  }
+
   set path(value) {
     this.callEvent.$hidden.path = value;
   }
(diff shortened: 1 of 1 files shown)


kgilpin:
I think the trick is to run the build + watch command (I think it's `yarn
run watch`) in the "packages/models" folder as you develop in
"packages/cli".

That way you can develop in both "packages" at once. This is the
primariy reason that all our JS code is in the same repo. You can ask
Dustin for more details.

PR #734, opened 2022-09-16
fix: users with older versions of pip can install

ahtrotta:
Fixes applandinc/board#159

### DIFF
--- a/packages/cli/src/cmds/agentInstaller/pythonAgentInstaller.ts
+++ b/packages/cli/src/cmds/agentInstaller/pythonAgentInstaller.ts
@@ -3,13 +3,14 @@
 import os from 'os';
 import { join } from 'path';
 import AgentInstaller from './agentInstaller';
-import CommandStruct from './commandStruct';
+import CommandStruct, { CommandOutput } from './commandStruct';
 import { exists } from '../../utils';
 import chalk from 'chalk';
 import { run } from './commandRunner';
 import { getOutput } from './commandUtil';
 import EncodedFile from '../../encodedFile';
 import { UserConfigError } from '../errors';
+import semver from 'semver';
 
 const REGEX_PKG_DEPENDENCY = /^\s*appmap\s*[=>~]+=.*$/m;
 // include .dev0 so pip will allow prereleases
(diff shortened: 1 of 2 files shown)


symwell (commented):
The two `describe` sections seem similar with the Python version being different and passing `--dry-run`.  I wonder if it helps to combine them into a single function.

packages/cli/tests/unit/agentInstall/installCommand.spec.ts
```diff
@@ -516,16 +516,170 @@ packages:
...
+          stdout:
+            'pip 22.2.2 from /home/user/.pyenv/versions/3.10.7/lib/python3.7/site-packages/pip (python 3.10)',
+          stderr: '',
+        });
+      };
+
+      const pythonVersion = (cmdStruct: CommandStruct) => {
+        expect(cmdStruct.program).toEqual('python');
+        expect(cmdStruct.args).toEqual(['--version']);
+        return Promise.resolve({ stdout: 'Python 3.10.7', stderr: '' });
```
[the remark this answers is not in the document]
ahtrotta: No, this is just for testing. My understanding is that that function is generating output for this call: `const version = await getOutput('python', ['--version'], this.path)` when the tests are run.

packages/cli/src/cmds/agentInstaller/pythonAgentInstaller.ts
```diff
@@ -121,11 +121,29 @@ export class PipInstaller extends PythonInstaller {
...
-      this.path
-    );
+    let commandArgs = ['install', '-r', this.buildFile];
+    let major: number;
+    let minor: number;
+    let patch: number;
+
+    try {
+      const pipVersionOutput = await getOutput('pip', ['--version'], this.path);
+      const VERSION_REGEX = /(?<major>\d+)\.(?<minor>\d+).(?<patch>\d+)/;
```
apotterri, line 131: This regex isn't a great match the version specifiers that pip uses: https://pypi.org/project/pip/#history .

We're already including the `semver` package in the cli. And, while pip's version specifiers don't follow semver, they're pretty close. You should be able to something like this (untested, make sure it works):
```
const supportsDryRun = semver.satisfies(semver.coerce(pipVersionOutput, {loose:true}, ">=22.2.0")
```

apotterri:
> The two `describe` sections seem similar with the Python version being different and passing `--dry-run`. I wonder if it helps to combine them into a single function.

Yeah, seems like there's a bunch of duplicated code here (especially `testE2E`). You should move it into a surrounding scope (which you may need to create with another `describe`).

apotterri (changes requested):

packages/cli/src/cmds/agentInstaller/pythonAgentInstaller.ts
ahtrotta: Good catch, thank you. All of the pip versions have a major and a minor version, so I just changed the regex to look for a major and a minor version. I think this should work, but let me know if you still think I should use the `semver` package.

ahtrotta: I just noticed that certain pre-release versions have a `b` in them (i.e. `22.1b1` pre-release). This would still work with the current regex, giving `major` as `22` and `minor` as `1`.

apotterri: I still think you should switch to `semver`. We don't need to add code that duplicates its functionality.

ahtrotta: Okay, I'll change it

apotterri (approved):
Looks good, thanks. Don't forget to rebase.

PR #736, opened 2022-09-18
Add pipenv support

apotterri:
Add support for Pipenv as an installer type.

This allows easy installation in https://github.com/chicagopython/chipy.org, so.... Fixes applandinc/board#149.


packages/scanner/package.json
```diff
@@ -77,6 +77,7 @@
     "minimatch": "^3.0.4",
     "octokat": "^0.10.0",
     "openapi-diff": "^0.23.5",
+    "ora": "~5",
```
apotterri, line 80: Prior to this change, we picked up `ora` because it's a dependency of `inquirer`. We're currently using v8.2.0 of `inquirer`, which is locked to v5.4.1 of `ora`.

.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: node_js
 os: linux
 dist: focal
-node_js: [14, 18]
+node_js: [14, 16, 18]
```
apotterri, line 4: Seems like we should be testing all the LTS versions.

packages/cli/src/cmds/agentInstaller/pythonAgentInstaller.ts
```diff
@@ -95,6 +95,54 @@ export class PoetryInstaller extends PythonInstaller {
...
+    const ret = await exists(this.buildFilePath);
+    return ret;
+  }
+
+  async initCommand(): Promise<CommandStruct> {
+    return new CommandStruct('pipenv', ['run', 'appmap-agent-init'], this.path);
+  }
+
+  async checkCurrentConfig(): Promise<void> {
+    // pipenv doesn't seem to offer a way to check the config, short of doing an install.
```
kgilpin, line 121: So, should we do an install then? I think that's how this feature should work.

kgilpin (commented):
I do think we should do a "preflight" install before we add appmap.

packages/cli/src/cmds/agentInstaller/pythonAgentInstaller.ts
```diff
@@ -109,7 +157,7 @@ export class PipInstaller extends PythonInstaller {
   }
 
   async available(): Promise<boolean> {
-    return await exists(this.buildFilePath);
+    return exists(this.buildFilePath);
```
ahtrotta, line 160: This should also have `await`

packages/cli/src/cmds/agentInstaller/pythonAgentInstaller.ts
ahtrotta: Yeah, that's what I do for `yarn` because there isn't a dry run feature

apotterri: Let me see what happens when I do that. If both installs take the same amount of time (i.e. this doubles the time to install the agent), does it still make sense to do this?

ahtrotta (changes requested):
As Kevin mentioned, I would just run the dependency installation for `checkCurrentConfig`. Also, there are a couple of spots that I commented on where you want `await` for an `async` function.

packages/cli/src/cmds/agentInstaller/pythonAgentInstaller.ts
apotterri: @kgilpin I confirmed that doing this does take twice as long (if the user hasn't previously installed the dependencies).

Should I still make this change?

kgilpin (approved):

PR #739, opened 2022-09-21
feat: Manual installation instructions exist for supported languages

### LINKED ISSUE #735: Manual installation instructions are included in the install guide

dustinbyrne:
[Try the updated project picker in the browser](https://5fe18baf66f2280021e634d0-dynhbribjb.chromatic.com/iframe.html?id=pages-vs-code--install-guide&args=&viewMode=story)

This PR:
- Lessens restrictions on which projects are considered supported. A project only must have an agent available for the language used.
- Alters language to depict which AppMap features are available to the project, instead of describing the project's own frameworks.
- Adds an 'automated installation' button.
- Adds manual installation instructions for supported projects (currently Ruby and Python).
- The installation button is will now retain focus within the viewport as projects are selected.

image

Resolves #735


ahtrotta (approved):
I made one little change to pass the project language to the backend so it knows when to use `npx`. Bu otherwise it looks good. I ran the tests locally and tested it on Windows and Linux and everything seems to work correctly.

PR #740, opened 2022-09-21
Update Github URLs to use getappmap

petecheslock:
Joachim mentioned that many of the changes in https://github.com/getappmap/appmap-intellij-plugin/pull/134 were part of this repository, if we updated here then he could pull latest changes.  
If there are folders I should ignore just let me know and i can remove those from this diff.


brikelly (approved):
I've browsed all the changes and they all look like correct substitutions.

brikelly (approved):
Revised changes look good too. Thanks!

PR #743, opened 2022-09-23
Poll files for changes in addition to watching

### LINKED ISSUE #742: appmap_dir is re-scanned every few seconds
Chokidar sometimes fails to notice new map files in appmap_dir, which is a serious UX issue given how our product needs to index and analyze them the moment they appear. Most fs-watching libraries - including chokidar - recommend a full rescan periodically in case they don't pick up a file, so we should do that too. 

Given that we'll only be scanning in a single directory and doing so non-recursively, it should be very lightweight and fast to rescan. We will initially start with a 5 second period between full rescans.

dividedmind:
Fixes #742


kgilpin (commented):
There’s no provision for node_modules or .gitignore. It might also be good to ignore directories in tmp that aren’t the appmap dir. WDYT?

dividedmind:
> There’s no provision for node_modules or .gitignore. It might also be good to ignore directories in tmp that aren’t the appmap dir. WDYT?

Yeah, but it was also true for the watch path, so it feels like a separate issue and should go in a later PR. I just went for the simplest fix now, there's definitely room for improvement.

kgilpin:
> > There’s no provision for node_modules or .gitignore. It might also be good to ignore directories in tmp that aren’t the appmap dir. WDYT?
> 
> Yeah, but it was also true for the watch path, so it feels like a separate issue and should go in a later PR. I just went for the simplest fix now, there's definitely room for improvement.

I’m sensitive about this because extensions have a reputation for killing performance on peoples machine. 

It seems like we need to make an effort to avoid scanning the worst directories.

dividedmind:
> I’m sensitive about this because extensions have a reputation for killing performance on peoples machine.

Our extension can sometimes indeed be pretty bad in this regard :) 
But I don't think directory scanning is anywhere near the worst offender in this regard.

Anyway, I'll keep working on it.

dustinbyrne (approved):
Looks good. This change also fixes Windows compatibility 👍

kgilpin (commented):
These optimizations look useful. I couldn't quite bring myself to push the Merge button since they don't have much in the way of testing.

dividedmind:
> These optimizations look useful. I couldn't quite bring myself to push the Merge button since they don't have much in the way of testing.

Well, if the optimizations broke anything there are tests that would break. I'm not sure what other tests you would like to see.

kgilpin:
I guess I would need to understand which tests were testing the affected
code, because it would be very non-obvious to me.

I need my job as reviewer to be as easy as possible - especially if I make
a mistake it's going to break all our indexing and scanning.

dividedmind:
> I guess I would need to understand which tests were testing the affected code, because it would be very non-obvious to me. I need my job as reviewer to be as easy as possible - especially if I make a mistake it's going to break all our indexing and scanning.

Fair. Although it sounds like you just don't have confidence in our test suite :) Which is understandable, I guess — if I had full confidence, I wouldn't have needed to verify if breaking these changes does indeed break tests!

kgilpin:
Caching is hard and easy to get wrong. Really hard to tell by looking at a
test suite if it’s going to catch a mistake in caching.

PR #745, opened 2022-09-26
feat: detect architecture mismatch errors in Ruby

ahtrotta:
Fixes getappmap/appmap-ruby#286

When there is an architecture mismatch, then the following is displayed to the user:

image

### DIFF
--- a/packages/cli/src/cmds/agentInstaller/agentInstallerProcedure.ts
+++ b/packages/cli/src/cmds/agentInstaller/agentInstallerProcedure.ts
@@ -6,6 +6,8 @@ import { AbortError } from '../errors';
 import { run } from './commandRunner';
 import UI from '../userInteraction';
 import AgentProcedure from './agentProcedure';
+import Telemetry from '../../telemetry';
+import CommandStruct from './commandStruct';
 
 export default class AgentInstallerProcedure extends AgentProcedure {
   async run(): Promise<void> {
@@ -56,7 +58,7 @@ export default class AgentInstallerProcedure extends AgentProcedure {
       }
     }
 
-    UI.status = 'Installing the AppMap agent...';
+    UI.status = 'Installing AppMap...';
 
     await this.installer.checkCurrentConfig();
     await this.installer.installAgent();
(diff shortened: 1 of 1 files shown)


packages/cli/src/cmds/agentInstaller/agentInstallerProcedure.ts
```diff
@@ -66,7 +68,36 @@ export default class AgentInstallerProcedure extends AgentProcedure {
...
+        stdout = initCommandOutput.stdout;
+      } catch (e) {
+        const error = e as Error;
+
+        if (
+          this.installer.name === 'Bundler' &&
+          error?.message.includes('but is an incompatible architecture')
+        ) {
+          await run(new CommandStruct('gem', ['uninstall', 'appmap', '-x', '--force'], this.path));
+          await run(new CommandStruct('bundle', [], this.path));
```
dustinbyrne, line 84: If I'm understanding the problem correctly, I don't believe we can automate this for them. This error occurs because `node` is run via x64 emulation and thus the child `bundle` process is as well, causing the `appmap` dependency to be built for the wrong architecture.

The resolution `gem uninstall appmap; bundle` works for the user because there's no x64 `node` process to trigger emulation.

dustinbyrne (changes requested):

packages/cli/src/cmds/agentInstaller/agentInstallerProcedure.ts
ahtrotta: Won't it depend on if the user is running VS Code natively or emulated? If they're running VS Code natively, then when we run the `bundle` command for them, it will know the architecture is `arm64`

dustinbyrne: As @dividedmind explained it, this is due to users running an x64 version of node. The command that's failing is essentially the same as what we're running to fix it: 

https://github.com/getappmap/appmap-js/blob/main/packages/cli/src/cmds/agentInstaller/rubyAgentInstaller.ts#L68

In regard to the architecture of VS Code, I'm not sure if it's an issue. It does seem totally possible that an (old) x64 version of VS Code would spawn terminals through Rosetta, but I'm unaware of how prevalent this may be.

packages/cli/src/cmds/agentInstaller/agentInstallerProcedure.ts
```diff
@@ -56,33 +58,66 @@ export default class AgentInstallerProcedure extends AgentProcedure {
...
+        error?.message.includes('but is an incompatible architecture')
+      ) {
+        await run(new CommandStruct('gem', ['uninstall', 'appmap', '-x', '--force'], this.path));
 
-    UI.success(successMessage.join('\n'));
+        const incompatibleArchitectureMessage = [
+          '\n',
+          chalk.bold.red('AppMap Installation Error!'),
+          '',
+          'Please run the following command in your terminal, then re-run the installer:',
```
dustinbyrne, line 102: Removing the gem is a good idea! One small nit-pick: this language looks like it needs to be updated.

dustinbyrne (approved):
:+1:

dividedmind:
BTW, I just discovered https://www.unix.com/man-page/osx/1/arch/ — it seems that maybe it'd work to run arch --arch arm64e bundle.

PR #747, opened 2022-09-27
fix: Poll files - tune

kgilpin:
Optimizations to fix/poll-files, moved into their own branch so that fix/poll-files is easier to merge.

WIP, needs tests.


dividedmind:
> WIP, needs tests.

What kind of tests? It seems to me to all be covered by existing tests.
(If you don't believe me, try breaking the changes and see the tests fail. I did.)

kgilpin:
How would I know that?

As a reviewer I can't be expected to re-perform that exercise.

dividedmind:
> How would I know that? As a reviewer I can't be expected to re-perform that exercise.

I don't know what to tell you. I updated the PR description with pointers to the relevant test cases. Lmk if you feel a test case is missing and you'd like it added, but that wouldn't really be related to these code changes.

dustinbyrne (approved):
The last commit ([perf: Limit scanning in watcher to two concurrent jobs](https://github.com/getappmap/appmap-js/pull/747/commits/767c3f7447b0309792ee2453f3dbb023d1401539)) seems worth shipping as a `fix`. We've seen cases where the process runs out of memory in the wild and this seems like it's likely related.

Does `index` need to constrain queue concurrency as well? If it's not already, it seems worth doing so.

PR #752, opened 2022-09-29
Send telemetry events on indexing

dividedmind:
The telemetry data sent includes number of AppMaps seen and some common metadata about them. An event is not sent more often than once per second.

Note this only applies to watch mode. The intention is to let us see when our users are succeeding with AppMap creation, and as live indexing automatically happens when using IDEs, this solution avoids having to call home from every agent. Telemetry can also be easily disabled in the editors.

Example event:
```json
    {
      "name": "appmap:index",
      "properties": {
        "app": "appland/AppLand",
        "client.name": "appmap",
        "client.url": "https://github.com/applandinc/appmap-ruby",
        "framework.rails": "5.2.4.2",
        "framework.rspec": "3.9.1",
        "labels": "api key, api",
        "language.name": "ruby",
        "language.version": "2.6.6",
        "language.engine": "ruby",
        "recorder.name": "rspec"
      },
      "metrics": {
        "appmaps": 2
      }
    }
```

Fixes https://github.com/getappmap/board/issues/167

### DIFF
--- a/packages/cli/src/lib/flattenMetadata.ts
+++ b/packages/cli/src/lib/flattenMetadata.ts
@@ -1,0 +1,37 @@
+import { Metadata } from '@appland/models';
+
+/** Flattens metadata into a string-string map suitable for use in telemetry.
+ * Ignores git, exception and fingerprints.
+ */
+export default function flattenMetadata(metadata: Metadata): Map<string, string> {
+  const result = new Map<string, string>();
+
+  if (metadata.app) result.set('app', metadata.app);
+
+  if (metadata.client) {
+    result.set('client.name', metadata.client.name);
+    result.set('client.url', metadata.client.url);
+    if (metadata.client.version) result.set('client.version', metadata.client.version);
+  }
+
+  if (metadata.frameworks)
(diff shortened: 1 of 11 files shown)


packages/cli/src/lib/flattenMetadata.ts
```diff
@@ -0,0 +1,34 @@
...
+  if (metadata.labels) result.set('labels', metadata.labels.join(', '));
+
+  if (metadata.language) {
+    result.set('language.name', metadata.language.name);
+    result.set('language.version', metadata.language.version);
+    if (metadata.language.engine) result.set('language.engine', metadata.language.engine);
+  }
+
+  if (metadata.name) result.set('name', metadata.name);
+  if (metadata.recorder?.name) result.set('recorder.name', metadata.recorder.name);
```
kgilpin, line 29: Please add recorder.type to this - new in 1.9.0

kgilpin (approved):

packages/cli/src/lib/flattenMetadata.ts
dividedmind: Oh, ok. It seems the type in models needs updating in this case. We really should try to keep them up to date.

PR #756, opened 2022-10-04
`scan --watch` works correctly even when the appmap dir doesn't initially exist

### LINKED ISSUE #753: `scan --watch` fails when appmap directory is not present
```sh-session
$ yarn start scan -v --watch -d /home/user/projects/rails-project
ValidationError: AppMap directory tmp/appmap does not exist, or is not readable.
    at /home/user/projects/appmap-js/packages/scanner/src/cli/validateFile.ts:12:11
    at Generator.throw (<anonymous>)
    at rejected (/home/user/projects/appmap-js/packages/scanner/src/cli/validateFile.ts:6:65)
```

dividedmind:
Also, fix paths handling for windows.

Fixes #753 .

### DIFF
--- a/packages/scanner/src/cli/scan/watchScan.ts
+++ b/packages/scanner/src/cli/scan/watchScan.ts
@@ -24,6 +24,26 @@ declare module 'async' {
   }
 }
 
+async function isDir(targetPath: string): Promise<boolean> {
+  try {
+    return (await stat(targetPath)).isDirectory();
+  } catch {
+    return false;
+  }
+}
+
+async function existingParent(targetPath: string): Promise<string> {
+  while (targetPath.length > 1) {
+    if (await isDir(targetPath)) break;
+    targetPath = path.dirname(targetPath);
+  }
+  return targetPath;
+}
+
+function isAncestorPath(ancestor: string, descendant: string): boolean {
+  return !path.relative(ancestor, descendant).startsWith('..');
+}
+
 export class Watcher {
   config?: TimestampedConfiguration;
   appmapWatcher?: chokidar.FSWatcher;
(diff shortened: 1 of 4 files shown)


ahtrotta:
It looks like most of the tests didn't run in the [Windows Travis build](https://app.travis-ci.com/github/getappmap/appmap-js/jobs/584668737), so I ran them on my local Windows machine and got these test failures:

image

packages/scanner/src/cli/scan/watchScan.ts
```diff
@@ -42,23 +62,40 @@ export class Watcher {
...
-      ignored: ['**/node_modules/**', '**/.git/**'],
+      ignored,
       usePolling: true,
       interval: 1000,
       persistent: false,
     });
 
-    const enqueue = this.enqueue.bind(this);
+    const enqueue = (filePath: string) =>
+      path.basename(filePath) === 'mtime' && this.enqueue(filePath);
```
ahtrotta, line 98: Does this function need to be bound to `this` so that it has the right execution context when passed as an argument below?

ahtrotta (changes requested):
Looks good overall. Just need to fix the Windows test failures and revert to using `sinon`

ahtrotta:
I've looked into the Windows test failures and the problem is the custom `ignored` function. If I switch it to use the `['**/node_modules/**', '**/.git/**']` glob pattern, then it works.

dividedmind:
> I've looked into the Windows test failures and the problem is the custom `ignored` function. If I switch it to use the `['**/node_modules/**', '**/.git/**']` glob pattern, then it works.

The `ignored` function is kinda crucial for cutting down the search tree and limiting the search to the appmap dir. I'll try to figure out where it's having issues.

dividedmind:
> Also, based on the conversation in standup today and from [this PR](https://github.com/getappmap/appmap-js/pull/762), it seems like the consensus is to keep using `sinon` instead of `jest` mocks.

I wouldn't call it a consensus per se, just that it didn't make sense to convert a bunch of existing tests without previous discussion ;)

dividedmind:
Thanks for testing with windows! I think the latest change did the trick. Fingers crossed.

ahtrotta:
It looks like most of the tests are working now :)

I'm still getting a failure on this one, though:

image

dividedmind:
> It looks like most of the tests are working now :)
> 
> I'm still getting a failure on this one, though:
> 
> image

This one looks unrelated to the pull request substance, it's just Windows being Windows and breaking on temp directory cleanup. I think it's safe to leave it as it is until we circle back to take the time to fix tests on Windows more broadly. WDYT?

ahtrotta (approved):
I agree, and I'm going start looking into Windows test failures soon.

dividedmind:
> I agree, and I'm going start looking into Windows test failures soon.

Thanks! I created an issue for this: #771

PR #762, opened 2022-10-05
fix: Don't show a misleading message when test fails with AppMaps

### LINKED ISSUE #728: "record test" issues a misleading message when tests fail
Image

The highlighted message is wrong, obviously. Note that this isn't the same as #719, because many (possibly all, I didn't check) of the AppMaps were created during the test run.

dividedmind:
When running `appmap record` if a test run failed a message was
shown suggesting no maps were created even if some, in fact, were.
We're not showing that message anymore (unless maps were indeed
not created).

Fixes #728

### DIFF
--- a/packages/cli/src/cmds/record/state/testCasesComplete.ts
+++ b/packages/cli/src/cmds/record/state/testCasesComplete.ts
@@ -9,7 +9,8 @@ export default async function testCasesComplete(recordContext: RecordContext): P
   // Handle command failures here, rather than in a separate state, so we maintain compatibility
   // with the Azure Function that processes telemetry events.
   if (recordContext.failures > 0 || recordContext.appMapsCreated === 0) {
-    UI.warn(`\n${chalk.yellow('!')} The test commands failed to create AppMaps\n`);
+    if (recordContext.appMapsCreated === 0)
+      UI.warn(`\n${chalk.yellow('!')} The test commands failed to create AppMaps\n`);
 
     const errors: string | string[] =
       recordContext.output?.join('\n').length > 0
--- a/packages/cli/tests/unit/record/testCasesComplete.spec.ts
+++ b/packages/cli/tests/unit/record/testCasesComplete.spec.ts
@@ -4,6 +4,7 @@ import RecordContext from '../../../src/cmds/record/recordContext';
 import testCasesComplete from '../../../src/cmds/record/state/testCasesComplete';
 import * as openTicket from '../../../src/lib/ticket/openTicket';
 import Configuration from '../../../src/cmds/record/configuration';
+import UI from '../../../src/cmds/userInteraction';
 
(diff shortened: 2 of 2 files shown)


apotterri:
Why is it desirable to get rid of `sinon`? Last I looked, the jest mocking/stubbing functionality seemed less rich....

packages/cli/src/cmds/record/state/testCasesComplete.ts
```diff
@@ -9,7 +9,8 @@ export default async function testCasesComplete(recordContext: RecordContext): P
   // Handle command failures here, rather than in a separate state, so we maintain compatibility
   // with the Azure Function that processes telemetry events.
   if (recordContext.failures > 0 || recordContext.appMapsCreated === 0) {
-    UI.warn(`\n${chalk.yellow('!')} The test commands failed to create AppMaps\n`);
+    if (recordContext.appMapsCreated === 0)
```
kgilpin, line 12: Isn’t the main issue that this command can think that no AppMaps were created when the test command runs for a second time and updates a bunch of existing AppMaps?

kgilpin: If AppMap time stamps are updated then this command can assume that AppMaps were in fact created and that would be a good fix to make as well.

apotterri: That's this issue: https://github.com/getappmap/appmap-js/issues/719 . I thought, for the sake of small batches, it was reasonable to split them up.

packages/cli/tests/unit/record/testCasesComplete.spec.ts
```diff
@@ -1,52 +1,75 @@
-import sinon from 'sinon';
```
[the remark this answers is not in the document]
dividedmind: But barely, and it doesn't really bring anything over what jest already provides. I figured it might be a good idea to incrementally remove it while I'm touching something anyway. I can rewrite it with sinon if you'd like, but I figure the less dependencies the better :)

dividedmind:
> Why is it desirable to get rid of `sinon`? Last I looked, the jest mocking/stubbing functionality seemed less rich....

It really isn't, the interface might be less fluid (or maybe just less similar to rspec), but capability-wise it's quite similar (and actually jest does have some nice extra features, such as very easy mocking of complete modules, which could be nice for pure unit tests; we don't currently use that though). I figured we use jest anyway (and probably won't change the test runner anytime soon, jest is pretty nice all in all), why dig into another dependency...

Anyway, I didn't mean this PR to be controversial, I can easily rewrite it with sinon (or just revert the migration and keep the new tests with jest) if it's questionable :)

apotterri (approved):
Looks good, thanks.

PR #764, opened 2022-10-05
feat: handle when bundler ignores test and development

ahtrotta:
Fixes getappmap/board#180

If the user has bundler configured to ignore both `test` and `development` groups, send a telemetry event and show them this error:

image


ahtrotta:
Here's the new message:

image

apotterri:
This looks better to me. I'd be interested to hear what other think. @kgilpin ?

apotterri (approved):

PR #766, opened 2022-10-05
fix: fix broken tests

### LINKED ISSUE #765: Fix broken tests for CLI

ahtrotta:
Fixes #765

### DIFF
--- a/packages/cli/src/cmds/runCommand.ts
+++ b/packages/cli/src/cmds/runCommand.ts
@@ -2,20 +2,8 @@ import { AbortError, ValidationError } from './errors';
 import { verbose } from '../utils';
 import UI from './userInteraction';
 import chalk from 'chalk';
-import Yargs, { command } from 'yargs';
 import { ExitCode } from './record/types/exitCode';
 import Telemetry from '../telemetry';
-
-async function invokeCommand() {}
-
-function flushTelemetry(exitCode: ExitCode, err?: Error) {
-  Telemetry.flush(() => {
-    // @types/yargs defines:
-    //   exit(code: number, err: Error): void;
-    // but in the implementation, err appears to be optional.
-    Yargs.exit(exitCode, err as any);
-  });
-}
 
 export default async function runCommand(
   commandPrefix: string,
@@ -28,7 +16,7 @@ export default async function runCommand(
       name: `${commandPrefix}:exit-ok`,
     });
 
-    flushTelemetry(0);
+    process.exitCode = 0;
     return ret;
   } catch (err) {
     if (err instanceof ValidationError) {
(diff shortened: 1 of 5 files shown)


packages/cli/src/cmds/runCommand.ts
```diff
@@ -28,7 +28,11 @@ export default async function runCommand(
       name: `${commandPrefix}:exit-ok`,
     });
 
-    flushTelemetry(0);
+    // only run this when not in Travis builds becuase
+    // this immediately exits the test suite in Travis
+    if (!process.env.TRAVIS) {
+      flushTelemetry(0);
```
dividedmind, line 34: I think flushTelemetry call (and function) should just be removed. This was needed because cli used to do `process.exit()` which killed the process before the telemetry had the chance to be flushed — this is discouraged in node precisely because it prevents any scheduled actions to run.

I believe I had changed all the places where that was done to set `process.exitStatus =` instead, which is the recommended way to set the exit status and avoids this problem. Therefore this is just not needed anymore.

packages/cli/tests/unit/depends.spec.ts
```diff
@@ -1,10 +1,11 @@
...
-const Fingerprinter = require('../../src/fingerprint/fingerprinter');
-const { listAppMapFiles, verbose } = require('../../src/utils');
+import { utimesSync } from 'fs';
+import { join } from 'path';
+import tmp from 'tmp';
+import fs from 'fs-extra';
+
+import Fingerprinter from '../../src/fingerprint/fingerprinter';
+import Depends from '../../src/depends';
+import { listAppMapFiles, verbose } from '../../src/utils';
```
dividedmind, line 8: Please separate style commits from substantial ones. It makes it unnecessarily difficult to review the PR and use `git blame` later.

dividedmind (changes requested):

dividedmind (approved):
This `exitCode` changes are perfect, thanks!

I'd still like you to separate the style changes in a separate commit. 

BTW, those fixup commits are a bit unorthodox. They don't seem to have been generated with `git commit --fixup`, so it's difficult to say what they should fix up (generally, here there's only one commit, I suppose :) ) and they don't get automatically applied by `git rebase --interactive`.

But that's fine as long as you clean this up before merging :) Thanks!

ahtrotta:
Great! 

Sorry that I didn't split up the commits, I thought you meant it as a thing to do in the future. I'll split it up before I merge it.

The fixup commits were just done manually by starting the commit message with `fixup!`. I usually squash my commits down to one or two commits before merging, so I just wrote the fixup commits like that as a notice to myself for which commits to squash. I'll be sure to use `git commit --fixup` in the future, though!

dividedmind:
> Great!
> 
> Sorry that I didn't split up the commits, I thought you meant it as a thing to do in the future. I'll split it up before I merge it.
> 
> The fixup commits were just done manually by starting the commit message with `fixup!`. I usually squash my commits down to one or two commits before merging, so I just wrote the fixup commits like that as a notice to myself for which commits to squash. I'll be sure to use `git commit --fixup` in the future, though!

Sure, makes sense! Just thought maybe you don't know about the `--fixup` and how it works together with `rebase --interactive`.

BTW, this should be good to go as soon as we merge #767 and this is rebased on top of that.

PR #767, opened 2022-10-06
fix: More reliable polling in index --watch

dividedmind:
Instead of using chokidar polling mode, periodically rescan the directory with glob to find appmaps. This is more reliable and avoids many corner cases of chokidar.

### DIFF
--- a/packages/cli/src/fingerprint/fingerprintWatchCommand.ts
+++ b/packages/cli/src/fingerprint/fingerprintWatchCommand.ts
@@ -6,15 +6,16 @@ import EventAggregator from '../lib/eventAggregator';
 import flattenMetadata from '../lib/flattenMetadata';
 import { intersectMaps } from '../lib/intersectMaps';
 import Telemetry from '../telemetry';
-import { verbose, listAppMapFiles } from '../utils';
+import { verbose } from '../utils';
 import { FingerprintEvent } from './fingerprinter';
 import FingerprintQueue from './fingerprintQueue';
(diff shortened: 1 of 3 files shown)


apotterri (approved):
Code looks good to me. As long as @kgilpin is on board with this change, I'd say we should go ahead with it.

ahtrotta:
I'm getting a ton of test failures in Windows, but most of them aren't relevant to this PR (I'll open a separate issue). Here are the ones that are:

image

image

ahtrotta:
I just built this and ran it manually in Windows, and it seems to work well :)

ahtrotta:
I figured out the problem with the `rejects appmaps that are too large` test. You just need to change lines 24 & 25 in `tests/unit/fingerprinter/fingerprint.spec.ts` to use `await`:

```
await file.write('{}', 500 * 1000 * 1000);
await file.close();
```

dividedmind:
> I figured out the problem with the `rejects appmaps that are too large` test. You just need to change lines 24 & 25 in `tests/unit/fingerprinter/fingerprint.spec.ts` to use `await`:
> 
> ```
> await file.write('{}', 500 * 1000 * 1000);
> await file.close();
> ```

Good catch, thanks! The telemetry failure is unrelated, but I've seen it too. I know what the problem is, but I haven't figured out a good fix yet. Fortunately it doesn't seem to occur in CI.

ahtrotta (approved):
Looks good to me!

Issue #770, opened 2022-10-07
TypeError: Cannot read properties of undefined (reading 'length') in validateProject


symwell:
Saw this error 
```
TypeError: Cannot read properties of undefined (reading 'length')
    at AgentInstallerProcedure.validateProject (C:\Users\test\AppData\Local\npm-cache\_npx\3007b677135bbf3c\node_modules\@appland\appmap\built\cmds\agentInstaller\agentProcedure.js:78:20)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async AgentInstallerProcedure.run (C:\Users\test\AppData\Local\npm-cache\_npx\3007b677135bbf3c\node_modules\@appland\appmap\built\cmds\agentInstaller\agentInstallerProcedure.js:67:13)
    at async _handler (C:\Users\test\AppData\Local\npm-cache\_npx\3007b677135bbf3c\node_modules\@appland\appmap\built\cmds\agentInstaller\install-agent.js:133:17)
    at async Object.handler (C:\Users\test\AppData\Local\npm-cache\_npx\3007b677135bbf3c\node_modules\@appland\appmap\built\cmds\agentInstaller\install-agent.js:222:35)
```

Seems to be coming from validateProject from this line
```
    const errors = Array.isArray(validationResult) ? validationResult : validationResult.errors;
    if (errors.length > 0) {
```

symwell:
To debug this I edited `~/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/gems/appmap-0.93.2/lib/appmap/command/agent_setup/validate.rb` so that instead of printing out a JSON verification output it prints invalid JSON, to trigger an error. Then called
```
~/src/appmap-js/packages/cli$ npm start install /home/test/src/sample_app_6th_ed
```

dustinbyrne:
I was actually able to recreate this just now with the following steps:
```sh
$ mkdir js-project
$ cd js-project
$ yarn init
   (enter enter enter...)
$ yarn
$ npx @appland/appmap install
```

I received the following output from the install command:
```
Installing AppMap agent for ....
? AppMap is about to be installed. Confirm the details below.
  !!! Telemetry disabled !!!: true
  Project type: yarn
  Project directory: /home/db/dev/tmp/js-project
  Git remote: [git remote failed, fatal: not a git repository (or any of the parent directories): .git]

  Is this correct? Yes
? An appmap.yml configuration file already exists. How should the conflict be resolved? Use existing
[...]

PR #772, opened 2022-10-10
fix: Fix TypeError: Cannot read properties of undefined

symwell:
If validateProject fails to `JSON.parse(stdout)`, print out the stdout that failed to be parsed and display an error message.

Fixes https://github.com/getappmap/appmap-js/issues/770

### DIFF
--- a/packages/cli/src/cmds/agentInstaller/agentProcedure.ts
+++ b/packages/cli/src/cmds/agentInstaller/agentProcedure.ts
@@ -77,21 +77,27 @@ export default abstract class AgentProcedure {
 
     let { stdout } = await this.validateAgent(validateCmd);
 
-    const validationResult = JSON.parse(stdout);
+    try {
+      const validationResult = JSON.parse(stdout);
 
-    const errors = Array.isArray(validationResult) ? validationResult : validationResult.errors;
-    if (errors.length > 0) {
-      throw new ValidationError(
-        errors
-          .map((e) => {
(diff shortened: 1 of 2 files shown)


dividedmind (changes requested):
Looks ok, but could you please add a test case for this? Thanks.

symwell:
I tested it like this https://github.com/getappmap/appmap-js/issues/770#issuecomment-1273587287.

Will try to add a testcase.

dividedmind:
> I tested it like this [#770 (comment)](https://github.com/getappmap/appmap-js/issues/770#issuecomment-1273587287).
> 
> Will try to add a testcase.

Yeah, I've seen, but it'd be nice to have an automated test so there's no regression.

symwell:
Added a testcase.  Waiting for the build.

symwell:
Fixed.

symwell:
I think the error is simply that `error` isn't an array.  There's no `error` in the validationResult.
```
validationResult is 
{
  filename: 'appmap.yml',
  configuration: { contents: 'name: js-project\npackages: []\n' }
}
```

I plan to create a new pr that checks if `error` exists.

symwell:
Simpler fix https://github.com/getappmap/appmap-js/pull/787

symwell:
Closing, since we now believe the real fix was https://github.com/getappmap/appmap-js/pull/787

PR #774, opened 2022-10-10
Fix cli tests in windows

ahtrotta:
Fixes getappmap/board#188


packages/cli/tests/sleeper.js
```diff
@@ -0,0 +1,5 @@
+const timeout = process.argv[2] || 1000;
+
+(async function sleep() {
+  await new Promise((resolve) => setTimeout(resolve, timeout));
+})();
```
dividedmind, line 5: Plain `setTimeout(() => {}, timeout)` at the top level should work too.

packages/cli/tests/unit/agentInstall/javaBuildToolInstaller.spec.ts
```diff
@@ -16,20 +16,21 @@ describe('JavaBuildToolInstaller', () => {
   });
 
   describe('when IntelliJ is installed', () => {
-    const intellijHome = '/IntelliJ';
+    const intellijHome = (process.platform === 'win32' ? '\\' : '/') + 'IntelliJ';
```
dividedmind, line 19: You could also use `path.sep` or, even clearer, `path.join('/', 'IntelliJ')` (I checked, it does the right thing on windows.)

packages/cli/tests/unit/agentInstall/javaBuildToolInstaller.spec.ts
```diff
@@ -16,20 +16,21 @@ describe('JavaBuildToolInstaller', () => {
...
     beforeEach(() => {
       sinon.stub(jbs, 'findIntelliJHome').returns(intellijHome);
     });
 
     it('appends JetBrains JDK to PATH', () => {
       addJetBrainsEnv();
-      expect(process.env.PATH).toMatch(/\/IntelliJ\/jbr\/bin/);
+      expect(process.env.PATH).toMatch(/[\/\\]IntelliJ[\/\\]jbr[\/\\]bin/);
```
dividedmind, line 26: `toContain(path.join(intellijHome, 'jbr', 'bin'))` would also verify the right separator was used.

packages/cli/tests/unit/agentInstall/javaBuildToolInstaller.spec.ts
```diff
@@ -16,20 +16,21 @@ describe('JavaBuildToolInstaller', () => {
...
     });
 
     it('sets JAVA_HOME if unset', () => {
       delete process.env.JAVA_HOME;
       addJetBrainsEnv();
-      expect(process.env.JAVA_HOME).toBe(intellijHome + '/jbr');
+      const expectedIntellijHome = intellijHome + (process.platform === 'win32' ? '\\jbr' : '/jbr');
```
dividedmind, line 32: This wants to be `path.join` :)

packages/cli/tests/unit/depends.spec.ts
```diff
@@ -10,7 +10,7 @@ import { listAppMapFiles, verbose } from '../../src/utils';
 if (process.env.DEBUG !== 'true') tmp.setGracefulCleanup();
 
 const fixtureDir = join(__dirname, 'fixtures', 'ruby');
-const now = Date.now();
+const now = Date.now() / 1000;
```
dividedmind, line 13: I don't understand this change.

packages/cli/tests/unit/depends.spec.ts
```diff
@@ -20,6 +20,7 @@ describe('Depends', () => {
   beforeAll(async () => verbose(process.env.DEBUG === 'true'));
   beforeEach(async () => {
     appMapDir = tmp.dirSync().name;
+    expected = join(appMapDir, 'user_page_scenario').split('\\').join('/');
```
dividedmind, line 23: `replaceAll` could work better :) also, if you just want to ensure usage of unix separators, you can use `path.posix.join`.

packages/cli/tests/unit/encodedFile.spec.ts
```diff
@@ -55,7 +55,8 @@ describe('EncodedFile', () => {
   });
 
   it('writes a 7-bit ASCII file', () => {
-    verifyUpdate('ascii', '\n');
+    const newlineChar = process.platform === 'win32' ? '\r\n' : '\n';
+    verifyUpdate('ascii', newlineChar);
```
dividedmind, line 59: I don't understand how it now passes on Windows. The test file uses `\n`, so the test should now fail, right?

BTW, perhaps it would be a good idea to move those ascii tests into the per-platform section above.

packages/cli/tests/unit/inspect.spec.ts
```diff
@@ -11,9 +11,9 @@ import { CodeObject, CodeObjectMatch } from '../../src/search/types';
 tmp.setGracefulCleanup();
 
 const fixtureDir = join(__dirname, 'fixtures', 'ruby');
-const appMapDir = tmp.dirSync().name;
+const appMapDir = tmp.dirSync().name.split('\\').join('/');
```
dividedmind, line 14: Or `replaceAll` ;)

packages/cli/tests/unit/inspect.spec.ts
```diff
@@ -11,9 +11,9 @@ import { CodeObject, CodeObjectMatch } from '../../src/search/types';
...
 
-const now = Date.now();
+const now = Date.now() / 1000;
```
dividedmind, line 16: This again..?

packages/cli/tests/unit/record/testCaseRecording.spec.ts
```diff
@@ -51,12 +55,24 @@ describe('record.testCaseRecording', () => {
 
       await TestCaseRecording.start(context);
       await TestCaseRecording.waitFor(context);
+      const fullExpectedCommand = process.platform === 'win32' ? 
+        'C:\\WINDOWS\\system32\\cmd.exe /d /s /c "node .\\tests\\sleeper.js 1000"' : 
+        '/bin/sh -c sleep 1';
+
+      // windows will throw an error when a process is killed using the pid
+      const windowsError = '\nTest command failed with status code 1: C:\\WINDOWS\\system32\\cmd.exe /d /s /c "node .\\tests\\sleeper.js 1000"';
```
dividedmind, line 63: This constant could be moved down to the `if (process.platform)`; also it could use `fullExpectedCommand` to make it a little DRYer.

dividedmind (changes requested):
Looks good, thanks! I had some questions, and some stylistic suggestions but you're free to take them or leave them.

The only thing I'm particularly (not too much though) concerned about is the prettier change. I'd rather we don't see crlf in the repository one day ;)

If the tests are fixed, can we change travis configuration to run the full CLI test suite on Windows now?

packages/cli/tests/unit/agentInstall/installCommand.spec.ts
```diff
@@ -28,7 +28,7 @@ import { withStubbedTelemetry } from '../../helper';
 const fixtureDir = path.join(__dirname, '..', 'fixtures');
 tmp.setGracefulCleanup();
 
-const invokeCommand = (
+const invokeCommand = async (
```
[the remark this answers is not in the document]
ahtrotta: True, I changed it back :)

packages/cli/tests/unit/agentInstall/installCommand.spec.ts
```diff
@@ -1053,7 +1053,7 @@ packages:
   });
 
   describe('Multi-project install flow', () => {
-    let expectedStubs;
+    let expectedStubs: any[];
```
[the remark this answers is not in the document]
ahtrotta: Okay, good to know, I changed it to `SinonStub[]`

packages/cli/tests/unit/record/action/configureHostAndPort.spec.ts
```diff
@@ -11,11 +11,12 @@ describe('record.action.configureHostAndPort', () => {
...
 
-  beforeEach(() => {
+  beforeEach(async () => {
     config = new TempConfig();
-    config.initialize();
+    await config.initialize();
     context = new RecordContext(config);
-    return context.initialize();
+    await context.initialize();
+    return;
```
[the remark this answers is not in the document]
ahtrotta: True, changed it back

packages/cli/tests/unit/agentInstall/javaBuildToolInstaller.spec.ts
ahtrotta: Yeah, that's cleaner. I changed it.

packages/cli/tests/unit/agentInstall/javaBuildToolInstaller.spec.ts
ahtrotta: True!

packages/cli/tests/unit/depends.spec.ts
ahtrotta: For some reason, when I use `replaceAll` I get this error:
image

I changed it to `replace`, though.

packages/cli/tests/unit/depends.spec.ts
ahtrotta: [Apparently, `utimesSync` takes timestamps in seconds, but `Date.now()` generates them in milliseconds](https://github.com/nodejs/node/issues/33227#issuecomment-623391936). But I'm confused as to why it works in Linux / Mac if that's the case...

packages/cli/tests/unit/inspect.spec.ts
ahtrotta: [Apparently, `utimesSync` takes timestamps in seconds, but `Date.now()` generates them in milliseconds](https://github.com/nodejs/node/issues/33227#issuecomment-623391936). But I'm confused as to why it works in Linux / Mac if that's the case...

packages/cli/tests/unit/encodedFile.spec.ts
ahtrotta: I'm not sure, but I've noticed that all files have `\r\n` by default in Windows. Even when I'm pulling them from GitHub, where they were generated with `\n`.

When I move the test into the per-platform section, I get these errors:
image

packages/cli/tests/unit/record/testCaseRecording.spec.ts
ahtrotta: Good idea!

ahtrotta:
Thanks for the excellent review! I addressed all of your comments. Let me know if you have any other questions or concerns.

I made it so all of the tests except for one run in Windows on Travis.

.prettierrc.cjs
```diff
@@ -4,7 +4,7 @@ module.exports = {
   semi: true,
   singleQuote: true,
   arrowParens: 'always',
-  endOfLine: 'lf',
+  endOfLine: 'auto',
```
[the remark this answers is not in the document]
dividedmind: This is weird. Sounds like something is changing your line endings; I suspect git. Maybe you need to change your git config? `git config --global core.autocrlf input` or `git config --global core.autocrlf false` should do the trick.

packages/cli/tests/unit/encodedFile.spec.ts
dividedmind: Yup, that's probably the problem of your git configuration (autocrlf).

packages/cli/tests/unit/inspect.spec.ts
dividedmind: It is weird. Too old a typescript or some type definitions? But I'd think they'd be pinned by yarn.

packages/cli/tests/testFilter.js
```diff
@@ -0,0 +1,17 @@
+const FILTERED_TESTS = process.platform === 'win32' ? ['fingerprintWatchCommand'] : [];
```
dividedmind, line 1: Why does this test need to be filtered? If there's something left to fix for this issue, that's fine, but please remember to create another one to keep track of the loose end.

dividedmind:
> Thanks for the excellent review! I addressed all of your comments. Let me know if you have any other questions or concerns.
> 
> I made it so all of the tests except for one run in Windows on Travis.

Thanks, that looks much better! Now I'm only worried about those line endings. Perhaps it would be a good idea to add a `gitattributes` per-file eol override for the affected expectation files to prevent git from breaking the tests. (It'd probably look like ```
packages/cli/tests/unit/fixtures/encodedFile/**/* eol=lf
packages/cli/tests/unit/fixtures/encodedFile/windows/* eol=crlf
``` — haven't checked if it actually works, YMMV. Another option would be to do `-text` for all these files so git leaves them alone.)

File: packages/cli/tests/testFilter.js (hunk 15 above)
Comment (ahtrotta) [reply packages/cli/tests/testFilter.js:1]: This one gives this error in Travis on Windows:

image

You can see it in the logs on this build: https://app.travis-ci.com/github/getappmap/appmap-js/jobs/585469980#L220

I'll make a new issue for it.

File: packages/cli/tests/unit/depends.spec.ts (hunk 5 above)
Comment (ahtrotta) [reply packages/cli/tests/unit/depends.spec.ts]: I agree, I implemented this

**Comment by ahtrotta** [comment]
I put all of the tests into the platform-specific `describe` blocks in `encodedFile.spec.ts`, and all of the tests except for one are passing: 6ea720473bb0c5c934156f0471a26ce9cfdf478c

The strange thing is that this test passes for me locally on Windows and on Linux, so I'm not sure how to debug this. I tried changing the `.travis.yml` file, but it didn't do anything.

Do you have any thoughts?

**Comment by ahtrotta** [comment]
Okay, I got it to work. Turns out that forcing `CRLF` in the Windows files using:
```
/packages/cli/tests/unit/fixtures/encodedFile/windows/**/* eol=crlf
```
was causing issues on Travis. I'm not sure why. But I needed to force `LF` for the Unix test files for it to work on Windows.

Let me know if there's anything else you think needs to be changed!

dividedmind (approved):
Looks great now, thanks for figuring out this pesky EOL business!

(Please remember to rebase and fixup everything before merging.)

PR #779, opened 2022-10-12
fix: fix scanner tests in windows

ahtrotta:
Fixes getappmap/board#189

This fixes all of the tests except one:
image

I've been trying to fix this one and it's been resisting my attempts. I'll make a separate issue for it. Any thoughts would be appreciated.

### DIFF
--- a/packages/openapi/src/provider.ts
+++ b/packages/openapi/src/provider.ts
@@ -1,5 +1,5 @@
 import { load } from 'js-yaml';
-import { URL } from 'url';
+import { fileURLToPath, URL } from 'url';
 import { OpenAPIV3 } from 'openapi-types';
 import http, { IncomingMessage } from 'http';
 import https from 'https';
@@ -32,7 +32,7 @@ const URLLoader = (protocol: Pick<typeof http, 'get'>) => {
 };
 
 const FileLoader = (url: URL): Promise<Buffer> => {
-  return readFile(url.pathname);
+  return readFile(fileURLToPath(url));
 };
 
 const ProtocolLoader: Record<string, Loader> = {
--- a/packages/scanner/src/rules/lib/parseRuleDescription.ts
+++ b/packages/scanner/src/rules/lib/parseRuleDescription.ts
@@ -7,7 +7,8 @@ export default function parseRuleDescription(id: string): string {
 
   if (!fs.existsSync(docPath)) return `No doc exists for rule ${id}`;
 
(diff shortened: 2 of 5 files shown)


packages/scanner/src/rules/lib/parseRuleDescription.ts
```diff
@@ -7,7 +7,12 @@ export default function parseRuleDescription(id: string): string {
 
   if (!fs.existsSync(docPath)) return `No doc exists for rule ${id}`;
 
-  const content = fs.readFileSync(docPath, 'utf-8');
+  let content = fs.readFileSync(docPath, 'utf-8');
+
+  // replace carriage return with newline in windows
+  if (process.platform === 'win32') {
+    content = content.replace(/\r\n/g, '\n');
```
dividedmind, line 14: It should be safe to do it regardless of the platform.

packages/openapi/src/provider.ts
```diff
@@ -31,9 +31,22 @@ const URLLoader = (protocol: Pick<typeof http, 'get'>) => {
...
-  return readFile(url.pathname);
-};
+  let path = url.pathname;
+
+  // if a URL is generated from a file path starting with a drive letter
+  // then the pathname is generated incorrectly (e.g. '/C:/Users/...')
+  // so this strips the leading backslash (e.g. 'C:/Users/...')
+  if (startsWithDriveLetter(url.pathname)) {
+    path = url.pathname.slice(1);
+  }
```
dividedmind, line 46: Seems like Node offers a `fileURLToPath` function to do this job, might be worth using in case there are other quirks.

packages/scanner/bin/build.js
```diff
@@ -0,0 +1,29 @@
...
+        copyFolderRecursively(currentSource, targetFolder);
+      } else {
+        copyFileSync(currentSource, currentDestination);
+      }
+    });
+  }
+}
+
+mkdirSync('built', { recursive: true });
+copyFolderRecursively('src/sampleConfig', 'built');
```
dividedmind, line 29: The package already pulls `fs-extra` as a dev dependency, might as well do `fsextra.copy('src/sampleConfig', 'built/sampleConfig')` instead of rolling your own. 

BTW, the name of this file is misleading — it doesn't build anything, it just prepares the output directory.

dividedmind (approved):
Looks good, thanks! Had a couple of suggestions for improvements :)

PR #787, opened 2022-10-14
fix: Don't crash in the installation if there were no validation errors.

symwell:
Fixes https://github.com/getappmap/appmap-js/issues/770

It's not easy to write a testcase for this.

```
$ mkdir js-project
$ cd js-project
$ yarn init
   (enter enter enter...)
$ yarn
$ npx @appland/appmap install
Installing AppMap agent for ....
  Project type: yarn
  Project directory: /home/test/tmp/js-project
  Git remote: [git remote failed, fatal: not a git repository (or any of the parent directories): .git]

  Is this correct? Yes
? An appmap.yml configuration file already exists. How should the conflict be resolved? Use existing
✔ Installing the AppMap agent...
⠋ Validating the AppMap agent...
  Validating the AppMap agent...
✔ Validating the AppMap agent...

   ╭─────────────────────────────────────────────────────────────────────╮
   │                                                                     │
   │            Success! The AppMap agent has been installed.            │
   │                                                                     │
   │                      NEXT STEP: Record AppMaps                      │
   │                                                                     │
   │   You can consult the AppMap documentation, or continue with the    │
   │     instructions provided in the AppMap code editor extension.      │
   │                                                                     │
   ╰─────────────────────────────────────────────────────────────────────╯

```

### DIFF
--- a/packages/cli/src/cmds/agentInstaller/agentProcedure.ts
+++ b/packages/cli/src/cmds/agentInstaller/agentProcedure.ts
@@ -80,7 +80,8 @@ export default abstract class AgentProcedure {
     const validationResult = JSON.parse(stdout);
 
     const errors = Array.isArray(validationResult) ? validationResult : validationResult.errors;
-    if (errors.length > 0) {
+    // if there were no errors then there's no "errors" key
+    if (errors && errors.length > 0) {
       throw new ValidationError(
         errors
           .map((e) => {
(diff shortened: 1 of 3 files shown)


packages/cli/src/cmds/agentInstaller/agentProcedure.ts
```diff
@@ -80,7 +80,8 @@ export default abstract class AgentProcedure {
     const validationResult = JSON.parse(stdout);
 
     const errors = Array.isArray(validationResult) ? validationResult : validationResult.errors;
-    if (errors.length > 0) {
+    // if there were no errors then there's no "errors" key
+    if (errors && errors.length > 0) {
```
dividedmind, line 84: Thanks, good catch!

That's what you get when you (@apotterri, I suppose) miss an implicit `any` :) Type checking would have caught that. I suppose it might be a good idea to add concrete types to this file and make sure it all typechecks one of these days.

dividedmind (approved):
Thanks! I added a test for you.

PR #790, opened 2022-10-17
feat: Don't crash if validateProject fails to JSON.parse(stdout)

symwell:
If validateProject fails to JSON.parse(stdout), print out the stdout that failed to be parsed and display an error message.

Fixes https://github.com/getappmap/board/issues/193.

This is basically https://github.com/getappmap/appmap-js/pull/772 rebased to integrate the recent `error` [fix](https://github.com/getappmap/appmap-js/pull/787).


dividedmind (approved):
Looks good; thanks for fixing the test!

PR #791, opened 2022-10-17
feat: run all tests on windows in ci

### LINKED ISSUE #771: Run more tests on Windows in CI
We currently only run a couple of selected tests in CI on Windows. This has lead to only discovering problems on this platform incidentally.

All tests should run on Windows. Note some tests might need some tweaking. It's ok to start with running all but a handful of exceptions to be fixed later.

ahtrotta:
Fixes #771

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,8 +25,7 @@ jobs:
       os: windows
       env:
         - YARN_GPG=no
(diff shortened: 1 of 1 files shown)


brikelly (approved):
LGTM!

PR #800, opened 2022-10-19
refactor: add org name in guess test commands

lachrist:
Normally, the user should have installed the js agent at this point. But we still see failures about https://registry.npmjs.org/appmap-agent-js not existing in telemetry. I'm not sure what to think about this.


symwell (approved):

brikelly:
Thanks @symwell ! I'm going to merge this now since it's affecting current users and the changes look great to me too.

brikelly (approved):
Looks correct, thanks!

symwell:
I see the tests passed in at least one configuration.  So it should be ok to merge.

PR #807, opened 2022-10-20
fix: scanner works on windows

### LINKED ISSUE #793: AppMap Analysis in VS Code does not produce findings on Windows

ahtrotta:
Fixes #793

### DIFF
--- a/packages/scanner/src/cli/scan/watchScan.ts
+++ b/packages/scanner/src/cli/scan/watchScan.ts
@@ -118,7 +118,7 @@ export class Watcher {
     this.queue.push(mtimePath);
   }
 
-  protected async scan(mtimePath: string): Promise<void> {
+  protected async scan(mtimePath: string, callback?: () => void | ErrorCallback): Promise<void> {
     assert(this.config, `config should always be loaded before appmapWatcher triggers a scan`);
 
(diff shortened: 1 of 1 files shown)


packages/scanner/src/cli/scan/watchScan.ts
```diff
@@ -118,7 +118,8 @@ export class Watcher {
     this.queue.push(mtimePath);
   }
 
-  protected async scan(mtimePath: string): Promise<void> {
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  protected async scan(mtimePath: string, callback?: Function): Promise<void> {
```
dustinbyrne, line 122: Good catch. We're compiling down to ES2015 so the native async on this method is replaced with a polyfill. I'm surprised this was only not working on Windows.

My only nit-pick here is the use of `Function` over `() => void | Promise<void>` or a similar typedef. ESLint doesn't like `Function` it because it'll accept any type of function or class.

dustinbyrne (approved):

packages/scanner/src/cli/scan/watchScan.ts
ahtrotta: Okee dokee, I'll change it

PR #823, opened 2022-10-26
fix: Listing AppMap files will now ignore non-permissioned directories

### LINKED ISSUE #803: Unreadable sub-directories should not raise an exception, they should be ignored
```sh
$ appmap record rest -d ~/dev/backend
...
EACCES: permission denied, scandir '/home/user/backend/data-volume'
```

dustinbyrne:
Glob would raise an error while searching for AppMap files if it encountered a directory it could not read due to permission errors. By specifying `strict: false`, the errors will be ignored and the search will continue on.

Resolves #803


ahtrotta (approved):
Looks good!

PR #832, opened 2022-11-01
feat: When recording tests, don't crash if directory path is invalid.

symwell:
If the directory path supplied as input doesn't exist or isn't a directory, let the user know.

Fixes https://github.com/getappmap/board/issues/202

record_test_file_does_not_exist

### DIFF
--- a/packages/cli/tests/unit/record/record.test.spec.ts
+++ b/packages/cli/tests/unit/record/record.test.spec.ts
@@ -83,8 +83,6 @@ describe('record test', () => {
     it('shows error if directory parameter is file instead of directory', async () => {
       const dirPrefix = tmpdir();
       const directoryParam = `${dirPrefix}/file_not_dir`;
-      // ensuring the file doesn't exist breaks after trying it a few times
-      // await unlink(directoryParam, (err) => {});
       closeSync(openSync(directoryParam, 'w'));
       expect(existsSync(directoryParam)).toEqual(true);
 
--- a/packages/cli/tests/unit/record/record.test.spec.ts
+++ b/packages/cli/tests/unit/record/record.test.spec.ts
@@ -12,6 +12,18 @@ import * as obtainTestCommands from '../../../src/cmds/record/prompt/obtainTestC
 import TestCaseRecording from '../../../src/cmds/record/testCaseRecording';
 import RecordContext, { RecordProcessResult } from '../../../src/cmds/record/recordContext';
 import Configuration from '../../../src/cmds/record/configuration';
+import RecordCommand from '../../../src/cmds/record/record';
+const { tmpdir } = require('os');
+const {
+  openSync,
+  closeSync,
+  symlinkSync,
+  existsSync,
+  unlink,
+  rmdir,
+  mkdir,
+  renameSync,
+} = require('fs');
 
 describe('record test', () => {
   let confirm: sinon.SinonStub,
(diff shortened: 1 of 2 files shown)


packages/cli/tests/unit/record/record.test.spec.ts
```diff
@@ -67,6 +79,64 @@ describe('record test', () => {
     });
   });
 
+  describe('record test with invalid directory parameter', () => {
+    it('shows error if directory parameter is file instead of directory', async () => {
+      const dirPrefix = tmpdir();
+      const directoryParam = `${dirPrefix}/file_not_dir`;
+      // ensuring the file doesn't exist breaks after trying it a few times
+      // await unlink(directoryParam, (err) => {});
```
ahtrotta, line 87: Could you remove these comments?

packages/cli/tests/unit/record/record.test.spec.ts
```diff
@@ -67,6 +79,64 @@ describe('record test', () => {
     });
   });
 
+  describe('record test with invalid directory parameter', () => {
+    it('shows error if directory parameter is file instead of directory', async () => {
```
ahtrotta, line 83: This test name is a little misleading. You're not actually testing if an error is shown. You're testing to make sure no error is thrown and that the `handler` function returns with `null`. Could you either rename the test or add logic for testing the output to the terminal?

packages/cli/tests/unit/record/record.test.spec.ts
```diff
@@ -67,6 +79,64 @@ describe('record test', () => {
...
+
+    it('shows error if directory parameter is invalid symlink', async () => {
+      const dirPrefix = tmpdir();
+      const symlinkDst = `${dirPrefix}/symlink_dst`;
+      const symlinkSrc = `${dirPrefix}/symlink_src`;
+      const symlinkDstRenamed = `${dirPrefix}/symlink_dst_renamed`;
+
+      await unlink(symlinkSrc, (err) => {});
+      await rmdir(symlinkDst, (err) => {});
+      await rmdir(symlinkDstRenamed, (err) => {});
```
ahtrotta, line 112: Why are you removing these before they've been created?

packages/cli/tests/unit/record/record.test.spec.ts
```diff
@@ -67,6 +79,64 @@ describe('record test', () => {
...
+        d: dirPrefix,
+      };
+
+      const ret = await RecordCommand.handler(argv);
+      expect(ret).toEqual(null);
+
+      await unlink(directoryParam, (err) => {});
+    });
+
+    it('shows error if directory parameter is invalid symlink', async () => {
```
ahtrotta, line 104: Same as above, this is a little misleading since you're not actually testing the output to the terminal.

ahtrotta (changes requested):
Thanks for doing this, I think it will definitely help users debug `ENOENT` errors. I just have a few small comments for the new tests.

packages/cli/tests/unit/record/record.test.spec.ts
symwell: If the test fails while developing and the test has to be rerun, the test can produce incorrect results if it doesn't start with a clean slate.

ahtrotta: :+1:

ahtrotta (approved):

PR #834, opened 2022-11-01
test: make sure Telemetry.sendEvent is stubbed

apotterri:
Fix some tests that weren't stubbing telemetry. Fixes getappmap/board#212.

### DIFF
--- a/packages/cli/tests/helper.ts
+++ b/packages/cli/tests/helper.ts
@@ -11,7 +11,7 @@ export function withSandbox() {
 
   return api;
 }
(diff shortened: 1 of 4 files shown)


dustinbyrne (approved):
The changes look good. I don't see any updates to `Telemetry.sendEvent` as indicated in the PR description however.

apotterri:
Yeah, I tried, but it broke too many tests. I pulled it out.

We might want to take a different approach, to avoid this problem in the future, but it seemed important to get this out sooner rather than later.

PR #836, opened 2022-11-03
ci: add sequence-diagram to smoketest

apotterri:


packages/cli/smoketest.sh
```diff
@@ -23,6 +23,7 @@ cd "${TESTDIR}"
 "./${BIN_NAME}" depends --appmap-dir ruby
 "./${BIN_NAME}" inventory --appmap-dir ruby
 "./${BIN_NAME}" openapi -d ruby -o /dev/null
+"./${BIN_NAME}" sequence-diagram -d ruby revoke_api_key.appmap.json
```
kgilpin, line 26: This is a good idea, but it was actually `@appland/appmap install` that failed in the wild last time.

apotterri: Yeah, I was just playing around to see if I could provoke a failure.

PR #837, opened 2022-11-04
fix: Don't run test commands that contain only whitespace

symwell:
Fixes https://github.com/getappmap/appmap-js/issues/802

### DIFF
--- a/packages/cli/src/cmds/record/prompt/obtainTestCommands.ts
+++ b/packages/cli/src/cmds/record/prompt/obtainTestCommands.ts
@@ -49,6 +49,8 @@ export default async function obtainTestCommands({ configuration }: RecordContex
           message: 'Test command (without env vars):',
         })
       )['testCommand'];
+      if (testCommand)
+        testCommand = testCommand.trim();
     } while (!testCommand);
 
     const { envVars } = await UI.prompt({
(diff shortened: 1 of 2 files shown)


packages/cli/src/cmds/record/prompt/obtainTestCommands.ts
```diff
@@ -49,6 +49,8 @@ export default async function obtainTestCommands({ configuration }: RecordContex
           message: 'Test command (without env vars):',
         })
       )['testCommand'];
+      if (testCommand)
+        testCommand = testCommand.trim();
     } while (!testCommand);
```
apotterri, line 54: Not sure why this was written with a `do ... while` loop, inquirer already has a way of enforcing valid input.

I think this would be better written as
```javascript
    const { testCommand } = await UI.prompt({
      name: 'testCommand',
      type: 'input',
      message: 'Test command (without env vars):',
      validate: (v) => (v && v.trim().length > 0 ? true : 'Please enter a test command'),
    });
```

apotterri (changes requested):
Seems like this might be better if it used the inquirer's built-in validation. LMK what you think of the change I suggested.

packages/cli/src/cmds/record/prompt/obtainTestCommands.ts
symwell: When I rewrite it this way some other part doesn't work, for what seems to be unrelated:

```
  ● record.prompt.obtainTestCommands › when commands are entered manually › test command must not be blank

    TypeError: Cannot read properties of undefined (reading 'toString')

      58 |
      59 |     let env = {};
    > 60 |     if (envVars.toString().trim() !== '') {
```

apotterri (approved):

PR #842, opened 2022-11-06
fix: Use exec semantic-release plugin to run yarn npm publish

### LINKED ISSUE #841: Publishing script doesn't work
The npm publish script seems to work, but fails silently.

dividedmind:
The custom script simply executed yarn and was broken (it didn't await for the command).

Fixes #841


apotterri (approved):
LGTM. Nice to have the publishing step be simplified.

PR #844, opened 2022-11-09
feat: Encourage users to commit their AppMap config.

symwell:
Add a prompt in the command-line installer that commits in Git any configuration files changed by appmap.

Addresses https://github.com/getappmap/board/issues/220

If the directory isn't a valid git repo, or if there's no valid `git` tool, the tool doesn't display this prompt.  There isn't any code that checks for the directory being valid or for git being present.  The calls to `git status -s` fail, the installer thinks nothing changed, and so it doesn't display the prompt.

Don't attempt to `git add appmap.yml` the first time the install runs.

Screenshots:

1) Installing in a directory that is a valid git repo

commit_hasgit

2) Installing in a directory that isn't a valid git repo. (I manually `rm -rf .git` in this dir.)

commit_doesnothavegit

3) Installing in a sub-directory.

commit_subdirectory

To test this:

```
git clone git@github.com:getappmap/appmap-js.git
cd appmap-js
git checkout sw/encourage_users_to_commit_their_appmap_config
yarn
yarn run build
cd packages/cli
npm start install -- -d ~/src/sample_app_6th_ed
```

### DIFF
--- a/packages/cli/src/cmds/agentInstaller/agentProcedure.ts
+++ b/packages/cli/src/cmds/agentInstaller/agentProcedure.ts
@@ -141,38 +141,48 @@ export default abstract class AgentProcedure {
     return files;
   }
 
-  async gitAdd(files: string[]): Promise<boolean> {
+  async gitAdd(files: string[]): Promise<any> {
     let params: string[] = ['add'];
     for (const file of files) {
       params.push(file);
     }
     try {
       const stdout = runSync(new CommandStruct('git', params, this.installer.path));
-      return true;
+      return {
+        success: true,
+        errorMessage: '',
+      };
     } catch (e) {
       const gitError = (e as Error).message.split('\n')[1];
-      const gitCommit = `[git commit failed, ${gitError}]`;
+      const gitAdd = `[git add failed, ${gitError}]`;
       // may want to print the error
+      return {
+        success: false,
+        errorMessage: gitAdd,
+      };
     }
-
-    return false;
   }
 
-  async gitCommit(files: string[], commitMessage: string): Promise<boolean> {
+  async gitCommit(files: string[], commitMessage: string): Promise<any> {
     let params: string[] = ['commit', '-m', commitMessage];
     for (const file of files) {
       params.push(file);
     }
     try {
       const stdout = runSync(new CommandStruct('git', params, this.installer.path));
-      return true;
+      return {
+        success: true,
+        errorMessage: '',
+      };
     } catch (e) {
       const gitError = (e as Error).message.split('\n')[1];
       const gitCommit = `[git commit failed, ${gitError}]`;
       // may want to print the error
+      return {
+        success: false,
+        errorMessage: gitCommit,
+      };
     }
-
-    return false;
   }
 
   async commitConfiguration(filesBefore: string[]) {
@@ -234,14 +244,20 @@ export default abstract class AgentProcedure {
         },
       });
 
-      // commitAdd is idempotent, and necessary to add to git
-      // appmap.yml, or even Gemfile.lock etc. when the installer runs
-      // for the first time
-      const commitAdd = await this.gitAdd(filesDiff);
-      const commitSuccess = await this.gitCommit(filesDiff, 'feat: Install AppMap.');
+      // gitAdd is idempotent, and necessary to add to git appmap.yml,
+      // Gemfile.lock etc. when the installer runs for the first time
+      const addGitReturn = await this.gitAdd(filesDiff);
+      if (!addGitReturn.success) {
+        UI.error(addGitReturn.errorMessage);
+      }
+
+      const commitGitReturn = await this.gitCommit(filesDiff, 'feat: Install AppMap.');
+      if (!commitGitReturn.success) {
+        UI.error(commitGitReturn.errorMessage);
+      }
 
       Telemetry.sendEvent({
-        name: commitSuccess
+        name: commitGitReturn.success
           ? `install-agent:commit_config:commit_success`
           : `install-agent:commit_config:commit_failure`,
         properties: {
@@ -249,7 +265,7 @@ export default abstract class AgentProcedure {
         },
       });
 
-      return commitSuccess;
+      return commitGitReturn.success;
     } else {
       Telemetry.sendEvent({
         name: `install-agent:commit_config:commit_no`,
--- a/packages/cli/src/cmds/agentInstaller/agentProcedure.ts
+++ b/packages/cli/src/cmds/agentInstaller/agentProcedure.ts
@@ -107,17 +107,6 @@ export default abstract class AgentProcedure {
     return validationResult;
   }
 
-  async filesLocallyModified(): Promise<string[]> {
-    let files: GitStatus[] = await this.gitStatus();
-    let filesModified: string[] = [];
-    files.filter((file) => {
-      if (file.status[0] !== '?') {
-        filesModified.push(file.file);
-      }
-    });
-    return filesModified;
-  }
-
   async gitStatus(file?: string): Promise<GitStatus[]> {
     let files: GitStatus[] = [];
     let params: string[] = ['status', '-s'];
@@ -152,6 +141,23 @@ export default abstract class AgentProcedure {
     return files;
   }
 
+  async gitAdd(files: string[]): Promise<boolean> {
+    let params: string[] = ['add'];
+    for (const file of files) {
+      params.push(file);
+    }
+    try {
+      const stdout = runSync(new CommandStruct('git', params, this.installer.path));
+      return true;
+    } catch (e) {
+      const gitError = (e as Error).message.split('\n')[1];
+      const gitCommit = `[git commit failed, ${gitError}]`;
+      // may want to print the error
+    }
+
+    return false;
+  }
+
   async gitCommit(files: string[], commitMessage: string): Promise<boolean> {
     let params: string[] = ['commit', '-m', commitMessage];
     for (const file of files) {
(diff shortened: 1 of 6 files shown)


packages/cli/src/cmds/agentInstaller/agentProcedure.ts
```diff
@@ -108,6 +107,222 @@ export default abstract class AgentProcedure {
...
+        file: 'appmap.yml',
+      },
+      { name: this.installer.name, file: this.installer.buildFile },
+    ];
+    // this is specific to Ruby
+    if (this.installer.name === 'Bundler')
+      filesWithInstaller.push({
+        name: this.installer.name,
+        file: 'Gemfile.lock',
+      });
```
symwell, line 217: Although the `buildFile` for Ruby is only `Gemfile`, we should also commit `Gemfile.lock` because this file changes too and people usually commit it too.

I thought of adding an extra get-ter in each installer called something like `configurationFiles` that returns an array of all files for each language.  WDYT?

symwell: Changed it to look at git status before and after we install.

Also verified it installs in sub-directories.

packages/cli/src/cmds/agentInstaller/agentProcedure.ts
```diff
@@ -108,6 +106,151 @@ export default abstract class AgentProcedure {
...
+          const file = line.substring(3);
+          if (gitStatus[0] !== '?') {
+            files.push(file);
+          }
+        }
+      } else {
+        gitStatus = '[no status]';
+      }
+    } catch (e) {
+      const gitError = (e as Error).message.split('\n')[1];
```
kgilpin, line 128: It's possible for this to catch non-Git-related errors. So it can be better to `try...catch` just the `runSync` command and then leave the rest of the code unguarded.

packages/cli/src/cmds/agentInstaller/agentProcedure.ts
```diff
@@ -108,6 +106,151 @@ export default abstract class AgentProcedure {
...
   }
 
+  async filesLocallyModified(): Promise<string[]> {
+    let files: string[] = [];
+
+    let gitStatus;
+    try {
+      const stdout = runSync(new CommandStruct('git', ['status', '-s'], this.installer.path));
+      if (stdout.length > 0) {
+        const lines = stdout.split('\n');
```
kgilpin, line 116: You may enjoy writing this in a functional style, like:

```
const files = stdout.split('\n').reduce((line) =>  { <line processing logic>; files.push(file) }).filter(Boolean)
```

`filter(Boolean)` removes null/undefined from the the Array, which will be placed there if there is an early return in the `reduce` function.

packages/cli/src/cmds/agentInstaller/agentProcedure.ts
```diff
@@ -108,6 +106,151 @@ export default abstract class AgentProcedure {
...
+    let gitCommit;
+    let isPartOfProject: boolean = false;
+    let itWorked: boolean = false;
+    try {
+      let params: string[] = ['commit', '-m', commitMessage];
+      for (const file of files) {
+        params.push(file);
+      }
+      const stdout = runSync(new CommandStruct('git', params, this.installer.path));
+      itWorked = true;
```
kgilpin, line 166: You can also simply `return true` from here.

packages/cli/src/cmds/agentInstaller/agentProcedure.ts
```diff
@@ -108,6 +106,151 @@ export default abstract class AgentProcedure {
...
+    } catch (e) {
+      const gitError = (e as Error).message.split('\n')[1];
+      gitCommit = `[git commit failed, ${gitError}]`;
+    }
+
+    return itWorked;
```
kgilpin, line 172: If you `return true` above then you can either `return false` here, or let it implicitly return `undefined` which will be falsey.

packages/cli/src/cmds/agentInstaller/agentProcedure.ts
```diff
@@ -108,6 +106,151 @@ export default abstract class AgentProcedure {
...
+      gitStatus = `[git status failed, ${gitError}]`;
+    }
+
+    return files;
+  }
+
+  async gitStatus(file: string): Promise<string> {
+    let gitStatus;
+    try {
+      const stdout = runSync(new CommandStruct('git', ['status', '-s', file], this.installer.path));
```
kgilpin, line 138: Since you are calling and parsing `git status` twice in this PR, you may want the same `gitStatus` function to be used in both cases, instead of having two places where you call and parse `git status`. The `file` argument can be optional.

packages/cli/src/cmds/agentInstaller/agentProcedure.ts
symwell: True.  Changed, here and in the other functions that use `runSync`.

packages/cli/src/cmds/agentInstaller/agentProcedure.ts
symwell: Refactored with `file` optional.

I had followed the pattern for [gitRemote](https://github.com/getappmap/appmap-js/pull/844/files#diff-f040a04bb00a8ecc8c0bedf2772c33579af265dbcdeb56e5d69ceadb4f721097L32) from line 32.  May want to refactor `gitRemote`.

packages/cli/src/cmds/agentInstaller/agentProcedure.ts
symwell: Simplified.

packages/cli/src/cmds/agentInstaller/agentProcedure.ts
symwell: I'm having trouble finding how to select in TS in a functional style.  I wanted to do something like this, in one statement.
```
  async filesLocallyModified(): Promise<string[]> {
    let files: GitStatus[] = await this.gitStatus();
    return files
      .filter((file) => file.status[0] !== '?')
      .reduce((file) => file.file);
  }
```

kgilpin: There are some examples here:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce

The function takes two arguments - the accumulator and the array entry

The second argument to the reduce function is the initial value - in this case, an empty Array.

symwell:
I believe this is ready to merge as is.

petecheslock:
```

> @appland/appmap@3.49.2 start /Users/petecheslock/repos/appmap-js/packages/cli
> ts-node src/cli.ts "install" "-d" "/Users/petecheslock/repos/appmap-merch-store"

? Multiple project types were found in /Users/petecheslock/repos/appmap-merch-store. Select one to continue. Bundler
Installing AppMap agent for appmap-merch-store...
  Project type: Bundler
  Project directory: /Users/petecheslock/repos/appmap-merch-store
  Git remote: origin    git@github.com:land-of-apps/solidus-demo.git (fetch)
  Ruby version: 2.6.6p146
  Gem home: /Users/petecheslock/.asdf/installs/ruby/2.6.6/lib/ruby/gems/2.6.0

  Is this correct? Yes
✔ Installing AppMap...
? AppMap recommends you have the installer commit in Git for you the following
  files, to not have to install AppMap again in this repository:
  Gemfile
  Gemfile.lock

  Commit? Yes
✔ Validating the AppMap agent...

   ╭─────────────────────────────────────────────────────────────────────╮
   │                                                                     │
   │              Success! AppMap has finished installing.               │
   │                                                                     │
   │                      NEXT STEP: Record AppMaps                      │
   │                                                                     │
   │   You can consult the AppMap documentation, or continue with the    │
   │     instructions provided in the AppMap code editor extension.      │
   │                                                                     │
   ╰─────────────────────────────────────────────────────────────────────╯
```


```
N 783ea87 (HEAD -> pc/add-appmap-test) feat: Install AppMap. [Pete Cheslock 3 minutes ago]
```

```
 git show 783ea87  
commit 783ea87ec529358eea4f2eab55c2e087c742bf64 (HEAD -> pc/add-appmap-test)
Author: Pete Cheslock <petecheslock@gmail.com>
Date:   Fri Nov 11 10:56:34 2022 -0500

    feat: Install AppMap.

diff --git Gemfile Gemfile
index 8d76b4f..50e1c05 100644
--- Gemfile
+++ Gemfile
@@ -3,6 +3,8 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }

 ruby '2.6.6'

+gem 'appmap', :groups => [:development, :test]
+
 gem 'rails', '~> 6.1'
 # Use pg as the database for Active Record
 gem 'pg'
diff --git Gemfile.lock Gemfile.lock
index 31ab0c7..35167b7 100644
--- Gemfile.lock
+++ Gemfile.lock
@@ -113,6 +113,11 @@ GEM
     apparition (0.6.0)
       capybara (~> 3.13, < 4)
       websocket-driver (>= 0.6.5)
+    appmap (0.93.5)
+      activesupport
+      method_source
+      rack
+      reverse_markdown
     autoprefixer-rails (10.3.1.0)
       execjs (~> 2)
     awesome_nested_set (3.4.0)
@@ -323,6 +328,8 @@ GEM
       http-cookie (>= 1.0.2, < 2.0)
       mime-types (>= 1.16, < 4.0)
       netrc (~> 0.8)
+    reverse_markdown (2.1.1)
+      nokogiri
     rspec-core (3.9.3)
       rspec-support (~> 3.9.3)
     rspec-expectations (3.9.4)
@@ -462,6 +469,7 @@ PLATFORMS

 DEPENDENCIES
   apparition
+  appmap
   binding_of_caller
   bootsnap (>= 1.1.0)
   bootstrap (~> 4.4.1)
```

packages/cli/src/cmds/agentInstaller/agentProcedure.ts
```diff
@@ -170,7 +170,9 @@ export default abstract class AgentProcedure {
       filesAfter.push(file.file);
     }
     let filesDiff: string[] = [];
-    filesDiff = filesAfter.filter((file) => !filesBefore.includes(file));
+    // don't try to commit appmap.yml because in a new installation
+    // it's not in git and causes git commit to fail
```
kgilpin, line 174: That is what `git add` is for.

kgilpin (changes requested):

ahtrotta:
If I run an old version of the installer:


image



It modifies `Gemfile`, `Gemfile.lock`, and adds `appmap.yml`. Then, if I run the new version of the installer, it doesn't ask if I want to commit these changes. This might be an unimportant edge case, but is this the expected behavior?


image

ahtrotta (approved):
I ran through a few scenarios and everything worked like I would expect it to (except the edge case I commented about). Based on manual testing it looks good to me, but I didn't review the code.

kgilpin (approved):

PR #847, opened 2022-11-10
feat: Add Telemetry event scan:started for beginning of analysis.

symwell:
Fixes https://github.com/getappmap/board/issues/201

Sets a timer that runs every 60 seconds to gather all queued telemetry messages, aggregate them, and send them as one message.

To test this run the following.  This extracts a [different branch](https://github.com/getappmap/appmap-js/compare/sw/emit_telemetry_event_upon_analysis__debug?expand=1) that adds [debugging statements](https://github.com/getappmap/appmap-js/compare/sw/emit_telemetry_event_upon_analysis...sw/emit_telemetry_event_upon_analysis__debug?expand=1) and sets the timer to 1 second but is otherwise the same.

```
# extract the branch with debugging statements
cd ~/src/appmap-js
git checkout sw/emit_telemetry_event_upon_analysis__debug

# in terminal 1 -- run the indexer
~/src/appmap-js/packages/cli$ yarn start index --watch --appmap-dir ~/src/sample_app_6th_ed/tmp/appmap/minitest

# in terminal 2 -- create appmaps
~/src/sample_app_6th_ed$ rm tmp/appmap/minitest/*;  APPMAP=true DISABLE_SPRING=true bundle exec rails test

# in terminal 3 -- run the scanner
~/src/appmap-js/packages/scanner$ yarn build && yarn run scanner scan --watch --appmap-dir ~/src/sample_app_6th_ed/tmp/appmap

```

You should see something like this: after 60 seconds the batched telemetry messages are sent and the batch array is emptied
aggregatedTelemetry_sendBatch

Example message:

```
{
  "name": "@appland/telemetry/scan:completed",
  "measurements": {
    "appland.telemetry.duration": 5.462,
    "appland.telemetry.numRules": 14,
    "appland.telemetry.numAppMaps": 67,
    "appland.telemetry.numFindings": 64
  },
  "properties": {
    "common.source": "@appland/telemetry",
    "common.os": "darwin",
    "common.platformversion": "21.5.0",
    "common.arch": "x64",
    "appmap.cli.machineId": "a97b99616932afb22e528243f52ae7a7e67862db5b9a9c745b9429d17215f187",
    "appmap.cli.sessionId": "c814cf48ac72de608bc2277ffa1846b94de6c582257a0a177763a1301209e601",
    "appland.telemetry.version": "1.0.0",
    "appland.telemetry.args": "/Users/kgilpin/source/appland/appmap-js/packages/scanner/src/cli.ts scan -d /Users/kgilpin/source/land-of-apps/sample_app_6th_ed --watch",
    "appland.telemetry.rules": "authz-before-authn, deserialization-of-untrusted-data, exec-of-untrusted-command, http-500, illegal-package-dependency, logout-without-session-reset, missing-authentication, missing-content-type, n-plus-one-query, secret-in-log, too-many-joins, too-many-updates, unbatched-materialized-query, update-in-get-request"
  }
}
```


packages/scanner/src/cli/scan/watchScan.ts
```diff
@@ -139,11 +141,18 @@ export class Watcher {
 
     const scanner = await buildScanner(true, this.config, [appmapFile]);
 
+    const startTime = Date.now();
+    Telemetry.sendEvent({
+      name: 'watchscan:started',
+    });
```
kgilpin, line 147: I'm glad that we are adding telemetry to `scan --watch`, but we are quickly going to regret sending a telemetry event for every single AppMap that is scanned. 

I would suggest that in `--watch` mode, telemetry should be batched up and sent periodically after there has been no activity for a defined period of time - like maybe 5 seconds - plus telemetry should be dumped on exit.

kgilpin (changes requested):

packages/scanner/src/cli/scan/watchScan.ts
symwell: By dumped on exit do you mean sent, or not sent?

packages/scanner/src/report/scanSummary.ts
```diff
@@ -20,3 +22,41 @@ export interface ScanSummary {
...
+  );
+  sourceScanSummary.appMapMetadata.recorders.forEach((value) =>
+    targetScanSummary.appMapMetadata.recorders.push(value)
+  );
+  sourceScanSummary.appMapMetadata.testStatuses.forEach((value) =>
+    targetScanSummary.appMapMetadata.testStatuses.push(value)
+  );
+  sourceScanSummary.appMapMetadata.exceptions.forEach((value) =>
+    targetScanSummary.appMapMetadata.exceptions.push(value)
+  );
```
dustinbyrne, line 61: I don't think it's worth capturing all of this AppMap metadata, especially without de-duplicating elements.

packages/scanner/src/report/scanSummary.ts
```diff
@@ -20,3 +22,41 @@ export interface ScanSummary {
...
 }
+
+export function aggregateScanSummary(
+  sourceScanSummary: ScanSummary,
+  targetScanSummary: ScanSummary
+) {
+  targetScanSummary.numAppMaps += sourceScanSummary.numAppMaps;
+  targetScanSummary.numChecks += sourceScanSummary.numChecks;
+  sourceScanSummary.rules.forEach((value) => targetScanSummary.rules.push(value));
+  sourceScanSummary.ruleLabels.forEach((value) => targetScanSummary.ruleLabels.push(value));
```
dustinbyrne, line 33: We should de-duplicate rules and labels

packages/scanner/src/cli/scan/watchScan.ts
```diff
@@ -44,13 +47,53 @@ function isAncestorPath(ancestor: string, descendant: string): boolean {
...
+    ) {
+      const telemetryName = 'watchscan:completed';
+      const firstElement: BatchedTelemetry = this.batchedTelemetry[0];
+      const aggregatedScanResults: ScanResults = new ScanResults(
+        firstElement.scanResults.configuration,
+        firstElement.scanResults.appMapMetadata,
+        firstElement.scanResults.findings,
+        firstElement.scanResults.checks
+      );
+      aggregatedScanResults.clear();
```
dustinbyrne, line 87: It's a little weird to be pulling out the first element to construct this object to then throw out the data (`clear`). Maybe the constructor should change to accept these arguments optionally? Or maybe we don't need this type?

dustinbyrne (changes requested):

packages/scanner/test/cli/watchScan.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+      msElapsed: 10,
+      msInserted: Date.now(),
+    });
+    watcher.sendScanResultsTelemetryInABatch();
+    expect(watcher.batchedTelemetry.length).toBe(1); // wasn't sent
+  });
+
+  it('sends telemetry events if the period elapsed', async () => {
+    const watcher: Watcher = new Watcher(options);
+    clearTimeout(watcher.interval); // disable the timer
```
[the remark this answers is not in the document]
dustinbyrne: I see now we're actually calling `unref` on the timer elsewhere so this `clearTimeout` isn't necessary to keep the tests from shutting down as I'd thought.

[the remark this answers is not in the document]
symwell: I'm trying to remove the unref on the timer.  I don't fully know the implications of calling unref on a timer in prod.

packages/scanner/src/cli/scan/watchScan.ts
symwell: It is.  Fixed.

packages/scanner/test/cli/watchScan.spec.ts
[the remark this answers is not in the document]
symwell: I think it's not worth turning the timer on and off.  It's unnecessary complexity.  A 5 second timer that runs one if statement isn't much of a performance overhead.

packages/scanner/src/report/scanSummary.ts
symwell: Deduplicated.

packages/scanner/src/report/scanSummary.ts
symwell: I went ahead and de-duplicated them.  I don't mind not aggregating them at all and sending a blank array.

symwell:
The extra `scan:completed` messages sent from CI stop being sent from CI when `commands.spec.ts` is removed, as tested in this [commit](https://github.com/getappmap/appmap-js/commit/d08bdd9825b9b455c2a3e995609197fde5fa155f) on a different branch.

This file uses `withStubbedTelemetry`. It should not have been sending messages.

symwell:
Fixed the telemetry issue.  This is ready to merge.

packages/scanner/test/cli/watchScan.spec.ts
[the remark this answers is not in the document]
dustinbyrne: I think it's better off in `src`. It'd be unexpected to have a hanging process as a result of having executed that code path in a test. 

`setInterval` is intended to run continuously until `clearInterval` has been called, freeing the timer. If the rest of the process logic has completed, the process will remain running until any outstanding timers have completed. In our case there's no condition under which the timer will ever be stopped, which is why the `unref` is necessary.

A `SIGINT` **does** appear to stop the timer which is likely why we don't notice this while running the script locally. However, if you run the following line in your terminal you'll notice the process won't by itself:
```sh
$ node -e 'setInterval(() => {}, 1000)'
```

With `clearTimeout` (which ideally would be `clearInterval`, but the behavior _appears_ to be the same) in the tests, the additional `unref` would not be necessary. However, we can't always guarantee there's a future point in which we stop the timer unless we do so alongside it's instantiation.

packages/scanner/test/cli/commands.spec.ts
```diff
@@ -49,6 +52,8 @@ describe('commands', () => {
           let watcher: watchScan.Watcher | undefined;
           jest.spyOn(watchScan, 'default').mockImplementation((opts) => {
             watcher = new watchScan.Watcher(opts);
+            clearTimeout(watcher.interval);
+            watcher.interval.unref(); // else jest doesn't exit
```
dustinbyrne, line 56: With `unref` in the `Watcher` constructor we can omit these in each test.

dustinbyrne (changes requested):

packages/scanner/src/cli/scan/watchScan.ts
```diff
@@ -44,13 +47,51 @@ function isAncestorPath(ancestor: string, descendant: string): boolean {
...
   appmapPoller?: chokidar.FSWatcher;
   configWatcher?: chokidar.FSWatcher;
+  batchedTelemetry: BatchedTelemetry[] = [];
+  sendPeriodMilliseconds = 5000;
+  interval: NodeJS.Timer;
+
+  constructor(private options: WatchScanOptions) {
+    this.interval = setInterval(() => {
+      this.sendScanResultsTelemetryInABatch();
+    }, this.sendPeriodMilliseconds);
```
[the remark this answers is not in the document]
symwell: I was unsure of `unref` having side effects.  It seems it doesn't.  Will put back in `src/`.

packages/scanner/src/report/scanSummary.ts
```diff
@@ -20,3 +22,64 @@ export interface ScanSummary {
...
+      )
+    ),
+  ];
+  targetScanSummary.appMapMetadata.exceptions = [
+    ...new Set(
+      targetScanSummary.appMapMetadata.exceptions.concat(
+        sourceScanSummary.appMapMetadata.exceptions
+      )
+    ),
+  ];
```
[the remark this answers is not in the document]
symwell: Now setting all `appMapMetadata` arrays to `[]`.

packages/scanner/test/cli/commands.spec.ts
symwell: Removed these `.unref` calls from each test.

packages/scanner/test/cli/watchScan.spec.ts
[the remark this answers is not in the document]
symwell: Did put `unref` in `src/`.  Used `clearInterval`.

packages/scanner/test/cli/commands.spec.ts
```diff
@@ -49,6 +52,7 @@ describe('commands', () => {
           let watcher: watchScan.Watcher | undefined;
           jest.spyOn(watchScan, 'default').mockImplementation((opts) => {
             watcher = new watchScan.Watcher(opts);
+            clearInterval(watcher.interval);
```
dustinbyrne, line 55: FWIW, we shouldn't need to call `clearInterval` anywhere due to the `unref`.

symwell: I tried [removing](https://github.com/getappmap/appmap-js/pull/847/commits/24843d05939c70e55e4c26fa19497fc76ef6ad78) the `clearInterval` and it sent telemetry messages on CI.  This makes me believe `withStubbedTelemetry` doesn't mock telemetry.

packages/scanner/src/cli/scan/watchScan.ts
```diff
@@ -44,13 +47,54 @@ function isAncestorPath(ancestor: string, descendant: string): boolean {
...
+      dateNow - this.batchedTelemetry[0].msInserted > this.sendPeriodMilliseconds
+    ) {
+      const aggregatedScanResults: ScanResults = new ScanResults(
+        {
+          checks: [],
+        },
+        {},
+        [],
+        []
+      );
```
dustinbyrne, line 88: This seems like a case for optional/default arguments in the constructor.

packages/scanner/src/report/scanSummary.ts
```diff
@@ -20,3 +20,29 @@ export interface ScanSummary {
...
+  targetScanSummary.appMapMetadata.labels = [];
+  targetScanSummary.appMapMetadata.apps = [];
+  targetScanSummary.appMapMetadata.clients = [];
+  targetScanSummary.appMapMetadata.frameworks = [];
+  targetScanSummary.appMapMetadata.git = [];
+  targetScanSummary.appMapMetadata.languages = [];
+  targetScanSummary.appMapMetadata.git = [];
+  targetScanSummary.appMapMetadata.recorders = [];
+  targetScanSummary.appMapMetadata.testStatuses = [];
+  targetScanSummary.appMapMetadata.exceptions = [];
```
dustinbyrne, line 47: re: default constructor arguments - on the other hand, this looks to me like this isn't the ideal type to be aggregating with.

packages/scanner/src/cli/scan/watchScan.ts
```diff
@@ -44,13 +47,54 @@ function isAncestorPath(ancestor: string, descendant: string): boolean {
...
+    this.interval.unref();
+  }
 
-  constructor(private options: WatchScanOptions) {}
+  public sendScanResultsTelemetryInABatch() {
+    const dateNow = Date.now();
+    if (
+      this.batchedTelemetry &&
+      this.batchedTelemetry.length > 0 &&
+      dateNow - this.batchedTelemetry[0].msInserted > this.sendPeriodMilliseconds
```
dustinbyrne, line 78: Instead of `this.batchedTelemetry[0].msInserted`, shouldn't this use the last element? This'll emit an event every five seconds without waiting for a period of inactivity.

packages/scanner/src/cli/scan/watchScan.ts
symwell: Added default arguments.

Optional arguments had cascading effects that spread to `src/` and `tests/`

packages/scanner/test/cli/commands.spec.ts
dustinbyrne: I stepped through with a debugger and can confirm `Telemetry.sendEvent` is stubbed. Maybe it's something else?

packages/scanner/src/cli/scan/watchScan.ts
symwell: "telemetry should be batched up and sent periodically after there has been no activity for a defined period of time - like maybe 5 seconds"

I see now that stated [goal](https://github.com/getappmap/appmap-js/pull/847#discussion_r1020316345) was to send data only after a period of inactivity.

The downside of sending data only after a period of inactivity is that if there's no inactivity it won't send data and possibly run out of memory (i.e. if the scanner is used in a high performance system that continually scans AppMaps that continually become available).  Whereas the current implementation guarantees data will always be sent periodically.

I'll change it to use the last element.

packages/scanner/test/cli/commands.spec.ts
dustinbyrne: Ah, I see. The mocks are fine, but with the `setInterval` it's possible for telemetry to be emit after the `afterEach` hook is run to restore the stubbed methods. I'm a little worried about this leaking telemetry into other test cases.

packages/scanner/src/report/scanSummary.ts
symwell: Now aggregating on ScanResults.

packages/scanner/src/cli/scan/watchScan.ts
symwell: Changed.

packages/scanner/test/cli/commands.spec.ts
symwell: I'm not sure how to ensure telemetry won't leak into other testcases.

I looked into (implemented but didn't push on this branch) enabling the timer when an event is added to the batch and disabling it when a batch is sent.  I think it can still leak telemetry into other testcases if the batch contains telemetry that didn't get sent.

packages/scanner/src/cli/scan/watchScan.ts
symwell: Moved to the entry point.

packages/scanner/src/cli/scan/watchScan.ts
```diff
@@ -44,16 +46,56 @@ function isAncestorPath(ancestor: string, descendant: string): boolean {
...
+    this.interval.unref();
+  }
 
-  constructor(private options: WatchScanOptions) {}
+  public sendScanResultsTelemetryInABatch() {
+    const dateNow = Date.now();
+    if (
+      this.batchedTelemetry &&
+      this.batchedTelemetry.length > 0 &&
+      dateNow - this.batchedTelemetry[this.batchedTelemetry.length - 1].msInserted >
```
kgilpin, line 79: What is the purpose of this clause? Why don't you just send all the telemetry that's in the Array?

kgilpin: This test seems wrong to me. If the last `batchedTelemetry` event is inserted at `dateNow` then `dateNow - dateNow > this.sendPeriodMilliseconds` will be false.

packages/scanner/test/cli/watchScan.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+    for (let i = 0; i < 10; i++) {
+      watcher.batchedTelemetry.push({
+        telemetryName: 'scan:complete',
+        scanResults: blankScanResults,
+        msElapsed: 10,
+        msInserted: Date.now() - watcher.sendPeriodMilliseconds - 1,
+      });
+    }
+    watcher.sendScanResultsTelemetryInABatch();
+    expect(watcher.batchedTelemetry.length).toBe(0); // they were all sent
```
kgilpin, line 51: There doesn't seem to be any test that actually asserts the data in the telemetry message?

kgilpin (changes requested):

packages/scanner/src/cli/scan/watchScan.ts
```diff
@@ -44,16 +46,56 @@ function isAncestorPath(ancestor: string, descendant: string): boolean {
...
   config?: TimestampedConfiguration;
   appmapWatcher?: chokidar.FSWatcher;
   appmapPoller?: chokidar.FSWatcher;
   configWatcher?: chokidar.FSWatcher;
+  batchedTelemetry: BatchedTelemetry[] = [];
+  sendPeriodMilliseconds = 5000;
+  interval: NodeJS.Timer;
+
+  constructor(private options: WatchScanOptions) {
+    this.interval = setInterval(() => {
```
kgilpin, line 66: There is a batcher here that doesn't rely on `setInterval`:

packages/cli/src/lib/eventAggregator.ts

packages/scanner/test/cli/watchScan.spec.ts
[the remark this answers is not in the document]
kgilpin: I've removed the need for all of this by:

a) Re-using (copying, unfortunately) the existing logic of EventAggregator, which uses `setTimeout` rather than `setInterval`
b) Making telemetry optional when the scanner is created - and disabling it in test cases

packages/cli/tests/unit/record/record.test.spec.ts
```diff
@@ -48,9 +42,10 @@ describe('record test', () => {
     cont = sinon.stub(UI, 'continue');
   });
 
-  afterEach(() => {
-    sinon.restore();
-  });
+  beforeEach(() => (tempDir = mkdtempSync(tmpdir())));
+
+  afterEach(() => sinon.restore());
```
dividedmind, line 47: Or just `afterEach(sinon.restore)`.

packages/scanner/src/cli/scan/watchScan.ts
```diff
@@ -46,49 +48,16 @@ function isAncestorPath(ancestor: string, descendant: string): boolean {
...
-      let totalMsElapsed = 0;
-      this.batchedTelemetry.forEach((item) => {
-        aggregatedScanResults.aggregate(item.scanResults);
-        totalMsElapsed += item.msElapsed;
-      });
-
-      sendScanResultsTelemetry(aggregatedScanResults, totalMsElapsed);
-      this.batchedTelemetry = [];
-    }
+    if (options.sendTelemetry) this.telemetry = new WatchScanTelemetry(this.scanEventEmitter);
```
dividedmind, line 60: Not sure if this option is really needed. Telemetry can be easily stubbed in tests.

packages/scanner/src/lib/eventAggregator.ts
```diff
@@ -0,0 +1,39 @@
+import { EventEmitter } from 'events';
+
+export type PendingEvent<E> = {
+  emitter: EventEmitter;
+  event: string;
+  args: E[];
```
dividedmind, line 6: This isn't the right typing — it's limited to events that have all arguments of the same type. Either change it to `args: E` (and `E` could be a `[ScanEvent]` single-element tuple) or to `arg: E` (to explicitly limit it to single-argument events).

packages/scanner/src/cli/scan/watchScanTelemetry.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export type ScanEvent = {
+  scanResults: ScanResults;
+  elapsed: number;
+};
+
+export class WatchScanTelemetry {
+  constructor(scanEvents: EventEmitter) {
+    new EventAggregator<ScanEvent>((events) => {
+      const scanEvents = events.map(({ args: [event] }) => event) as ScanEvent[];
```
dividedmind, line 13: This `as ScanEvent[]` shouldn't be necessary with the type parameter.

packages/scanner/src/lib/eventAggregator.ts
```diff
@@ -0,0 +1,39 @@
...
+  private pending: PendingEvent<E>[] = [];
+  private push(emitter: EventEmitter, event: string, args: E[]) {
+    this.pending.push({ emitter, event, args });
+    this.refresh();
+  }
+
+  private timeout?: NodeJS.Timeout;
+  private refresh() {
+    if (this.timeout) clearTimeout(this.timeout);
+    this.timeout = setTimeout(this.emitPending.bind(this), this.maxMsBetween);
```
dividedmind, line 27: I just learned that timeouts can be refreshed with `timeout.refresh()` without recreating them. Could be useful especially if there are a lot of events.

packages/scanner/src/report/scanResults.ts
```diff
@@ -81,7 +81,7 @@ export class ScanResults {
     return new ScanResults(this.configuration, this.appMapMetadata, findings, this.checks);
   }
 
-  aggregate(sourceScanResults: ScanResults) {
+  aggregate(sourceScanResults: ScanResults): void {
```
dividedmind, line 84: This could probably be dried up with watchScanTelemetry.

packages/scanner/test/cli/watchScanTelemetry.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+
+describe(watchScanTelemetry.WatchScanTelemetry, () => {
+  let emitter: EventEmitter;
+  let sinon: Sinon.SinonSandbox;
+  let _telemetry: watchScanTelemetry.WatchScanTelemetry;
+  let sendScanResultsTelemetry: Sinon.SinonStub;
+
+  beforeEach(() => jest.useFakeTimers());
+  beforeEach(() => (emitter = new EventEmitter()));
+  beforeEach(() => (sinon = Sinon.createSandbox()));
```
dividedmind, line 46: Is there a reason to create a new sandbox instead of using the global one?

dividedmind (approved):
Typing could be improved, other than that it looks ok.

packages/scanner/src/cli/scan/watchScan.ts
symwell: 1) The purpose is to send data after a period of inactivity.  Not to send data at a specific period.  If data are being collected for 10 minutes, and then there's inactivity for 5 seconds, data will be sent at 10:05.

I [believe](https://github.com/getappmap/appmap-js/pull/847#discussion_r1029606874) it's better to send data every 5 seconds instead of after a period of inactivity.

2) 
```
# this
dateNow - dateNew > this.sendPeriodMilliseconds

# reduces to
0 > 5000

# reduces to
false
```

kgilpin (approved):

symwell:
@dustinbyrne It needs your approval before it can be merged.

PR #853, opened 2022-11-14
fix: Ignore requests without path info

### LINKED ISSUE #845: OpenAPI generation fails parsing an AppMap
AppMap Extension Version `0.49.3`
Appmap-js version `3.49.2`

[Using Mastodon in Land-of-apps - Appmap branch](https://github.com/land-of-apps/mastodon/tree/appmap)

After running tests and generating appmaps - i attempted to generate an OpenAPI doc but got the following error.

```
Error: Error parsing tmp/appmap/rspec/ApplicationController_before_action_store_current_location_does_not_store_location_for_user_if_it_is_devise_controller.appmap.json: TypeError: Cannot read property 'toString' of undefined
    at OpenAPICommand.collectAppMap (/home/vscode/.npm/_npx/1609/lib/node_modules/@appland/appmap/built/cmds/openapi.js:47:19)
```

Relevant AppMap
[ApplicationController_before_action_store_current_location_does_not_store_location_for_user_if_it_is_devise_controller.appmap.json.zip](https://github.com/getappmap/appmap-js/files/9972966/ApplicationController_before_action_store_current_location_does_not_store_location_for_user_if_it_is_devise_controller.appmap.json.zip)

kgilpin:
HTTP server or client requests without path info should be ignored when generating OpenAPI. Verify this with a test using a wild AppMap.

See https://github.com/getappmap/appmap-js/pull/853/files#diff-e1ccded3fd99486c3c18cca624f1f45126b85e20a6383ae734e1ed7a01a78ff5R108

Fixes #845

### DIFF
--- a/packages/cli/tests/unit/fingerprint/fingerprintWatchCommand.spec.ts
+++ b/packages/cli/tests/unit/fingerprint/fingerprintWatchCommand.spec.ts
@@ -9,6 +9,7 @@ import { verbose } from '../../../src/utils';
 import OriginalTelemetry from '../../../src/telemetry';
 import { once } from 'events';
 import Fingerprinter from '../../../src/fingerprint/fingerprinter';
+import { MaxMSBetween } from '../../../src/lib/eventAggregator';
 
 jest.mock('../../../src/telemetry');
 const Telemetry = jest.mocked(OriginalTelemetry);
@@ -93,10 +94,10 @@ describe(FingerprintWatchCommand, () => {
       it('aggregates indexes that occur less than one second apart', async () => {
         placeMap();
         await once(handler, 'index');
-        jest.advanceTimersByTime(512);
+        jest.advanceTimersByTime(MaxMSBetween / 2.0);
         placeMap('../fixtures/ruby/user_page_scenario.appmap.json');
         await once(handler, 'index');
-        jest.advanceTimersByTime(2048);
+        jest.advanceTimersByTime(MaxMSBetween * 2.0);
 
         expect(Telemetry.sendEvent).toHaveBeenCalledTimes(1);
         const data = Telemetry.sendEvent.mock.calls[0][0];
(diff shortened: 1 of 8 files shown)


packages/openapi/src/rpcRequest.ts
```diff
@@ -102,11 +102,15 @@ class ClientRPCRequest implements RPCRequest {
 }
 
 function isServerEvent(event: Event): event is ServerRpcEvent {
-  return !!event.httpServerRequest && !!event.httpServerResponse;
+  return (
+    !!event.httpServerRequest &&
+    !!event.httpServerResponse &&
+    !!(event.httpServerRequest.normalized_path_info || event.httpServerRequest?.path_info)
```
kgilpin, line 108: This is actually the fix, here. The AppMap contains a request with `path_info: ''` and no `normalized_path_info`. `''` is falsey in JS.

packages/cli/tests/unit/fingerprint/fingerprintWatchCommand.spec.ts
```diff
@@ -114,10 +115,10 @@ describe(FingerprintWatchCommand, () => {
       it('send separate events for indexes that occur more than one second apart', async () => {
         placeMap();
         await once(handler, 'index');
-        jest.advanceTimersByTime(1024);
+        jest.advanceTimersByTime(MaxMSBetween * 1.1);
```
symwell, line 118: 1.1 here but 2 below looks inconsistent.  It should work with 1 too.

symwell (approved):

packages/cli/tests/unit/fingerprint/fingerprintWatchCommand.spec.ts
kgilpin: MaxMSBetween is `1000`, so I didn't understand the choice of 1024, 2048, 512 etc since we are working with time here and not bytes.

PR #858, opened 2022-11-16
feat: Sequence diagram diff upgrade

kgilpin:


kgilpin:
Changes incorporated into primary sequence diagram PR.

PR #859, opened 2022-11-16
fix: class can be at the the root of classmap

lachrist:
When I read the [spec](https://github.com/getappmap/appmap#classmap), I don't see that `class` code objects are forbidden from being at the root of classmap. I updated the JSON schema to let them be there.

### DIFF
--- a/packages/validate/src/macro.yml
+++ b/packages/validate/src/macro.yml
@@ -27,7 +27,9 @@ properties:
   classMap:
     type: array
     items:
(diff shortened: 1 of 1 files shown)


dividedmind (approved):

lachrist:
@dividedmind actually, classes cannot be at the root of classmap. Lemme close this PR.

> Classes are forbidden from being at the root of the class map
> I’ll update the spec description. There is a lot of code that depends on this invariant.
-- Kevin

PR #862, opened 2022-11-17
fix: Change copy in prompt and commit message used when configuring AppMap for a project.

symwell:
Addresses https://github.com/getappmap/board/issues/220

copy_new_prompt

copy_new_commit

### DIFF
--- a/packages/cli/src/cmds/agentInstaller/agentProcedure.ts
+++ b/packages/cli/src/cmds/agentInstaller/agentProcedure.ts
@@ -226,8 +226,9 @@ export default abstract class AgentProcedure {
       type: 'confirm',
       name: 'commit',
       message: [
-        `AppMap recommends you have the installer commit in Git for you the following`,
-        `  files, to not have to install AppMap again in this repository:`,
+        `Commit these files to your repo so that everyone on your team can use AppMap`,
+        `  without them having to repeat the setup process. Bring runtime code analysis`,
+        `  to your whole team!`,
         filesMessages,
         '  ',
         '  Commit?',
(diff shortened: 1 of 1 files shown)


packages/cli/src/cmds/agentInstaller/agentProcedure.ts
```diff
@@ -251,7 +252,14 @@ export default abstract class AgentProcedure {
         UI.error(addGitReturn.errorMessage);
       }
 
-      const commitGitReturn = await this.gitCommit(filesDiff, 'feat: Install AppMap.');
+      const commitGitReturn = await this.gitCommit(filesDiff, `feat: Configure AppMap for this project
+
+AppMap is a free and open-source runtime code analysis tool.
+
+AppMap has been installed and configured per installation instructions
```
brikelly, line 259: Maybe slightly clearer would be "...and configured using the installation instructions at:"

packages/cli/src/cmds/agentInstaller/agentProcedure.ts
```diff
@@ -251,7 +252,14 @@ export default abstract class AgentProcedure {
         UI.error(addGitReturn.errorMessage);
       }
 
-      const commitGitReturn = await this.gitCommit(filesDiff, 'feat: Install AppMap.');
+      const commitGitReturn = await this.gitCommit(filesDiff, `feat: Configure AppMap for this project
```
brikelly, line 255: I think we should drop the "feat: " part. It would be confusing to any team that doesn't regularly use it.

brikelly (changes requested):

packages/cli/src/cmds/agentInstaller/agentProcedure.ts
symwell: Dropped.

packages/cli/src/cmds/agentInstaller/agentProcedure.ts
symwell: Reworded.

PR #865, opened 2022-11-20
Better typescript support in models package

dividedmind:
- Change bundler from rollup to tsup.
- Add support for ts-jest.
- Add TypeScript linting.


packages/models/.eslintrc.cjs
```diff
@@ -24,5 +22,23 @@ module.exports = {
...
+        '@typescript-eslint/explicit-module-boundary-types': 'warn',
+        '@typescript-eslint/no-empty-function': 'warn',
+        '@typescript-eslint/no-this-alias': 'warn',
+        '@typescript-eslint/no-unsafe-argument': 'warn',
+        '@typescript-eslint/no-unsafe-assignment': 'warn',
+        '@typescript-eslint/no-unsafe-call': 'warn',
+        '@typescript-eslint/no-unsafe-member-access': 'warn',
+        '@typescript-eslint/no-unsafe-return': 'warn',
+        '@typescript-eslint/restrict-plus-operands': 'warn',
+        '@typescript-eslint/restrict-template-expressions': 'warn',
```
dustinbyrne, line 40: We're currently violating all of these?

dividedmind: The legacy javascript code is — we didn't typecheck these files before. That's why I turned these warnings off only for js files.

PR #874, opened 2022-11-26
Continue watching correctly after skipping files

dividedmind:
Fix a bug that would prevent watch scanning from working correctly once a couple files has been skipped (perhaps because they have been already up-to-date).

The problem was that `scan()` function had been modified to take and call a callback, but that change had omitted several early returns; in those cases, the callback would never have been called and the jobs would never finish and let the queue continue.

@ahtrotta , could you verify that this still resolves whatever issue #807 fixed?

### DIFF
--- a/packages/scanner/src/cli/scan/watchScan.ts
+++ b/packages/scanner/src/cli/scan/watchScan.ts
@@ -3,6 +3,7 @@ import * as chokidar from 'chokidar';
 import assert from 'assert';
 import path from 'path';
 import { queue } from 'async';
+import { callbackify } from 'node:util';
 
 import { formatReport } from './formatReport';
 import { default as buildScanner } from './scanner';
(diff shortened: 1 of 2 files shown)


dividedmind:
Note to self: retry build at https://github.com/getappmap/vscode-appland/pull/538 after this is merged.

kgilpin (approved):

kgilpin:
@dustinbyrne you should look at this too because I looked at #807 and I don't understand it.

packages/scanner/src/cli/scan/watchScan.ts
```diff
@@ -111,14 +112,16 @@ export class Watcher {
     );
   }
 
-  private queue = queue<string>(this.scan.bind(this), 2);
+  // do not remove callbackify, apparently on windows
+  // passing plain async function doesn't work (?)
+  private queue = queue<string>(callbackify(this.scan.bind(this)), 2);
```
dustinbyrne, line 117: Do you know why this is?

dustinbyrne (approved):

ahtrotta (approved):
I just tested and this seems to work fine on Windows.

packages/scanner/src/cli/scan/watchScan.ts
ahtrotta: After some digging around, I noticed that only the first N items were being processed on Windows, where N is the concurrency (in our case 2). I googled around and found this [SO post](https://stackoverflow.com/questions/57448099/async-queue-crashes-after-processing-3-initial-elements-of-the-queue-with-concur), and then I added the `callback` to the worker and then all of the items were processed.

kgilpin:
> I just tested and this seems to work fine on Windows.

@ahtrotta can you resolve the merge conflict and merge it?

PR #883, opened 2022-12-02
fix: Prevent EPERM/EACCES errors while watching AppMaps if a directory cannot be read

### LINKED ISSUE #880: `glob` fails to read a file or directory and causes the indexer to exit with error code 1
```
Error: EPERM: operation not permitted, scandir 'C:\dev\myproject'
Emitted 'error' event on Glob instance at:
at Glob._readdirError (globalStorage\appland.appmap\node_modules\glob\glob.js:610:14)
at globalStorage\appland.appmap\node_modules\glob\glob.js:553:12
at FSReqCallback.oncomplete (node:fs:188:23) {
errno: -4048,
code: 'EPERM',
syscall: 'scandir',
path: 'C:\\dev\\myproject'
}
```

dustinbyrne:
If the user does not have permissions to read a directory under `--appmap-dir`, an `EPERM`/`EACCES` error would panic the process (even in `--watch` mode).

Instead, these errors will be ignored and we'll continue scanning what we can.

I wasn't sure of an easy way to test an `EPERM` without involving other OS users, so I've omitted the test for now. Behavior should otherwise be the same.

Resolves #880

### DIFF
--- a/packages/cli/src/fingerprint/globber.ts
+++ b/packages/cli/src/fingerprint/globber.ts
@@ -3,6 +3,7 @@ import { Glob, IGlob } from 'glob';
 import { assert } from 'console';
 import { stat } from 'fs/promises';
 import { Stats } from 'fs-extra';
+import { verbose } from '../utils';
 
 class Globber extends EventEmitter {
   constructor(private pattern: string, public interval = 1000) {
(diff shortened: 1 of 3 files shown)


packages/cli/src/fingerprint/globber.ts
```diff
@@ -14,7 +15,11 @@ class Globber extends EventEmitter {
   public scanNow() {
     assert(!this.currentGlob);
     this.timeout = undefined;
-    this.currentGlob = new Glob(this.pattern, { ignore: ['**/node_modules/**', '**/.git/**'] })
+    this.currentGlob = new Glob(this.pattern, {
+      ignore: ['**/node_modules/**', '**/.git/**'],
+      strict: false,
```
apotterri, line 20: `strict` defaults to `false`, so this has no effect.

apotterri (changes requested):
I think this really needs to include a test.

dustinbyrne:
I've added a test which fails prior to this fix. Turns out it was more simple than expected :+1:

packages/cli/src/fingerprint/globber.ts
apotterri: Sorry about that, looks like their doc needs an update: https://github.com/isaacs/node-glob#options .

apotterri (approved):
Thanks for the test, looks good.

apotterri:
Build's red, though,

PR #885, opened 2022-12-05
Faster smoke testing and other build improvements

dividedmind:


ahtrotta (approved):
Thanks for doing this!

PR #887, opened 2022-12-05
Feat/finding info in appmap

### LINKED ISSUE #876: Finding details are added to map view in IDE
Analysis finding information is added into the map-view in the IDE.

## Findings
Image

## Finding details
Image

ahtrotta:
Fixes #876 
Replaces #868

Now findings information can be viewed within an AppMap (if it has one or more associated findings).

I haven't written tests for this change yet, but I'm planning on doing that as my first task in the new year.

Each finding appears in the details panel:

image

When clicked, the trace filter gets populated and the user can iterate through the events associated with the clicked finding:

image

Each event also displays more detailed information about the finding in the details panel:

image


packages/components/src/components/DetailsSearch.vue
```diff
@@ -264,57 +319,81 @@ export default {
...
     }
 
     &-list {
       margin: 0;
-      padding: 0;
+      padding: 0 0.5rem;
       list-style: none;
+      li {
+        &:last-of-type {
+          border-bottom: 0;
```
dustinbyrne, line 374: This makes the last label look like it's being clipped
image

packages/components/src/components/DetailsSearch.vue
```diff
@@ -179,6 +210,13 @@ export default {
     selectObject(type, object) {
       if (type === 'labels') {
         this.$store.commit(SELECT_LABEL, object);
+      } else if (type === 'analysis-finding') {
+        const { events } = object;
+        if (events && events.length >= 1) {
+          this.$store.commit(SELECT_OBJECT, events);
```
dustinbyrne, line 216: I'd expect to be brought to finding details, not an event. A finding may consist of many events and the finding details panel should help us navigate them.

packages/components/src/components/FindingDetailsPanel.vue
```diff
@@ -0,0 +1,274 @@
...
+    padding: 0 1rem;
+    .code {
+      color: $white;
+      margin-top: 0;
+      font-family: monospace;
+    }
+  }
+  .heading,
+  section {
+    border-bottom: 3px solid $gray2;
```
dustinbyrne, line 144: There's some inconsistency here with the border width. The outer border is `1px` while the inner is `3px`. We're also not adding an outer border anywhere else, but I think this is more of an issue of the finding details needing to be its own panel.
image

packages/components/src/components/DetailsSearch.vue
dustinbyrne: I'm not sure if this is new or not, but the labels also differ between `DetailsSearch` and `DetailsPanelEvent`
image

packages/components/src/components/FindingDetailsPanel.vue
dustinbyrne: The margins look like they could be more consistent here as well

packages/components/src/components/DetailsSearch.vue
```diff
@@ -150,6 +158,29 @@ export default {
...
+        const { traceFilter } = fragment;
+
+        const events = fragment.traceFilter.split(' ').map((idStr) => {
+          const id = Number(idStr.split(':')[1]);
+          return eventsById[id];
+        });
+
+        items['analysis-finding'].data.push({
+          object: {
+            name: `${finding.impactDomain}:${finding.ruleTitle}`,
```
dustinbyrne, line 176: Missing a space (e.g. `Performance: N plus 1 SQL query`).

packages/components/src/components/DetailsSearch.vue
```diff
@@ -150,6 +158,29 @@ export default {
...
+        return map;
+      }, {});
+
+      this.findings.forEach(({ finding, appMapUri: { fragment } }) => {
+        const { traceFilter } = fragment;
+
+        const events = fragment.traceFilter.split(' ').map((idStr) => {
+          const id = Number(idStr.split(':')[1]);
+          return eventsById[id];
+        });
```
dustinbyrne, line 172: IMO one of the major drivers for this feature is to replace the use of the trace filter for navigating events within a finding. We should drop this.

packages/components/src/components/DetailsSearch.vue
```diff
@@ -264,57 +319,81 @@ export default {
...
     &-list {
       margin: 0;
-      padding: 0;
+      padding: 0 0.5rem;
       list-style: none;
+      li {
+        &:last-of-type {
+          border-bottom: 0;
+        }
+      }
```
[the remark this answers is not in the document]
dustinbyrne: It's a little more consistent now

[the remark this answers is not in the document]
dustinbyrne: ...maybe?

image
vs
image

packages/diagrams/src/scss/components/_diagram.scss
```diff
@@ -108,7 +108,7 @@
         }
 
         &.database {
-          fill: $database;
+          fill: red;
```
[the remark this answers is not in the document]
typaulhus: I pushed some fixes for the styling pieces you mentioned. :)

dustinbyrne (approved):

PR #893, opened 2022-12-08
fix: Increase number of open file descriptors limit when indexer starts

symwell:
When the indexer starts, call `setrlimit` to increase the limit of open file descriptors to a high value.  If this doesn't succeed try a few lower values.  The limit effects only the indexer process until the process exits.  It doesn't effect other processes.

Sample output on Linux:
indexer_changes_limit

The node module `posix`, which provides `setrlimit`, wasn't installing on CI on Windows, giving the following error.  Changed CI for Windows to install python3 before `yarn run build`.
```
➤ YN0000: │ posix@npm:4.2.0 STDERR gyp ERR! find Python 
➤ YN0000: │ posix@npm:4.2.0 STDERR gyp ERR! find Python Python is not set from command line or npm configuration
➤ YN0000: │ posix@npm:4.2.0 STDERR gyp ERR! find Python Python is not set from environment variable PYTHON
➤ YN0000: │ posix@npm:4.2.0 STDERR gyp ERR! find Python checking if "python3" can be used
➤ YN0000: │ posix@npm:4.2.0 STDERR gyp ERR! find Python - "python3" is not in PATH or produced an error
➤ YN0000: │ posix@npm:4.2.0 STDERR gyp ERR! find Python checking if "python" can be used
➤ YN0000: │ posix@npm:4.2.0 STDERR gyp ERR! find Python - executable path is "C:/ProgramData/chocolatey/lib/mingw/tools/install/mingw64/opt/bin/python.exe"
➤ YN0000: │ posix@npm:4.2.0 STDERR gyp ERR! find Python - version is "2.7.9"
➤ YN0000: │ posix@npm:4.2.0 STDERR gyp ERR! find Python - version is 2.7.9 - should be >=3.6.0
➤ YN0000: │ posix@npm:4.2.0 STDERR gyp ERR! find Python - THIS VERSION OF PYTHON IS NOT SUPPORTED
```

Fixes https://github.com/getappmap/appmap-js/issues/892

I was unable to reproduce this `EMFILE` error.  Only the `ENOSPC` error was getting raised.  This made me believe the root cause isn't file watching but a user having a lot of file descriptors open in their session.

The reason to increase the limit, rather than catch `EMFILE` and disable file watching, falling back to file polling only, is that if the user is at the limit then file polling may not work either.

Maybe it should do both, increase the limit and catch `EMFILE`.


symwell:
**This doesn't work on Windows**.  There's no support for `setrlimit` on Windows in the `posix` or `node-posix-ext` modules.  Also, the `posix` module doesn't get imported on Windows, as shown on [CI](https://app.travis-ci.com/github/getappmap/appmap-js/jobs/591101009).

I plan to close this pr and catch `EMFILE`.

PR #895, opened 2022-12-09
fix: Disable file watching on "ENOSPC: System limit for number of file watchers reached" error

symwell:
When the indexer processes an appmap_dir that has more files than the OS limit for watching files, FSWatch raises an ENOSPC error.  Catch this specific type of ENOSPC error and disable file watching.  File changes continue to be detected through file polling.

Output of manual test. After disabling file watching, file polling continues to run.
file_watching_disabled

Fixes https://github.com/getappmap/appmap-js/issues/894

If the reason for watching too many files was that the directory monitored wasn't precisely the `appmap_dir` but something more general, like `.` monitoring two parent directories up, then saving the `appmap_dir` in `appmap.yml` through [this pr](https://github.com/getappmap/appmap-js/pull/896) may be enough to make this ENOSPC error disappear without this fix.

This fix should still be applied because choosing to discard error messages instead of crash is still needed, as shown in https://github.com/getappmap/appmap-js/pull/898

### DIFF
--- a/packages/cli/src/fingerprint/fingerprintWatchCommand.ts
+++ b/packages/cli/src/fingerprint/fingerprintWatchCommand.ts
@@ -46,7 +46,7 @@ export default class FingerprintWatchCommand {
   }
 
   async watcherErrorFunction (error: Error) {
-    if (error.message.includes("ENOSPC: System limit for number of file watchers reached")) {
+    if (this.watcher && error.message.includes("ENOSPC: System limit for number of file watchers reached")) {
       console.warn(error.stack);
       console.warn("Will disable file watching. File polling will stay enabled.");
       await this.watcher?.close();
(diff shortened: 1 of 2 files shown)


packages/cli/src/fingerprint/fingerprintWatchCommand.ts
```diff
@@ -45,6 +45,19 @@ export default class FingerprintWatchCommand {
     }
   }
 
+  async watcherErrorFunction (error: Error) {
+    if (error.message.includes("ENOSPC: System limit for number of file watchers reached")) {
```
ahtrotta, line 49: We should check to see if the watcher was already removed before catching the error. This line could be:
```
if (this.watcher && error.message.includes("ENOSPC: System limit for number of file watchers reached")) {
```

packages/cli/src/fingerprint/fingerprintWatchCommand.ts
```diff
@@ -68,7 +81,8 @@ export default class FingerprintWatchCommand {
     })
       .on('add', this.added.bind(this))
       .on('change', this.changed.bind(this))
-      .on('unlink', this.removed.bind(this));
+      .on('unlink', this.removed.bind(this))
+      .on('error', error => this.watcherErrorFunction(error));
```
ahtrotta, line 85: Can you change this to better match the other event handler functions?
```
.on('error', this.watcherErrorFunction.bind(this));
```

ahtrotta (approved):

PR #898, opened 2022-12-12
fix: Don't crash indexer on an "UNKNOWN: unknown error, lstat" error

symwell:
Attempts to fix https://github.com/getappmap/appmap-js/issues/897. Created from branch `sw/disable_watching_on_ENOSPC_error`

This error is an exception thrown by chokidar while file watching.

I wasn't able to reproduce this error.  I created symlinks and deleted them programmatically when the queue processes the file but this produced an ENOENT error.
```
File /home/test/tmp/too_many_open_files/minitest/Static_pages_controller_should_get_help.appmap.json does not exist or is not a file.
Deleted File successfully.
[Error: ENOENT: no such file or directory, lstat '/home/test/tmp/too_many_open_files/minitest/Static_pages_controller_should_get_home.appmap.json'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'lstat',
  path: '/home/test/tmp/too_many_open_files/minitest/Static_pages_controller_should_get_home.appmap.json'
}
```

There's a claim this `UNKNOWN: unknown error, lstat` error is the result of file corruption down at the NTFS level.
https://github.com/nodejs/node/issues/43439

### DIFF
--- a/packages/cli/src/fingerprint/fingerprintWatchCommand.ts
+++ b/packages/cli/src/fingerprint/fingerprintWatchCommand.ts
@@ -59,6 +59,8 @@ export default class FingerprintWatchCommand {
             errorStack: error.stack,
           },
         });
+    } else if (error.message.includes("UNKNOWN: unknown error, lstat")) {
+      console.warn(error.stack);
     } else {
       // let it crash if it's some other error, to learn what the error is
       throw error;
--- a/packages/cli/src/fingerprint/fingerprintWatchCommand.ts
+++ b/packages/cli/src/fingerprint/fingerprintWatchCommand.ts
@@ -61,6 +61,13 @@ export default class FingerprintWatchCommand {
         });
     } else if (error.message.includes("UNKNOWN: unknown error, lstat")) {
       console.warn(error.stack);
+      Telemetry.sendEvent({
+          name: `index:watcher_error:unknown`,
+          properties: {
+            errorMessage: error.message,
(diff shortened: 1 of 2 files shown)


ahtrotta:
I don't think we should catch this error if we don't understand why it's happening. It's possible it could be a [UTF-8 problem](https://github.com/paulmillr/chokidar/issues/891). Or it could be the corrupt file like you linked to above. Or it might have to do with a faulty [npm cache](https://stackoverflow.com/a/62044109). Bottom line is, I think we shouldn't mess with it until we understand it. It does seem like it's a Windows problem, so maybe someone with a Windows machine should tackle this.

dustinbyrne:
I'd be fine with not propagating unknown errors into a panic as long as we emit the exception as a telemetry event:
https://github.com/getappmap/appmap-js/blob/dd80f9258f6dc75838b97e9cb54ae6048c0be673/packages/cli/src/fingerprint/fingerprintWatchCommand.ts#L58-L59

Is there a chance this method would be invoked repeatedly on an unreadable file? If so, that's a bigger issue.

symwell:
Added the ability to save unreadable files in a Set and ignore them.

Added logic in testcases.  Also tested by copying and executing this logic for `ENOSPC` which is easily reproducible.  It showed files were added to `unreadableFiles` correctly.
unreadableFiles

If we throw error then the indexer will exit.

ahtrotta (approved):

PR #912, opened 2022-12-15
fix: Don't let scanner crash on `EACCES: permission denied` error

symwell:
Before this change, the scanner crashed when it encountered a directory that didn't have read permission.
scanner_EACCESS_before

After this change, the scanner doesn't crash.
scanner_EACCES_after

This crash is probably caused by using a general path for `appmap_dir` instead of the expected `tmp/appmap`, along with the scanner not handling this exception.  The fix is to set `ignorePermissionErrors: true` in chokidar's options.

There's a testcase.  When `ignorePermissionErrors` is commented out the testcase fails, as expected.

Fixes https://github.com/getappmap/appmap-js/issues/899.

### DIFF
--- a/packages/scanner/test/cli/scan.spec.ts
+++ b/packages/scanner/test/cli/scan.spec.ts
@@ -16,6 +16,7 @@ import tmp from 'tmp-promise';
 import assert from 'assert';
 import { withStubbedTelemetry } from '../helper';
 import { FSWatcher } from 'chokidar';
+import { mkdir, chmod } from 'fs/promises';
 
 process.env['APPMAP_TELEMETRY_DISABLED'] = 'true';
 delete process.env.APPLAND_API_KEY;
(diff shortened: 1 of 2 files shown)


packages/scanner/test/cli/scan.spec.ts
```diff
@@ -288,6 +289,17 @@ describe('scan', () => {
...
 
+    it('does not raise on EACCES: permission denied', async () => {
+      await createIndex(secretInLogMap);
+      const permissionDeniedDir = join(tmpDir, 'permission_denied_dir');
+      await mkdir(permissionDeniedDir);
+      await fsextra.copy(secretInLogMap, join(permissionDeniedDir, basename(secretInLogMap)));
+      chmod(permissionDeniedDir, 0o000);
+      await createWatcher();
+      await waitForSingleFinding();
+      chmod(permissionDeniedDir, 0o777); // else the next testcase fails
```
dustinbyrne, line 300: This should (?) be deterministic but it'd be good to `await` these `chmod` calls in case something changes in the future.

dustinbyrne (approved):

PR #922, opened 2022-12-21
fix: Don't encourage `git commit` for `pip`

symwell:
When adding AppMap with the installer to a project that uses `pip`, don't prompt the user to commit the changes to git. It would affect production because `appmap-python` will soon [record by default](https://github.com/getappmap/appmap-python/pull/208) in production.  Affecting production is not what the user would want.

For `pipenv` and `poetry` this isn't necessary because they provide a way to add AppMap to a development configuration.

In general the installers go out of their way to ensure their respective agents don't get installed in production.

Before this change, installing in a `pip` project prompted the user to commit to git.
installer_pip_before

After this change, installing in a `pip` project doesn't prompt the user to commit to git.
installer_pip_after

Tested with repo [`git@github.com:symwell/try_appmap_python.git`](https://github.com/symwell/try_appmap_python)

### DIFF
--- a/packages/cli/src/cmds/agentInstaller/agentProcedure.ts
+++ b/packages/cli/src/cmds/agentInstaller/agentProcedure.ts
@@ -191,6 +191,18 @@ export default abstract class AgentProcedure {
       properties: {},
     });
 
+    // don't commit if Python with pip; it would enable AppMap in production
+    if (this.installer.name === 'pip') {
+      Telemetry.sendEvent({
+        name: `install-agent:commit_config:should_not_commit`,
+        properties: {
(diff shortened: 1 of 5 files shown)


kgilpin (changes requested):
Why will appmap-python soon record by default in production?

apotterri:
I think there's a better change that can be made here. During installation in a `pip` project, we should ask the user which file they'd like us to put `appmap` in, preferably with some language indicating that it's a dev dependency.

symwell:
Closing since https://github.com/getappmap/appmap-js/pull/925 was merged instead.

PR #924, opened 2022-12-22
feat: remove fonky default test command

lachrist:
The default js test command seems to record 0 appmaps in many cases. It  could be that there is no tests or that the project is not using node to  run tests. Also, the command is confusing because it mixes  `--recorder=process` with a test command whereas we want to push ad-hoc  recorders for tests.


dustinbyrne (approved):
I'm not sure why the build is failing but I feel strongly it's unrelated to this change. There's definitely a case for using `--recorder=process`, but I agree this isn't the right time for it.

:+1:

PR #925, opened 2022-12-22
fix: For pip, prompt user to provide requirements file for dev

symwell:
If multiple requirements files are present in a pip project, prompt the user to choose the one used during development and testing.

Without this change AppMap would be enabled for production and this is probably not what the user wants.

This pr is a better path forward compared to https://github.com/getappmap/appmap-js/pull/922.

The user will see this when there are multiple files that match the glob `*requirements*.txt`: 
image


packages/cli/src/cmds/agentInstaller/pythonAgentInstaller.ts
```diff
@@ -146,26 +150,68 @@ export class PipenvInstaller extends PythonInstaller {
...
+        type: 'list',
+        name: 'buildFile',
+        message: 'Please choose the requirements file used for development:',
+        choices,
+        default: defaultChoice[0],
+      },
+    ]);
+
+    return basename(buildFile);
   }
```
dustinbyrne, line 198: Should these be marked as `private`? We don't typically use a leading underscore to imply visibility, especially when the method is actually public.

If it's a matter of accessibility within tests, something like one of the following would work:
```ts
(installer as unknown) as { _chooseBuildFile(): Promise<string> }
(installer as unknown) as { _findBuildFiles(): Promise<Array<string>> }
```

```ts
interface PythonInstallerInternal {
  _chooseBuildFile() => Promise<string>;
  _findBuildFiles() => Promise<Array<string>>;
}

...

installer as unknown as PythonInstallerInternal
```

Depending on the TS config you may even be able to get away with simply
```ts
installer['_findBuildFiles']
```

packages/cli/src/utils.ts
```diff
@@ -32,25 +32,33 @@ export function baseName(fileName: string) {
...
  * is not a file.
  */
+async function statFile(filePath: PathLike): Promise<Stats | null> {
+  try {
+    return await fsp.stat(filePath);
+  } catch (e) {
+    return null;
+  }
+}
+
```
dustinbyrne, line 42: It looks like this function was inserted between another function and its JSDoc comment. The JSDoc for `statFile` is incorrect.

packages/cli/src/utils.ts
```diff
@@ -32,25 +32,33 @@ export function baseName(fileName: string) {
...
-  }
-  if (!fileStat.isFile()) {
+  const fileStat = await statFile(filePath);
+  if (!fileStat || !fileStat.isFile()) {
     return null;
   }
   return fileStat.mtimeMs;
 }
 
+export async function isFile(filePath: PathLike): Promise<boolean | null> {
```
dustinbyrne, line 57: Nitpick, but do we need to return `null` here? It won't fit neatly into a boolean without the caller casting the result.

dustinbyrne (changes requested):
Looks good. I left some minor comments below.

packages/cli/src/cmds/agentInstaller/pythonAgentInstaller.ts
apotterri: Yeah, they should be `private`.  The leading underscore was spillover from python. ;)

packages/cli/src/utils.ts
apotterri: Thanks, I'll fix this.

packages/cli/src/utils.ts
apotterri: No, we don't. I'll fix it.

dustinbyrne (approved):

PR #928, opened 2023-01-04
fix: Fix a bug preventing 'Update in GET request' rule from working

dividedmind:
The rule was missing a correct `enumerateScope` setting.

### DIFF
--- a/packages/scanner/src/rules/updateInGetRequest.ts
+++ b/packages/scanner/src/rules/updateInGetRequest.ts
@@ -74,6 +74,7 @@ export default {
   id: 'update-in-get-request',
   title: 'Data update performed in GET or HEAD request',
   scope: 'http_server_request',
+  enumerateScope: true,
   labels: [Audit],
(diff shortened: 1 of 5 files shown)


kgilpin (commented):
Thanks for adding this test. Since it's so easy to add test data, it would improve my comfort level with the PR if the test cases also included mixed event types, for example regular function events mixed in with the SQL, to show that the rule will work properly on an AppMap that is more realistic (e.g. it won't blow up if it sees non-SQL events).

dividedmind:
> Thanks for adding this test. Since it's so easy to add test data, it would improve my comfort level with the PR if the test cases also included mixed event types, for example regular function events mixed in with the SQL, to show that the rule will work properly on an AppMap that is more realistic (e.g. it won't blow up if it sees non-SQL events).

Added in 47ed717

kgilpin (approved):
Nice! Now please migrate our other tests to this system :-)

PR #932, opened 2023-01-09
feat: Collect git metadata

kgilpin:
Collect and emit git metadata in the Fingerprint watcher. So, for anyone who has `git` and runs a current indexer, we should get the git telemetry. 

The `git` commands are borrowed from appmap-ruby.

### DIFF
--- a/packages/cli/src/lib/flattenMetadata.ts
+++ b/packages/cli/src/lib/flattenMetadata.ts
@@ -2,7 +2,7 @@ import { Metadata } from '@appland/models';
 import sanitizeURL from './repositoryInfo';
 
 /** Flattens metadata into a string-string map suitable for use in telemetry.
- * Ignores git, exception and fingerprints.
+ * Ignores exception and fingerprints.
  */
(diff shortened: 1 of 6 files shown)


packages/cli/src/lib/flattenMetadata.ts
```diff
@@ -26,8 +26,14 @@ export default function flattenMetadata(metadata: Metadata): Map<string, string>
     if (metadata.language.engine) result.set('language.engine', metadata.language.engine);
   }
 
-  if (metadata.git?.repository) {
-    result.set('git.repository', sanitizeURL(metadata.git.repository));
+  if (metadata.git) {
+    Object.keys(metadata.git as any).forEach((key) => {
+      const saveKey = key.replace(/^git_/, '');
```
kgilpin, line 31: `appmap-ruby` emits `git_` keys for some reason. Normalize these.

kgilpin:
I've returned this to a Draft state so we can further discuss how we want this to work before merging it.

kgilpin:
Not going to merge at this time. Can revisit later if needed.

PR #936, opened 2023-01-10
feat: Detect yarn packages as sub-projects

symwell:
Before this change, installing in projects that used `yarn` with multiple packages didn't detect the packages as sub-projects. Installation gave an error:
subprojects_before_failed

After this change, installation detects packages as sub-projects.
subprojects_after_select

Note **installation doesn't succeed** (on my machine).  It gives a [different error](https://user-images.githubusercontent.com/111290954/211630706-b129b9f5-8019-40c9-8603-eb36defd303a.png). This error exists on the `main` branch too and [isn't resolved](https://github.com/getappmap/appmap-js/issues/797#issuecomment-1376080238) by this pr. We believe running `npm exec` instead of `npx` in the agent installation step addresses this.

Fixes https://github.com/getappmap/appmap-js/issues/797


dustinbyrne:
I'm not really understanding the error. Does `npx @appland/appmap-agent-js init .` work when not launched from the installer? The fact that this is a yarn workspace shouldn't make a difference.

packages/cli/src/cmds/agentInstaller/projectConfiguration.ts
```diff
@@ -81,12 +84,42 @@ async function resolveSelectedInstallers(
...
   }
 }
 
+export function getYarnPackages(availableInstallers: AgentInstaller[]): string[] {
+  let yarnPackages: string[] = [];
+  availableInstallers.forEach((installer) => {
+    if (installer.name === 'yarn') {
+      const packageJsonFilename = join(installer.path, 'package.json');
+      if (existsSync(packageJsonFilename)) {
+        const data = readFileSync(packageJsonFilename, 'utf-8');
```
dustinbyrne, line 99: I'm nit-picking here, but we should be opting for the asynchronous versions of these methods. It'll allow other work to be processed while it hits the disk.

symwell:
> I'm not really understanding the error. Does `npx @appland/appmap-agent-js init .` work when not launched from the installer? The fact that this is a yarn workspace shouldn't make a difference.

`npx @appland/appmap-agent-js init .` works when not launched from the installer.

```
~/src/appmap-js.two$ npx @appland/appmap-agent-js init 
{
  "filename": "appmap.yml",
  "configuration": {
    "contents": "name: appmap-js.two\npackages:\n  - path: packages/cli/built\n  - path: packages/client/dist/src\n  - path: packages/models/dist\n  - path: packages/cli\n  - path: packages/client\n  - path: packages/components\n  - path: packages/diagrams\n  - path: packages/models\n  - path: packages/openapi\n  - path: packages/scanner\n  - path: packages/sequence-diagram\n  - path: packages/telemetry\n  - path: packages/validate/bin\n  - path: packages/validate/lib\n  - path: packages/validate/schema\n  - path: packages/validate/src\n  - path: packages/validate/test\n"
  }
}

~/tmp/js-project2/packages/dir_one$ npx @appland/appmap-agent-js init .
{
  "filename": "appmap.yml",
  "configuration": {
    "contents": "name: .\npackages: []\n"
  }
}
```

packages/cli/src/cmds/agentInstaller/projectConfiguration.ts
```diff
@@ -81,12 +84,42 @@ async function resolveSelectedInstallers(
...
+        if ('workspaces' in packageJson) {
+          const workspaces = packageJson['workspaces'];
+          // process as glob because it could be something like
+          // workspaces: [ 'packages/*' ]
+          workspaces.forEach((workspace) => {
+            const matches = glob.sync(workspace, {
+              matchBase: true,
+              cwd: installer.path,
+              ignore: ['**/node_modules/**', '**/.git/**'],
+            });
```
dustinbyrne, line 113: Set `strict` to false here to avoid permission errors thrown by `glob`. I'd normally say this is a case we don't need to worry about, but as we've learned, edge cases are not really edge cases once this is running out in the wild.

packages/cli/src/cmds/agentInstaller/projectConfiguration.ts
symwell: I went ahead and changed this.  Initially I tried `const data = await fs.readFile(packageJsonFilename, 'utf-8');` but the resulting `data` couldn't get processed by `JSON.parse` until I switched the loop from `.forEach` to `for (const ...) {`.

But then I used synchronous `lstatSync` because async `lstat` didn't seem to return `isDirectory()`.

packages/cli/src/cmds/agentInstaller/projectConfiguration.ts
symwell: Added.

dividedmind:
What if the user wants to install appmap in the root workspace? It sounds reasonable to want to do that.

dividedmind:
Further testing shows that with this branch, you can still install into the root workspace by answering 'no' to the `This directory contains sub-projects. Would you like to choose sub-projects for installation?` prompt (presumably with the same yarn1-related bug as in the OP). However, I've ran into another issue that again looks related to yarn version mismatch:

<pre>$ yarn start install -d ~/projects/ws-test/
<font color="#26A269">?</font> <b>This directory contains sub-projects. Would you like to choose sub-projects for installation?</b> <font color="#2AA1B3">Yes</font>
<font color="#26A269">?</font> <b>Select the projects to install AppMap to.</b> <font color="#2AA1B3">pkg1, pkg2</font>
Installing AppMap agent for <font color="#12488B">pkg1</font>...
<font color="#26A269">?</font> <b>AppMap is about to be installed. Confirm the details below.</b>
<b>  </b><font color="#12488B"><b>!!! Telemetry disabled !!!</b></font><b>: true</b>
<b>  </b><font color="#12488B"><b>Project type</b></font><b>: yarn</b>
<b>  </b><font color="#12488B"><b>Project directory</b></font><b>: /home/divide/projects/ws-test/packages/pkg1</b>
<b>  </b><font color="#12488B"><b>Git remote</b></font><b>: [no remote]</b>

<b>  Is this correct?</b> <font color="#2AA1B3">Yes</font>
<font color="#2AA1B3">⠇</font> Installing AppMap...Error: An error occurred while running the command: <font color="#C01C28">yarn install --immutable</font>
The command exited with code <font color="#C01C28">1</font>
<font color="#C01C28">&gt;  </font>➤ YN0000: ┌ Resolution step
<font color="#C01C28">&gt;  </font>➤ YN0000: └ Completed
<font color="#C01C28">&gt;  </font>
<font color="#C01C28">&gt;  </font>➤ YN0000: ┌ Post-resolution validation
<font color="#C01C28">&gt;  </font>➤ YN0000: │ @@ -1,9 +1,9 @@
<font color="#C01C28">&gt;  </font>➤ YN0000: │  # This file is generated by running &quot;yarn install&quot; inside your project.
<font color="#C01C28">&gt;  </font>➤ YN0000: │  # Manual changes might be lost - proceed with caution!
<font color="#C01C28">&gt;  </font>➤ YN0000: │  
<font color="#C01C28">&gt;  </font>➤ YN0000: │  __metadata:
<font color="#C01C28">&gt;  </font>➤ YN0028: │ -  version: 4
<font color="#C01C28">&gt;  </font>➤ YN0028: │ +  version: 6
<font color="#C01C28">&gt;  </font>➤ YN0000: │  
<font color="#C01C28">&gt;  </font>➤ YN0000: │  &quot;pkg1@workspace:packages/pkg1&quot;:
<font color="#C01C28">&gt;  </font>➤ YN0000: │    version: 0.0.0-use.local
<font color="#C01C28">&gt;  </font>➤ YN0000: │    resolution: &quot;pkg1@workspace:packages/pkg1&quot;
<font color="#C01C28">&gt;  </font>➤ YN0000: │ 
<font color="#C01C28">&gt;  </font>➤ YN0028: │ The lockfile would have been modified by this install, which is explicitly forbidden.
<font color="#C01C28">&gt;  </font>➤ YN0000: └ Completed
<font color="#C01C28">&gt;  </font>➤ YN0000: Failed with errors in 0s 25ms
<font color="#C01C28">✖</font> Installing AppMap...

<font color="#C01C28">!</font> Error in project configuration:
Error: An error occurred while running the command: <font color="#C01C28">yarn install --immutable</font>
The command exited with code <font color="#C01C28">1</font>
<font color="#C01C28">&gt;  </font>➤ YN0000: ┌ Resolution step
<font color="#C01C28">&gt;  </font>➤ YN0000: └ Completed
<font color="#C01C28">&gt;  </font>
<font color="#C01C28">&gt;  </font>➤ YN0000: ┌ Post-resolution validation
<font color="#C01C28">&gt;  </font>➤ YN0000: │ @@ -1,9 +1,9 @@
<font color="#C01C28">&gt;  </font>➤ YN0000: │  # This file is generated by running &quot;yarn install&quot; inside your project.
<font color="#C01C28">&gt;  </font>➤ YN0000: │  # Manual changes might be lost - proceed with caution!
<font color="#C01C28">&gt;  </font>➤ YN0000: │  
<font color="#C01C28">&gt;  </font>➤ YN0000: │  __metadata:
<font color="#C01C28">&gt;  </font>➤ YN0028: │ -  version: 4
<font color="#C01C28">&gt;  </font>➤ YN0028: │ +  version: 6
<font color="#C01C28">&gt;  </font>➤ YN0000: │  
<font color="#C01C28">&gt;  </font>➤ YN0000: │  &quot;pkg1@workspace:packages/pkg1&quot;:
<font color="#C01C28">&gt;  </font>➤ YN0000: │    version: 0.0.0-use.local
<font color="#C01C28">&gt;  </font>➤ YN0000: │    resolution: &quot;pkg1@workspace:packages/pkg1&quot;
<font color="#C01C28">&gt;  </font>➤ YN0000: │ 
<font color="#C01C28">&gt;  </font>➤ YN0028: │ The lockfile would have been modified by this install, which is explicitly forbidden.
<font color="#C01C28">&gt;  </font>➤ YN0000: └ Completed
<font color="#C01C28">&gt;  </font>➤ YN0000: Failed with errors in 0s 25ms
</pre>

(BTW, do we need to repeat that log?)

symwell:
> What if the user wants to install appmap in the root workspace? It sounds reasonable to want to do that.

When `package.json` contains something like `workspace: []` the installer still gives the error `Running this command will add the dependency to the workspace root...`.  Maybe the installer should check for a blank workspace array and tell the user to remove `workspace: []` from `package.json`.

dividedmind:
> > What if the user wants to install appmap in the root workspace? It sounds reasonable to want to do that.
> 
> When `package.json` contains something like `workspace: []` the installer still gives the error `Running this command will add the dependency to the workspace root...`. Maybe the installer should check for a blank workspace array and tell the user to remove `workspace: []` from `package.json`.

This is caused by yarn v1 needing `-W` flag to install in the root workspace. The message seems pretty clear to me.

packages/cli/src/cmds/agentInstaller/projectConfiguration.ts
```diff
@@ -81,12 +86,47 @@ async function resolveSelectedInstallers(
...
     throw new Error('Invalid selection');
   }
 }
 
+export async function getYarnPackages(availableInstallers: AgentInstaller[]): Promise<string[]> {
+  let yarnPackages: string[] = [];
+  for (const installer of availableInstallers) {
+    if (installer.name === 'yarn') {
+      const packageJsonFilename = join(installer.path, 'package.json');
+      if (await exists(packageJsonFilename)) {
```
dividedmind, line 100: I think the recommended idiom is to try to read and rescue the error instead; the file can still disappear after `exists` so it doesn't do much good to even bother calling it.

packages/cli/src/cmds/agentInstaller/projectConfiguration.ts
```diff
@@ -81,12 +86,47 @@ async function resolveSelectedInstallers(
...
+              strict: false,
+            });
+            for (const yarnPackage of matches) {
+              // packages must be directories; don't break on leftover files
+              const isDir = lstatSync(join(installer.path, yarnPackage)).isDirectory();
+              if (isDir) {
+                yarnPackages.push(yarnPackage);
+              }
+            }
+          }
```
dividedmind, line 121: I don't think this is a good method of doing this. Not all directories matching the globs will be workspaces. You should instead call `yarn workspaces list --json` and parse the output — as a bonus, you'll get the root workspace in the list and will be able to handle it the same way as the others.

dividedmind (changes requested):
While this kind of works as expected (although it's not necessarily evident that answering 'no' will install in the root), I think it's safer to call out to `yarn workspaces list --json` to get a canonical list of workspaces (you can have directories matching workspaces globs that don't contain package.json and thus aren't workspaces); this list will include the root workspace which thus it will be possible to handle it in the same way as the others.

Other than that, this patch doesn't address the original problem you ran into, which is not providing the -W flag to yarn v1 when installing into the root workspace. I've also uncovered a problem with yarn version selection (I'm not sure if this will actually be a problem or only happens when running from the source tree with `yarn start`). I think perhaps those two should be spun off into separate issues.

packages/cli/src/cmds/agentInstaller/projectConfiguration.ts
symwell: `exists` is used in 14 other locations in the code.

packages/cli/src/cmds/agentInstaller/projectConfiguration.ts
symwell: With yarn 1.22.19 the workspaces command fails...

```
$ yarn workspaces list --json
yarn workspaces v1.22.19
error Invalid subcommand. Try "info, run"
info Visit https://yarnpkg.com/en/docs/cli/workspaces for documentation about this command.

$ yarn run workspaces list --json
yarn run v1.22.19
error Command "workspaces" not found.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

...even though it's present in the list of available commands
```
$ yarn help | grep -i workspace
    --focus                             Focus on a single workspace by installing remote copies of its sibling workspaces.
    - workspace
    - workspaces
```

Installed newer version of yarn with
```
npm install -g yarn@2.4.3
```

With yarn 2.4.3
```
$ yarn workspaces list --json
{"location":".","name":"yarn_with_subprojects"}
{"location":"packages/subproject_one","name":"subproject_one"}
{"location":"packages/subproject_two","name":"subproject_two"}
```

--

Maybe we need to differentiate between using yarn 1 and yarn >=  2?
- yarn 1 can use the existing logic
- yarn 2 can use `yarn workspaces list --json`

symwell: I implemented detecting yarn 1 and yarn >=2.  Now changing the testcases.

packages/cli/src/cmds/agentInstaller/projectConfiguration.ts
symwell: Changed `exists` to `try/catch`.

symwell:
FWIW now I get a `Lipzip error: Malloc failure`
```
~/src/appmap-js/packages/cli$ npm start install -- -d /home/test/tmp/js-project2

> @appland/appmap@3.60.1 start
> ts-node src/cli.ts install-agent -d /home/test/tmp/js-project2

? This directory contains sub-projects. Would you like to choose sub-projects for installation? Yes
? Select the projects to install AppMap to. dir_one
Installing AppMap agent for dir_one...
  Project type: yarn
  Project directory: /home/test/tmp/js-project2/packages/dir_one
  Git remote: [git remote failed, fatal: not a git repository (or any of the parent directories): .git]

  Is this correct? Yes
⠙ Installing AppMap...Error: An error occurred while running the command: yarn install --immutable
The command exited with code 1
>  ➤ YN0000: ┌ Resolution step
>  ➤ YN0001: │ Libzip Error: dir_one@file:/home/test/tmp/js-project2/packages/dir_one::locator=dir_one%40workspace%3A.: Malloc failure
>      at f.makeLibzipError (/home/test/.nvm/versions/node/v18.8.0/lib/node_modules/yarn/bin/yarn.js:2:476882)
>      at f.getBufferAndClose (/home/test/.nvm/versions/node/v18.8.0/lib/node_modules/yarn/bin/yarn.js:2:477550)
>      at d (/home/test/.nvm/versions/node/v18.8.0/lib/node_modules/yarn/bin/yarn.js:2:137576)
>      at async E.resolvers.getCandidates (/home/test/.nvm/versions/node/v18.8.0/lib/node_modules/yarn/bin/yarn.js:2:141503)
>      at async n.getCandidates (/home/test/.nvm/versions/node/v18.8.0/lib/node_modules/yarn/bin/yarn.js:2:344518)
>      at async n.getCandidates (/home/test/.nvm/versions/node/v18.8.0/lib/node_modules/yarn/bin/yarn.js:2:344518)
>      at async /home/test/.nvm/versions/node/v18.8.0/lib/node_modules/yarn/bin/yarn.js:2:366174
>      at async Module.w (/home/test/.nvm/versions/node/v18.8.0/lib/node_modules/yarn/bin/yarn.js:2:425647)
>      at async C (/home/test/.nvm/versions/node/v18.8.0/lib/node_modules/yarn/bin/yarn.js:2:366137)
>      at async Promise.all (index 1)
>  ➤ YN0000: └ Completed in 48s 199ms
>  ➤ YN0000: Failed with errors in 48s 202ms
>  ➤ YN0061: @npmcli/move-file@npm:1.1.2 is deprecated: This functionality has been moved to @npmcli/fs
```

packages/cli/src/cmds/agentInstaller/projectConfiguration.ts
symwell: The tests pass locally but not on CI.

packages/cli/tests/unit/agentInstall/installCommand.spec.ts
```diff
@@ -749,27 +751,50 @@ appmap_dir: tmp/appmap
...
+      if (installerName === 'yarn') {
+        sinon
+          .stub(commandRunner, 'run')
+          .onCall(callIdx++)
+          .callsFake(nodeVersion)
+          .onCall(callIdx++)
+          .callsFake(getYarnWorkspaces)
+          .onCall(callIdx++)
+          // why must installAgent be here and twice for the yarn test to pass?
+          .callsFake(installAgent)
```
symwell, line 773: Needing to do this seems strange to me.

packages/cli/src/cmds/agentInstaller/projectConfiguration.ts
symwell: The tests now pass on CI.

symwell:
> not providing the -W flag to yarn v1 when installing into the root workspace.

Added -W flag for yarn 1.

packages/cli/tests/unit/agentInstall/installCommand.spec.ts
lachrist: Yes that *is* strange. It would be nice to resolve this before merging.

lachrist (approved):
First, `fix: Improve testcase to contain (incorrectly) files among the packages` is marked as a `fix:` but it only change test files. Wouldn't this create a strange changelog? Shouldn't this be a marked as`test:`?

Second, the title of the two actual fix commits do not reflect their contents:
  - `fix: Detect yarn packages as sub-projects` implements `getYarnSubprojectsVersionAboveOne`
  - `fix: Detect subprojects for yarn2` changes `getYarnSubprojectsVersionOne`

I would either squash these commits together or disentangle them -- ie: the first focuses on `yarn@1` and the second focuses on `yarn@>1`.

Otherwise, the code looks legitimate and is tested.

I still approved it because I'm not sure how important it is the maintain a clean history for the CLI.

symwell:
> ⠇ Installing AppMap...Error: An error occurred while running the command: yarn install --immutable

This error is because `yarn install` was called with `--immutable` to check if installing with yarn in this repository would work.  The solution (I think) is for the user to run `yarn install` on their own before installing AppMap.

packages/cli/tests/unit/agentInstall/installCommand.spec.ts
symwell: Also strange:
- the test [`installAgent`](https://github.com/getappmap/appmap-js/blob/main/packages/cli/tests/unit/agentInstall/installCommand.spec.ts#L894) function and [`javaScriptAgentInstaller.installAgent`](https://github.com/getappmap/appmap-js/blob/main/packages/cli/src/cmds/agentInstaller/javaScriptAgentInstaller.ts#L107) both get called.
- changing the command issued by the test `installAgent` function (to something like `echo hi`) leads to the testcase not failing.

functions_called_twice

packages/cli/src/cmds/agentInstaller/projectConfiguration.ts
```diff
@@ -98,37 +195,50 @@ async function resolveSelectedInstallers(
...
+
+          return {
+            name: ent.name,
+            path: subProjectPath,
+            availableInstallers: installers,
+          };
+        })
+      )
+    ).filter(Boolean) as ProjectConfiguration[];
+  }
```
dividedmind, line 241: Instead of bolting more functionality into this already long function, I'd suggest keeping it as is and using another short function to tie it and `getYarnSubprojects()` together. What do you think?

packages/cli/tests/unit/agentInstall/installCommand.spec.ts
dividedmind: Maybe it could be made cleaner with jest instead of sinon. Jest mocking is pretty solid, plus you can chain `.mockImplementationOnce` to build a sequence without interspersing it with `onCall`.

packages/cli/tests/unit/fixtures/javascript/yarn_with_subprojects/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "yarn_with_subprojects",
+  "packageManager": "yarn@3.2.1",
+  "private": true,
+  "workspaces": [ "packages/*" ]
+}
```
dividedmind, line 6: I wonder if you even need this file for if you're mocking the yarn outputs anyway?

dividedmind (changes requested):
It'd be good if we could get away with not trying to reproduce yarn1 workspace resolution and instead use `yarn workspaces info`.

packages/cli/src/cmds/agentInstaller/projectConfiguration.ts
```diff
@@ -81,12 +87,103 @@ async function resolveSelectedInstallers(
...
+  availableInstallers: AgentInstaller[]
+): Promise<ProjectConfiguration[]> {
+  let subprojects: ProjectConfiguration[] = [];
+
+  for (const installer of availableInstallers) {
+    if (installer.name === 'yarn') {
+      if (await (installer as YarnInstaller).isYarnVersionOne()) {
+        await getYarnSubprojectsVersionOne(dir, installer as YarnInstaller, subprojects);
+      } else {
+        await getYarnSubprojectsVersionAboveOne(dir, installer as YarnInstaller, subprojects);
```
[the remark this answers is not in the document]
symwell: Declared a local variable that coerces the type and used this variable instead of repeated coercion. 
```
      const yarnInstaller = installer as YarnInstaller;
```

packages/cli/src/cmds/agentInstaller/projectConfiguration.ts
```diff
@@ -81,12 +87,103 @@ async function resolveSelectedInstallers(
...
+          const isDir = lstatSync(join(installer.path, yarnPackage)).isDirectory();
+          if (isDir) {
+            const subProjectPath = join(dir, yarnPackage);
+            const subProjectName = basename(yarnPackage);
+            subprojects.push({
+              name: subProjectName,
+              path: subProjectPath,
+              // must set the new path in the installer
+              availableInstallers: [new YarnInstaller(subProjectPath)],
+            } as ProjectConfiguration);
```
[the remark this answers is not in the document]
symwell: Fixed.

packages/cli/src/cmds/agentInstaller/projectConfiguration.ts
```diff
@@ -81,12 +87,103 @@ async function resolveSelectedInstallers(
...
+    if (line !== '') {
+      const yarnWorkspace = JSON.parse(line);
+      const subProjectPath = join(dir, yarnWorkspace['location']);
+      const subProjectName = basename(yarnWorkspace['name']);
+      subprojects.push({
+        name: subProjectName,
+        path: subProjectPath,
+        // must set the new path in the installer
+        availableInstallers: [new YarnInstaller(subProjectPath)],
+      } as ProjectConfiguration);
```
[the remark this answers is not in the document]
symwell: Fixed.

packages/cli/src/cmds/agentInstaller/projectConfiguration.ts
```diff
@@ -81,12 +87,103 @@ async function resolveSelectedInstallers(
...
+  // Sanity check. Verify everything resolved properly. This branch should never occur.
   if (projects.some(({ selectedInstaller }) => !selectedInstaller)) {
     throw new Error('Invalid selection');
   }
 }
 
+// With yarn 1 we get the following error.  Try to get the packages manually.
+// $ yarn run workspaces list --json
+// yarn run v1.22.19
+// error Command "workspaces" not found.
```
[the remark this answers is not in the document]
symwell: `yarn workspaces info --json` didn't work when [I tried](https://github.com/getappmap/appmap-js/pull/936#discussion_r1071339077).  In some projects this error doesn't occur, and these projects look identical in structure to the ones that give the error.

Changed code to now handle this error, and also to use `yarn workspaces info --json`.

packages/cli/tests/unit/fixtures/javascript/yarn_with_subprojects/package.json
symwell: I used it to run the installer manually.
```
npm start install -- -d /home/test/src/appmap-js/packages/cli/tests/unit/fixtures/javascript/yarn_with_subprojects
```
It didn't help much because I got the [error](https://user-images.githubusercontent.com/111290954/211630706-b129b9f5-8019-40c9-8603-eb36defd303a.png) in the OP.   Maybe it'd be helpful to other devs.

packages/cli/src/cmds/agentInstaller/projectConfiguration.ts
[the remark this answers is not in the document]
dividedmind: It seems in that comment you tried `yarn workspaces list`, not `info`. `list` is v2 command, `info` should be used in v1.

packages/cli/src/cmds/agentInstaller/projectConfiguration.ts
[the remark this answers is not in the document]
dividedmind: A type check would have been cleaner, but ok :)

packages/cli/src/cmds/agentInstaller/projectConfiguration.ts
```diff
@@ -81,12 +87,109 @@ async function resolveSelectedInstallers(
...
+  // Remove any text before and after JSON
+  let lines = output.stdout.split('\n');
+  let outputClean = '';
+  for (const line of lines) {
+    if (line.startsWith('error')) return;
+    if (!(line.startsWith('yarn workspaces v') || line.startsWith('Done in '))) outputClean += line;
+  }
+
+  Object.entries(JSON.parse(outputClean)).forEach(([workspaceName, workspaceData]) => {
+    const data = workspaceData as Yarn1Workspace; // is there a better way to use this type?
```
dividedmind, line 135: A better way would be to verify it's the expected type and let TypeScript know about it. (Since the type is very small and only used here, it might not be necessary to even declare it.) Here's a suggestion:
```typescript
function yarn1Workspaces(info: unknown): [string, string][] {
  assert(info && typeof info === 'object');
  return Object.entries(info).map(([name, data]: [string, unknown]) => {
    assert(data && typeof data === 'object');
    assert('location' in data);
    assert(typeof data.location === 'string');
    return [name, data.location];
  });
}

for (const [name, location] of yarn1Workspaces(JSON.parse(outputClean))) {
  const path = join(dir, location),
  subprojects.push({
    name: basename(name), path
    availableInstallers: [new YarnInstaller(path)]
  });
}
```
Note how `unknown`s force the type checker to make sure you're verifying the object types before using them.

dividedmind (approved):
Looks good, but please make sure to squash the commits before merging. "Fixed: Address feedback from code review" changelog entry is bound to puzzle some users. Thanks! :)

packages/cli/src/cmds/agentInstaller/javaScriptAgentInstaller.ts
```diff
@@ -117,4 +118,22 @@ export class YarnInstaller extends JavaScriptInstaller {
...
+      if (!version) {
+        throw new UserConfigError('Could not detect yarn version');
+      }
+
+      isVersionOne = !semver.satisfies(version, '>= 2.0.0');
+    } catch (err) {
+      throw new UserConfigError(err as string);
+    }
+
+    return new Promise(resolve => resolve(isVersionOne));
```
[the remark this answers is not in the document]
symwell: Changed.

packages/cli/src/cmds/agentInstaller/projectConfiguration.ts
```diff
@@ -1,10 +1,16 @@
 import chalk from 'chalk';
-import { promises as fs, constants as fsConstants } from 'fs';
+import { promises as fs, constants as fsConstants, lstatSync } from 'fs';
```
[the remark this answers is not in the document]
symwell: Removed.

packages/cli/src/cmds/agentInstaller/projectConfiguration.ts
symwell: Changed.

symwell:
> Maybe it could be made cleaner with jest instead of sinon. Jest mocking is pretty solid, plus you can chain .mockImplementationOnce to build a sequence without interspersing it with onCall.

I tried using jest with `.mockImplementationOnce ` instead of sinon.  Same result.

packages/cli/tests/unit/agentInstall/installCommand.spec.ts
symwell: > Maybe it could be made cleaner with jest instead of sinon. Jest mocking is pretty solid, plus you can chain .mockImplementationOnce to build a sequence without interspersing it with onCall.

I tried using jest with .mockImplementationOnce  instead of sinon. Same result.

packages/cli/src/cmds/agentInstaller/projectConfiguration.ts
symwell: Having an extra short function that calls this one could decrease readability because it adds one more level of indirection the developer must follow to trace this logic.

PR #959, opened 2023-01-19
fix: do not run components tests on windows

ahtrotta:


packages/components/.storybook/run.js
```diff
@@ -46,8 +46,13 @@ async function runStorybook() {
   });
 }
 
+function isWindows() {
+  return !!process.platform.match(/win/);
```
dustinbyrne, line 50: This will match `darwin`. Use `/^win/`.

ahtrotta: Oh, good catch!

PR #964, opened 2023-01-20
fix: OpenAPI supports python builtins.NoneType

kgilpin:
Support Python builtins.NoneType. It's reported as `object` until a better example comes along. That's how we handle NilClass in Ruby.


apotterri (approved):
As soon as the build is green, this is good to go. Thanks for fixing this.

PR #967, opened 2023-01-20
feat: OpenAPI max file size

kgilpin:
## Demo

https://asciinema.org/a/v69jrmpHLOasAH27JagH1wBeR

### DIFF
--- a/packages/cli/src/cmds/openapi.ts
+++ b/packages/cli/src/cmds/openapi.ts
@@ -1,7 +1,7 @@
 import { join } from 'path';
 
-import { existsSync, promises as fsp } from 'fs';
-import { readFile } from 'fs/promises';
+import { existsSync, promises as fsp, Stats } from 'fs';
+import { readFile, stat } from 'fs/promises';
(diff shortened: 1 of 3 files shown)


packages/cli/src/cmds/openapi.ts
```diff
@@ -111,6 +145,10 @@ export default {
     args.option('appmap-dir', {
       describe: 'directory to recursively inspect for AppMaps',
     });
+    args.option('max-size', {
+      describe: 'maximum AppMap size that will be processed, in filesystem-reported MB',
```
apotterri, line 149: "filesystem-reported" seems like a slightly odd way to describe this: "filesystem-reported" as opposed to what? Maybe this should just be "in MB"?

apotterri (approved):
One minor wording change to consider, otherwise good-to-go.

kgilpin:
Yeah I would be inclined to change it if builds didn’t take an hour :-)

kgilpin:
It means that a MB is 1024*1024, not 1e6

PR #978, opened 2023-01-23
fix: Fix basePath to not return '' when there's ( and )

symwell:
Before this change, `basePath` produced paths that were blank like
```
requestPath (/locale/{locale})/api/users/{id} produced 
requestPath (/locale/{locale})/api/readinglist produced 
```

After this change, `basePath` produces paths like
```
requestPath (/locale/{locale})/api/users/{id} produced (/locale/{locale})/api/users/{id}
requestPath (/locale/{locale})/api/readinglist produced (/locale/{locale})/api/readinglist
```
With this change the [`openapi.yml`](https://github.com/forem/forem/compare/main...symwell:forem:sw/use_appmap_to_create_openapi_doc?expand=1#diff-5016bdc926b6c0f649f363739d09159a1db664aae19a08fa6b38b03406f040dc) file for Forem was generated with many paths instead of a path named `''`. `openapi.yml` contains all 42 API calls that were found in the AppMap files.

I didn't understand why `basePath` was attempting to `.split('(')`.  Maybe this logic was obsolete; maybe it tried to remove format specifiers, like convert `'/org/:org_id(.:format)'` into `'/org/{org_id}` , which are removed by [`appmap-ruby`](https://github.com/getappmap/appmap-ruby/blob/master/lib/appmap/util.rb#L149) instead.

### DIFF
--- a/packages/openapi/src/model.ts
+++ b/packages/openapi/src/model.ts
@@ -27,7 +27,6 @@ export default class Model {
 
   static basePath(path: string): string {
     const pathTokens = path
-      .split('(')[0]
       .split('/')
       .map((entry) => {
         if (entry.match(/^:(.*)/)) {
--- a/packages/cli/src/cmds/openapi.ts
+++ b/packages/cli/src/cmds/openapi.ts
@@ -76,7 +76,7 @@ class OpenAPICommand {
       });
     } catch (e) {
       // Re-throwing this error crashes the whole process.
-      // So if these is a malformed AppMap, indicate it here but don't blow everything up.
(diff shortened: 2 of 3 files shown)


packages/openapi/src/model.ts
```diff
@@ -27,7 +27,6 @@ export default class Model {
 
   static basePath(path: string): string {
     const pathTokens = path
-      .split('(')[0]
```
kgilpin, line 30: Right, this is probably there to clean up a Ruby behavior.

kgilpin (approved):

PR #979, opened 2023-01-24
fix: Prevent data from being dropped during serialization

dustinbyrne:
This fixes two cases. First, event properties would not always be serialized when the value was an object. Only in cases where the object contained an Array or a Set would it have been serialized. Second, `http_server_response.status_code` was being re-written to `http_server_response.status`. This likely wasn't resulting in any bugs, but would change the contents of an AppMap after serialization occurred. `status` and `status_code` can now be safely interchanged for backwards compatibility reasons in case anything downstream depends on one or the other.

### DIFF
--- a/packages/cli/src/fingerprint/fingerprinter.ts
+++ b/packages/cli/src/fingerprint/fingerprinter.ts
@@ -16,6 +16,10 @@ const renameFile = promisify(gracefulFs.rename);
 /**
  * CHANGELOG
  *
+ * # 1.1.4
+ *
+ * Add missing status_code to normalized AppMaps.
+ *
  * # 1.1.3
  *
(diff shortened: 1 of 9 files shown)


dustinbyrne:
Once this is released, `@appland/appmap` will need to be released with a bump in its `@appland/models` version.

kgilpin (approved):

kgilpin:
Does the release bot not know to do that?

dustinbyrne:
I ended up deleting the comment because I'm not actually sure what the version constraints are off hand. It may end up being pulled in automatically. But no, `semantic-release` won't publish downstream dependencies automatically.

dustinbyrne:
Looks like this fixed an issue in the `Canonicalize` class for `update`. The tests needed to be updated to reflect the change.

PR #981, opened 2023-01-25
Hide Share button by default

### LINKED ISSUE #980: Remove Share button

ahtrotta:
Fixes #980

### DIFF
--- a/packages/components/src/pages/VsCodeExtension.vue
+++ b/packages/components/src/pages/VsCodeExtension.vue
@@ -94,7 +94,7 @@
         </template>
         <template v-slot:controls>
           <v-popper
-            v-if="appMapUploadable"
+            v-if="shareEnabled"
             class="hover-text-popper"
             text="Create a link to this AppMap."
             placement="left"
@@ -401,14 +401,8 @@ export default {
       currentTraceFilterIndex: 0,
       showShareModal: false,
       shareURL: undefined,
+      shareEnabled: false,
     };
-  },
-
-  props: {
-    appMapUploadable: {
-      type: Boolean,
-      default: false,
-    },
   },
 
   watch: {
@@ -774,6 +768,10 @@ export default {
 
       this.isLoading = false;
       this.$store.commit(SET_FILTERED_MAP, this.filteredAppMap);
+    },
+
+    setShareEnabled(shareEnabled) {
+      this.shareEnabled = shareEnabled;
     },
 
     showInstructions() {
--- a/packages/components/src/stories/VsCodeExtension.stories.js
+++ b/packages/components/src/stories/VsCodeExtension.stories.js
@@ -36,9 +36,10 @@ export default {
       },
       defaultValue: 'default',
     },
-  },
-  args: {
-    appMapUploadable: true,
+    shareEnabled: {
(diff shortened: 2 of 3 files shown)


packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -401,16 +401,10 @@ export default {
...
+      shareEnabled: false,
     };
   },
 
-  props: {
-    appMapUploadable: {
-      type: Boolean,
-      default: false,
-    },
-  },
```
dustinbyrne, line 412: `appMapUploadable` was already defaulted to `false` - what's the behavioral change this PR makes?

packages/components/src/stories/VsCodeExtension.stories.js
```diff
@@ -36,9 +36,10 @@ export default {
       },
       defaultValue: 'default',
     },
-  },
-  args: {
-    appMapUploadable: true,
+    shareEnabled: {
+      control: 'boolean',
+      defaultValue: 'true',
```
dustinbyrne, line 41: This declares the control as a boolean but sets the default value to a string. Is this right?

packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -401,16 +401,10 @@ export default {
       currentTraceFilterIndex: 0,
       showShareModal: false,
       shareURL: undefined,
+      shareEnabled: false,
```
dustinbyrne, line 404: What's the benefit of moving this to a data property?

ahtrotta: I moved it to a data property because I wanted to use a feature flag. In VS Code, we use the `ExtensionSettings` class to keep track of feature flags, and it seemed easier to import that class into the `appmapEditorProvider.ts` file rather than `appmapView.js` (see VS Code PR [here](https://github.com/getappmap/vscode-appland/pull/608/commits/1fdf70b4943a56cfc9d582e3dc686e47e702c96e)).

dustinbyrne: You're right, importing backend logic into the frontend won't work because the VS Code API isn't available within the context of the browser. Instead we need a channel of communication between the backend and frontend to exchange information like feature flags.

We're doing this to a greater extent in the instructions web view. Here, we take the initial state (which is probably enough in this case) and initialize the root Vue applications data properties:
https://github.com/getappmap/vscode-appland/blob/master/web/src/installGuideView.js#L34

Then the flag is provided via props:
https://github.com/getappmap/vscode-appland/blob/master/web/src/installGuideView.js#L24

The only reason we're using a data property here is because we want the flag to be reactive. By modifying the data property later on, it's propagated to any child elements where it's passed as a prop. We have another message handler to handle the state change explicitly:
https://github.com/getappmap/vscode-appland/blob/master/web/src/installGuideView.js#L109-L114

So that's all to say that I don't think we need this change. We just need to supply the feature flag boolean to the prop from within the extension frontend. After all, there's minimal difference between passing a value via a prop or via method call on creation. We're just setting a flag, so there shouldn't be any technical limitation we're hitting.

PR #985, opened 2023-01-27
fix: throttle the globber's stat() calls

### LINKED ISSUE #969: Scan and index: High CPU usage when idle
Image

apotterri:
Have the globber use a bit less CPU: when it scans to check for updates, sleep for 100ms between calls to stat().

Fixes #969 .


kgilpin:
Can you explain the change a little more? I am not quite getting the intent from the code alone. How it used to work, how it works now, why you chose to do it this way.

apotterri:
Sure. 

Prior to these changes, we used `glob` to scan for `*.appmap.json` files. As each was found one, the code would call `fs.promises.stat` on it to get its `mtime`. When `stat` returned, it would emit an `'add'` or `'change'` event as appropriate. Once the scan finished, it would look through the files it previously found, and emit an `'unlink'` event for each that had been removed. Finally,  it would emit an `'end'` event, then use `setTimeout` to schedule another scan 1s in the future.

Now, it uses `glob` to scan for all `*.appmap.json` files. Once the scan finishes, it iterates through the files it previously found, and emits an `'unkink`' for each that has been removed. Then, it iterates through the results of the scan. It calls `stat` on the first one, and when that `stat` finishes, it schedules the next `stat` for 100ms in the future. When the final `stat` finishes, it emits an `'end'` event, and schedules the next scan for 1s in the future.

With the previous approach, there was no way to gate the calls to `stat`. The code iterated through them as fast `glob` found them. When there were many files, this would lead to high CPU utilization. Introducing a slight delay between `stat` calls gives the CPU a chance to do other things, leading to a better user experience.

kgilpin:
The purpose of the glob scan is to pick up changes that the watcher might miss. So limiting it’s throughout to 10 files per second seems too slow. It’s not uncommon for a project to have at least hundreds of AppMaps if not more. If it takes several minutes for changes to be reflected in the index, then it doesn’t seem like the user will feel like the updates are happening responsively enough. 

How did you arrive at the 100ms delay between stat calls? Could this be reduced substantially and still get the effect we need? Even 1ms should be a pretty long time when we are taking about a system call as simple as stat; no?

How about making this into a CLI parameter so that we can tune it from the client (eg the code editor) without a recompile?

apotterri:
I was under the impression that the changes that the watcher might miss were edge cases, but I don't know the specifics. 

This approach, including the 100ms delay, was basically what @dividedmind suggested. I think it would be helpful to hear his perspective on these changes.

Having said that, I'll play with the delay and see how it effects CPU utilization. I'll also look into making it configuable.

kgilpin:
Ok, let me know when you are satisfied with it.

apotterri:
**This needs a rebase before it gets merged.**

I tried a 10ms delay instead of a 100ms, in a repo with ~2600 AppMaps. CPU utilization was a bit bursty, but generally seemed acceptable -- it was never enough to be noticeable while using the machine, and didn't show spikes on Activity Monitor. I added the hidden option `--watch-delay` to configure it, just in case.

I haven't gotten any input from @dividedmind yet. LMK if there's anything else you'd like to see

packages/cli/src/cli.ts
```diff
@@ -361,6 +363,12 @@ yargs(process.argv.slice(2))
       args.option('appmap-dir', {
         describe: 'directory to recursively inspect for AppMaps',
       });
+      args.options('watch-delay', {
+        type: 'number',
+        default: 10,
+        describe: 'delay between stat calls when watching, in milliseconds',
+      });
```
kgilpin, line 370: Seems like it would be called stat-delay then?

apotterri:
I originally named it 'watch-stat-delay', because it's only relevant to
watching. I thought that was kind of verbose, but maybe it's not?

kgilpin:
It’s ok to be verbose since it also obscure :-)

apotterri:
Ok, I went with `--watch-stat-delay`.

kgilpin (approved):

packages/cli/src/fingerprint/globber.ts
```diff
@@ -84,8 +112,6 @@ class Globber extends EventEmitter {
 interface Globber {
   on(event: 'end', listener: () => void): this;
   on(event: 'add' | 'change' | 'unlink', listener: (path: string) => void): this;
-  emit(event: 'end'): boolean;
-  emit(event: 'add' | 'change' | 'unlink', path: string): boolean;
```
dividedmind, line 88: Why remove this part of the type? It's useful for the consumers.

packages/cli/src/fingerprint/globber.ts
```diff
@@ -48,13 +62,27 @@ class Globber extends EventEmitter {
       if (!files.has(f)) this.remove(f);
     }
 
-    this.emit('end');
+    this.updateStats(found, 0, () => {
+      this.emit('end');
+
+      if (this.running) this.timeout = setTimeout(this.scanNow.bind(this), this.interval);
```
dividedmind, line 68: This timeout should probably be `unref()`-ed to make it behave nicer on termination.

packages/cli/src/fingerprint/index.js
```diff
@@ -1,19 +1,6 @@
...
-  } else {
-    cmd = new FingerprintDirectoryCommand(dir);
-  }
-  return cmd.execute();
-}
 
 module.exports = {
   algorithms,
   canonicalize,
-  fingerprintDirectory,
```
dividedmind, line 18: Ideally this cleanup should go in a separate commit.

dividedmind (approved):
I had a couple of suggestions to improve the readability a bit, but it looks ok, thanks! This is exactly what I had in mind.

packages/cli/src/fingerprint/globber.ts
apotterri: `emit` is called by Globber itself, not consumers/subscribers. What's useful about it code outside this class?

apotterri:
Good suggestions, thanks. I'll incorporate them before I merge.

packages/cli/src/fingerprint/globber.ts
dividedmind: You're right. I think it's still useful for typechecking the calls inside the Globber.

packages/cli/src/fingerprint/globber.ts
dividedmind: Ah, sorry, my mistake, it can't be unref-ed — if you unref it then if the watcher is down the process would exit immediately after one loop. Same goes for the stat timeout.

packages/cli/src/fingerprint/globber.ts
apotterri: Fair enough. I'll put it back.

packages/cli/src/fingerprint/globber.ts
```diff
@@ -48,13 +62,27 @@ class Globber extends EventEmitter {
...
+      cb();
+      return;
+    }
 
-    if (this.running) this.timeout = setTimeout(this.scanNow.bind(this), this.interval);
+    setTimeout(() => {
+      this.statFile(filesToCheck[idx]);
+      this.updateStats(filesToCheck, ++idx, cb);
+    }, this.statDelayMs);
   }
```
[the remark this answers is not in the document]
apotterri: I agree that this is better. I wasn't sure that it was ok to make an EventEmitter callback `async`, but I see that it is. I'll make this change.

dividedmind (changes requested):
Seems like you actually forgot to include the spec in the latest iteration?

apotterri:
Yup, I sure did. Thanks for noticing that!

It looks like my rebase may have caused a test failure, too. I'm looking at that now.

apotterri:
Actually, looks like that build failure happens on `main`, too, and didn't happen when I reran the build.

PR #987, opened 2023-01-31
fix: `CodeObject#location` may be `undefined`

dustinbyrne:
If a CodeObject's location is empty, it'll be omitted upon serialization. This means that if an AppMap was serialized, saved out and loaded again, access to `data.location` would result in an undefined value. Instead of returning `undefined`, return an empty string.

This unblocks https://github.com/getappmap/vscode-appland/pull/606. A test case uses a Python AppMap which contains an empty location, resulting in failures after it's indexed.

### DIFF
--- a/packages/models/src/codeObject.js
+++ b/packages/models/src/codeObject.js
@@ -216,8 +216,8 @@ export default class CodeObject {
   classLocations(paths = new Set()) {
     this.children.forEach((child) => child.classLocations(paths));
 
-    if (this.type === CodeObjectType.FUNCTION) {
-      const tokens = this.data.location.split(':', 2);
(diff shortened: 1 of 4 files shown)


kgilpin:
I don't know, I would expect `location` to be undefined in some normal circumstances. For example, when code comes from an unknown location (e.g. native code, or dynamically compiled `eval`). Can we allow `location == undefined`?

ahtrotta (approved):

dustinbyrne:
Sure, I've changed it so `location` is a `string | undefined`. I've marked it as a breaking change because of course this will introduce `tsc` errors downstream.

PR #989, opened 2023-01-31
Telemetry to better understand sign in

### LINKED ISSUE #988: Telemetry exists to better understand user signin behavior
Events need to be emitted from within the AppMap so that we can better understand user behavior regarding signin.

ahtrotta:
Fixes #988 

This is the first of two PRs to help better understand user behavior. The follow-up PR is [here](https://github.com/getappmap/vscode-appland/pull/609).

This PR causes the frontend to emit events so we can collect telemetry about how users interact with an AppMap:

- When a user clicks on the Filter button
- When a user clicks on the Reset button
- When a user focuses on the search bar
- When a user clicks on a list item in the sidebar
- When a user switches view types by using the tabs

spots

**NOTES**

- There was already an event that was emitted when a user clicked on the Reset button [here](https://github.com/getappmap/appmap-js/blob/main/packages/components/src/pages/VsCodeExtension.vue#L1012), but it was also triggered when clearing the selection in the sidebar, so I made a new event that was only triggered when clicking on the reset button.
- There was already an event that was emitted when a user changed the AppMap view type between `Dependency Map` and `Trace View` [here](https://github.com/getappmap/appmap-js/blob/main/packages/components/src/pages/VsCodeExtension.vue#L809), but it was triggered no matter how the view was changed instead of just when the user clicked on one of the tabs. It was also triggered whenever an AppMap was opened, so I created a new event that was triggered only when the view was changed by clicking the tab.

### DIFF
--- a/packages/components/src/components/DetailsSearch.vue
+++ b/packages/components/src/components/DetailsSearch.vue
@@ -12,6 +12,7 @@
           placeholder="search code objects..."
           ref="searchInput"
           v-model="filter"
+          @focus="searchFocused"
         />
       </div>
     </form>
@@ -212,6 +213,10 @@ export default {
       }
 
       return false;
+    },
+
+    searchFocused() {
+      this.$root.$emit('sidebarSearchFocused');
     },
   },
 };
--- a/packages/components/src/components/DetailsSearch.vue
+++ b/packages/components/src/components/DetailsSearch.vue
@@ -191,6 +191,7 @@ export default {
 
   methods: {
     selectObject(type, object) {
+      this.$root.$emit('selectObjectInSidebar', type);
       if (type === 'labels') {
         this.$store.commit(SELECT_LABEL, object);
       } else {
--- a/packages/components/src/components/TabButton.vue
+++ b/packages/components/src/components/TabButton.vue
@@ -1,5 +1,5 @@
 <template>
-  <button :class="classes">
+  <button :class="classes" @click="clickTab">
     {{ label }}
   </button>
 </template>
@@ -23,6 +23,13 @@ export default {
       required: true,
     },
     isActive: Boolean,
+    reference: String,
+  },
+
+  methods: {
+    clickTab() {
+      this.$root.$emit('clickTab', this.reference);
+    },
   },
 };
 </script>
--- a/packages/components/src/components/Tab.vue
+++ b/packages/components/src/components/Tab.vue
@@ -21,6 +21,7 @@ export default {
       type: String,
       required: true,
     },
+    reference: String,
   },
 
   data() {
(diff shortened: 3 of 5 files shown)


packages/components/src/components/TabButton.vue
```diff
@@ -23,6 +23,13 @@ export default {
       required: true,
     },
     isActive: Boolean,
+    reference: String,
+  },
+
+  methods: {
+    clickTab() {
+      this.$root.$emit('clickTab', this.reference);
```
dustinbyrne, line 31: Somewhere there's a `changeTab` event already being emitted.
https://github.com/getappmap/vscode-appland/blob/master/web/src/appmapView.js#L108-L114

Unless this is to delineate between the user clicking the tab explicitly vs selecting an object which is only represented in a single view?

packages/components/src/components/DetailsSearch.vue
```diff
@@ -191,6 +191,7 @@ export default {
 
   methods: {
     selectObject(type, object) {
+      this.$root.$emit('selectObjectInSidebar', type);
```
dustinbyrne, line 194: FYI we're using `stateChange` elsewhere to pick up on selectedObject changes:
https://github.com/getappmap/vscode-appland/blob/master/web/src/appmapView.js#L87-L106

It won't tell you whether or not it was selected via the sidebar though.

dustinbyrne (approved):

packages/components/src/components/TabButton.vue
ahtrotta: Yeah, sorry, I forgot to finish adding all of my notes to the PR description before I tagged you, but I wanted to delineate between clicking the tab and any other way of changing the view state.

packages/components/src/components/DetailsSearch.vue
ahtrotta: Yeah, similar to the above comment, I wanted a way to see if something was selected via the sidebar specifically. I'm also not sure that the `stateChange` event is working correctly since I can't see any `selected_object` events https://portal.azure.com#@d03f5316-f32a-4633-a0ec-dc62399e533b/blade/Microsoft_OperationsManagementSuite_Workspace/Logs.ReactView/resourceId/%2Fsubscriptions%2Fa4af4729-21ee-4e0a-953c-33cc9db5295c%2FresourceGroups%2Fvscode-extension/source/LogsBlade.AnalyticsShareLinkToQuery/q/H4sIAAAAAAAAA3MsKHAtS80rKeaqUSjPSC1KVfBLzE1VSM7PK0nMzCtWUCpOzUlNLklNic9PygIylKwB70eFojIAAAA%253D/timespan/P7D.

PR #991, opened 2023-02-02
feat: Upload command

### LINKED ISSUE #962: The CLI can upload maps to the server
The old appland CLI tool (written in Golang) supports the `upload` command. We need to port that command (https://github.com/getappmap/appland-cli#upload) to the JS/TS CLI. We do not need to port the authentication or contexts features. Just expect APPLAND_API_KEY or APPMAP_API_KEY in the environment. stats has already been ported over.

dividedmind:
Adds upload command. Fixes #962 

Still no full feature-parity with the go cli, I'll keep adding to this or as new PRs if this has been merged.

### DIFF
--- a/packages/cli/src/cmds/upload.ts
+++ b/packages/cli/src/cmds/upload.ts
@@ -1,0 +1,113 @@
+import { CommandModule } from 'yargs';
+import { AppMap, loadConfiguration, Mapset } from '@appland/client';
+import { listAppMapFiles } from '../utils';
+import { readFile } from 'fs/promises';
+import assert from 'assert';
+import UI from './userInteraction';
+import { ValidationError } from './errors';
+
+interface Arguments {
+  appmapDir: string;
+  app?: string;
+}
+
+type Metadata = {
+  application?: string;
+  branch?: string;
+  commit?: string;
+};
+
+function applyOrCheck(data: { [x: string]: string | undefined }, key: string, input?: string) {
+  if (!input) return;
+
+  const existing = data[key];
+  if (existing) {
+    if (existing !== input)
+      throw new ValidationError(`"${key}" mismatch ("${existing}" !== "${input}")`);
+  } else {
+    data[key] = input;
+  }
+}
+
+function checkMetadata(current: Metadata, input: unknown) {
+  if (!input) return; // no metadata in appmap, ie. matches
+
+  assert(typeof input === 'object');
+
+  applyOrCheck(current, 'application', input['app']);
+
+  const git = input['git'];
+  if (typeof git !== 'object') return;
+  applyOrCheck(current, 'branch', git['branch']);
+  applyOrCheck(current, 'commit', git['commit']);
+}
+
+async function collect(appmapDir: string): Promise<[string[], Metadata]> {
+  const metadata: Metadata = {};
+  const paths: string[] = [];
+
+  await listAppMapFiles(appmapDir, (path) => {
+    paths.push(path);
+  });
+
+  for (const path of paths)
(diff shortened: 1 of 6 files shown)


packages/cli/src/cmds/upload.ts
```diff
@@ -0,0 +1,113 @@
...
+  let total = paths.length;
+  const { baseURL } = loadConfiguration();
+  UI.progress(`Uploading ${total} AppMaps for ${app} to ${baseURL}...`);
+
+  const uuids: string[] = [];
+  let processed = 0;
+  for (const path of paths) {
+    processed += 1;
+    UI.status = `[${processed}/${total}] ${path}`;
+    const data = await readFile(path); //  TODO: stream this instead of slurping the whole thing?
```
kgilpin, line 83: One thing we do need to support in this v1 is the max file size parameter.

dividedmind: > One thing we do need to support in this v1 is the max file size parameter.

It isn't a parameter in the original :)

symwell (commented):
In one of the runs I got this:
```
~/src/appmap-js/packages/cli$ npm start upload ~/src/forem/tmp/appmap

> @appland/appmap@3.64.2 start
> ts-node src/cli.ts upload /home/test/src/forem/tmp/appmap

Examining AppMaps...
ValidationError: "commit" mismatch ("68e79de6e400a5c5ec31287c68e49e287eb8a540" !== "19c6df892255e0ed11478705330c73e0ab4159be")
    at applyOrCheck (/home/test/src/appmap-js/packages/cli/src/cmds/upload.ts:26:13)
    at checkMetadata (/home/test/src/appmap-js/packages/cli/src/cmds/upload.ts:42:3)
    at collect (/home/test/src/appmap-js/packages/cli/src/cmds/upload.ts:54:5)
    at async Object.handler (/home/test/src/appmap-js/packages/cli/src/cmds/upload.ts:61:29)
```

Is this ok?  As a user I understand there's some mismatch but don't understand what I did wrong.  This is on `sw/use_appmap_to_create_openapi_doc` from https://github.com/symwell/forem

packages/cli/src/cmds/upload.ts
kgilpin: Well, whatever you want to call this:

https://github.com/getappmap/appland-cli/blob/master/cmd/upload.go#L21

kgilpin: Otherwise the upload will be rejected by the server and it will fail without recourse.

kgilpin (changes requested):
`-d` should be required to set the working directory. The `--appmap-dir` should specify the location of AppMaps relative to the working directory. If `--appmap-dir` is not specified, the default value is read from the _appmap.yml_. If no dir is specified, and no default is available, it's an error. This is the common pattern throughout the rest of the CLI, so for consistency, this command should work the same.

packages/cli/src/cmds/upload.ts
```diff
@@ -0,0 +1,136 @@
...
+          MAX_APPMAP_SIZE / 1024
+        } KiB.`,
+        'Use --force if you want to upload it anyway.',
+      ].join('\n')
+    );
+}
+
+const command: CommandModule<{}, Arguments> = {
+  command: 'upload [appmap-dir]',
+  builder: {
```
kgilpin, line 116: Add option `-d` - working directory.

packages/cli/src/cmds/upload.ts
```diff
@@ -0,0 +1,136 @@
...
+    appmapDir: {
```
kgilpin, line 117: Change to `--appmap-dir` option flag. Resolve relative to the working directory.

PR #999, opened 2023-02-07
fix: add missing @appland/client dependency

lachrist:
fixes https://github.com/getappmap/appmap-js/issues/998

I don't understand how this is only noticed now:

- `packages/cli/package.json` did not have a `@appland/client` dependency since at least edea424c8ad533118e23c6e983322ce44932b0c8 (3.50.0)
- `packages/cli/src/cmds/upload.ts` did have `import { AppMap, loadConfiguration, Mapset } from '@appland/client';` since its creation 7f83aa52890c86c604acb1233ca1b8b21dd74b69

Maybe it would be a good idea to invest in some e2e testing so that it does not happen in the future.

**EDIT `upload.ts` was only recently added, please disregard the above**

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -68,6 +68,7 @@
     "vuex"
   ],
   "dependencies": {
+    "@appland/client": "workspace:^1",
     "@appland/components": "workspace:^2",
     "@appland/diagrams": "workspace:^1",
     "@appland/models": "workspace:^2.0.0",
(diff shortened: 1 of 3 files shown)


lachrist:
I use `npm pack` in my 2e2 testing in `appmap-agent-js` to be protected against this kind of failures.

dividedmind (approved):
Good catch, thanks!

PR #1004, opened 2023-02-08
test: Improved smoke test

### LINKED ISSUE #1002: E2E smoke test catches all kinds of dependency errors
Our smoke test should have caught #998 , but didn't

dividedmind:
For some reason our smoke test didn't catch new missing dependency of cli on the client. I've realized yarn doesn't always use the correct packages installing files when resolutions include relative path. Changing the smoke test to instead use absolute paths seems to resolve the issue and allow the test to catch it.

Fixes #1002

### DIFF
--- a/ci/make-smoke-package-json.mjs
+++ b/ci/make-smoke-package-json.mjs
@@ -1,0 +1,29 @@
+#!/usr/bin/env node
+
+import fs from 'fs/promises';
+import { join } from 'path';
+import { pathToFileURL } from 'url';
+
+const pkgs = (await fs.readdir('.'))
+  .filter((f) => /@appland-.*\.tgz/.test(f))
+  .map((f) => f.replace(/\.tgz$/, ''));
+
+const pwd = process.cwd();
+
+function file(pkg) {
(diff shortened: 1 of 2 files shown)


ci/make-smoke-package-json.mjs
```diff
@@ -0,0 +1,28 @@
...
+import { join } from 'path';
+
+const pkgs = (await fs.readdir('.'))
+  .filter((f) => /@appland-.*\.tgz/.test(f))
+  .map((f) => f.replace(/\.tgz$/, ''));
+
+const pwd = process.cwd();
+
+function file(pkg) {
+  return 'file://' + join(pwd, pkg) + '.tgz';
```
lachrist, line 13: What about: ``fromPathToFileURL(join(pwd, `${pkg}.tgz`))``?

lachrist (approved):
APPROVED

dividedmind:
> APPROVED
> 
> Do we know why the eslint rule `no-extraneous-dependencies` did not catch it? You told me it was enable but I don't see it explicitly stated in any eslint config files. Is it enabled by default? Or in some preset?

It's in a preset, but due to #1005 eslint is broken in this project anyway.

PR #1022, opened 2023-02-13
Don't provide `requestBody` for `DELETE` requests

dustinbyrne:
Fixes https://github.com/getappmap/appmap-js/issues/1010


dividedmind (commented):
:+1:

PR #1026, opened 2023-02-14
fix: Fix `git status` call causing sendTelemetry to exit

symwell:
Fix the [error](https://github.com/getappmap/appmap-js/pull/960#discussion_r1100675607) of `spawn` called in git status causing sendTelemetry to exit.

- For gitState, use `commandRunner.run` instead of `spawn`. It's an existing general abstraction. Change commandRunner to return the exit code.
- To use commandRunner in `telemetry.ts` put it under package `telemetry`. `telemetry.ts` was used through a symlink, so continue to use symlinks.  I'm not sure if using symlinks and keeping telemetry a private package are a good idea. This pr *tries* to not address this but perhaps symlinks lead to the smoketest failing.
```
~/src/appmap-js/packages/scanner/src$ ls -l commandStruct.ts commandRunner.ts telemetry.ts
lrwxrwxrwx 1 test test 32 Feb 14 10:54 telemetry.ts -> ../../telemetry/src/telemetry.ts
lrwxrwxrwx 1 test test 36 Feb 14 10:54 commandStruct.ts -> ../../telemetry/src/commandStruct.ts
lrwxrwxrwx 1 test test 36 Feb 14 10:54 commandRunner.ts -> ../../telemetry/src/commandRunner.ts
```
- Fix lint issues under [node 18](https://app.travis-ci.com/github/getappmap/appmap-js/builds/260601956).
- Add `exports` in `package.json` to address smoketest failing [on CI](https://app.travis-ci.com/github/getappmap/appmap-js/jobs/595855958#L4653) as a result of this refactoring, possibly because of the symlinks, based on [this](https://stackoverflow.com/questions/71917783/typescript-builds-files-with-incorrect-path-in-monorepo) info.


packages/telemetry/src/telemetry.ts
```diff
@@ -289,22 +293,31 @@ class GitProperties {
...
+      }
+    }
+    switch (returnCode) {
+      case 127:
+        return GitState.NotInstalled;
+      case 128:
+        return GitState.NoRepository;
+      default:
+        return GitState.Ok;
+    }
```
dustinbyrne, line 320: What about this actually fixed the issue?

dustinbyrne:
Am I doing something wrong?
```
$ APPMAP_TELEMETRY_DEBUG=1 yarn start openapi --appmap-dir ~/dev/applandinc/appmap-server/tmp/appmap/rspec
Error: Cannot find module '/home/db/dev/applandinc/appmap-js/node_modules/@appland/appmap/built/cmds/agentInstaller/commandStruct'
    at createEsmNotFoundErr (node:internal/modules/cjs/loader:1012:15)
    at finalizeEsmResolution (node:internal/modules/cjs/loader:1005:15)
    at resolveExports (node:internal/modules/cjs/loader:522:14)
    at Function.Module._findPath (node:internal/modules/cjs/loader:562:31)
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:971:27)
    at Function.Module._resolveFilename.sharedData.moduleResolveFilenameHook.installedValue [as _resolveFilename] (/home/db/dev/applandinc/appmap-js/node_modules/@cspotcode/source-map-support/source-map-support.js:679:30)
    at Function.Module._load (node:internal/modules/cjs/loader:833:27)
    at Module.require (node:internal/modules/cjs/loader:1057:19)
    at Module.patchedRequire [as require] (/home/db/dev/applandinc/appmap-js/node_modules/diagnostic-channel/dist/src/patchRequire.js:16:46)
    at require (node:internal/modules/cjs/helpers:103:18) {
  code: 'MODULE_NOT_FOUND',
  path: '/home/db/dev/applandinc/appmap-js/node_modules/@appland/appmap/package.json'
}
```

symwell:
I'm getting the same error.

packages/telemetry/src/telemetry.ts
symwell: When I remove the exports commit https://github.com/getappmap/appmap-js/pull/1026/commits/e38fd196d417c75fb980848fc64f623054742827 and add debug statements I see `sendTelemetry` returns without the process exiting.

rm_exports

packages/telemetry/src/telemetry.ts
```diff
@@ -4,12 +4,15 @@ import * as os from 'os';
...
 import { TelemetryClient, setup as AppInsights } from 'applicationinsights';
 import Conf from 'conf';
-import { ChildProcess, exec as execCallback, spawn as spawnCallback } from 'child_process';
+import { exec as execCallback } from 'child_process';
 import { promisify } from 'util';
 import { PathLike } from 'fs';
+import { CommandReturn } from '@appland/appmap/src/cmds/agentInstaller/commandStruct';
+import CommandStruct from '@appland/appmap/src/cmds/agentInstaller/commandStruct';
+import { run } from '@appland/appmap/src/cmds/agentInstaller/commandRunner';
+import { ChildProcessError } from '@appland/appmap/src/cmds/errors';
```
dustinbyrne, line 13: This would mean anything importing telemetry also must include `@appland/appmap` as a dependency. It won't be handled automatically by dependency management tools because `@appland/telemetry` isn't a real package - it's imported via a symlink. 

Won't this break the `@appland/scanner` binary once published? I don't see any addition of `@appland/appmap` to the dependencies.

packages/telemetry/src/telemetry.ts
```diff
@@ -4,12 +4,15 @@ import * as os from 'os';
 import { sync as readPackageUpSync } from 'read-pkg-up';
 import { TelemetryClient, setup as AppInsights } from 'applicationinsights';
 import Conf from 'conf';
-import { ChildProcess, exec as execCallback, spawn as spawnCallback } from 'child_process';
+import { exec as execCallback } from 'child_process';
 import { promisify } from 'util';
 import { PathLike } from 'fs';
+import { CommandReturn } from '@appland/appmap/src/cmds/agentInstaller/commandStruct';
+import CommandStruct from '@appland/appmap/src/cmds/agentInstaller/commandStruct';
+import { run } from '@appland/appmap/src/cmds/agentInstaller/commandRunner';
```
dustinbyrne, line 12: Isn't `commandStruct` part of this package now?

packages/cli/package.json
```diff
@@ -26,6 +26,12 @@
       "prettier --write"
     ]
   },
+  "exports": {
+    "./src/cmds/agentInstaller/commandStruct": "./built/cmds/agentInstaller/commandStruct",
+    "./src/cmds/agentInstaller/commandRunner": "./built/cmds/agentInstaller/commandRunner",
+    "./src/utils": "./built/utils",
+    "./src/cmds/errors": "./built/cmds/errors"
+  },
```
dustinbyrne, line 34: What's this?

dustinbyrne:
This seems like a big change and I'm not really sure what the fix is. Originally the Git contributors and git status calls were not resolving before the process exited and switching to `commandRunner` seems to have fixed that? `commandRunner` is also using `spawn` internally so what is the difference between then and now? It seems like what ever the delta is should be the actual fix.

dustinbyrne (changes requested):
This seems like a big change and I'm not really sure what the fix is. Originally the Git contributors and git status calls were not resolving before the process exited and switching to `commandRunner` seems to have fixed that? `commandRunner` is also using `spawn` internally so what is the difference between this and the previous implementation? It seems like what ever the delta is should be the actual fix.

packages/telemetry/src/telemetry.ts
symwell: It seems it will break it.  I think changing the symlink should be changed to be right under `src/` so that the full package name isn't needed.  Now addressed in https://github.com/getappmap/appmap-js/pull/1026/commits/2dc2649bf31ca31d8de922cbc7f52a65c6fbee1e

packages/telemetry/src/telemetry.ts
symwell: Same as above, the symlink should be in `src/`.

symwell:
The smoketests error is because commandRunner imports utils.ts (for verbose) and cmds/errors (for ChildProcessError).  Then utils.ts imports @appland/models.  This can lead to a long dependency where one file requires many others, and also lead to dependency management to break because telemetry importing models doesn't necessarily mean the packages that import telemetry also import models.

If this pr is the path forward, telemetry would need to become a proper package instead of a private package with symlinks.

symwell:
Found smaller fix, added in https://github.com/getappmap/appmap-js/pull/1031.

Closing.

PR #1029, opened 2023-02-15
fix: Don't error out with yarn 1 projects without workspaces

### LINKED ISSUE #1020: Failing to detect yarn 1 project throws exception and exits.
Started to see this error
```
An error occurred while running the command: yarn workspaces info --json
The command exited with code 1
>  yarn workspaces v1.22.19
>  error Cannot find the root of your workspace - are you sure you're currently in a workspace?
>  info Visit https://yarnpkg.com/en/docs/cli/workspaces for documentation about this command.
```

[Slack](https://applanding.slack.com/archives/C01H8V71K51/p1676310540497459)

I think it needs a try/catch.  The modeled [test](https://github.com/getappmap/appmap-js/pull/936/files#diff-88a638717307cb37ed82e64548ab4b1947cd23936c8bd64d8b0f12d4b5e7fea2R995) `'yarn getYarnSubprojects for yarn 1, error'` for this error expects output from stdout but instead an exception is raised.

It was definitely introduced in https://github.com/getappmap/appmap-js/pull/936/files

<del>The testcase can be improved, but I think this isn't a bug.  If a project contains leftover yarn files then the installer did the right thing of detecting yarn, attempting to use it, not finding a workspace, and throwing an exception.</del>

This is breaking any JS install without yarn >= 2

dividedmind:
Fixes #1020

### DIFF
--- a/packages/cli/src/cmds/agentInstaller/projectConfiguration.ts
+++ b/packages/cli/src/cmds/agentInstaller/projectConfiguration.ts
@@ -3,12 +3,12 @@ import chalk from 'chalk';
 import { promises as fs, constants as fsConstants } from 'fs';
 
 import { basename, join, resolve } from 'path';
-import { AbortError, InvalidPathError } from '../errors';
+import { AbortError, ChildProcessError, InvalidPathError } from '../errors';
 import UI from '../userInteraction';
 import AgentInstaller from './agentInstaller';
 import getAvailableInstallers from './installers';
 import { YarnInstaller } from './javaScriptAgentInstaller';
-import CommandStruct from './commandStruct';
+import CommandStruct, { CommandReturn } from './commandStruct';
 import { run } from './commandRunner';
 
 export interface ProjectConfiguration {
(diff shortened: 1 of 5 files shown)


packages/cli/src/cmds/agentInstaller/projectConfiguration.ts
```diff
@@ -125,7 +125,13 @@ async function getYarnSubprojectsVersionOne(
...
 ) {
   const cmd = new CommandStruct('yarn', ['workspaces', 'info', '--json'], installer.path);
-  const output = await run(cmd);
+  let output: CommandReturn;
+  try {
+    output = await run(cmd);
+  } catch (e) {
+    if (ChildProcessError.check(e)) return;
+    throw e;
+  }
```
symwell, line 134: Since control flow returns or stops on an exception, I think line 140 can be removed.  This line was trying to parse the string `error` and return.
```
if (line.startsWith('error')) return;
```

Similarly, I think the existing test `yarn getYarnSubprojects for yarn 1, error` can be removed.  It was expecting to parse the string `error`, which isn't produced as output.  And the new test you added seems to cover this.

symwell (commented):
I got the following error.  It could be my local environment.
yarn1

dividedmind:
> I got the following error. It could be my local environment.
> 
> Does this work for you locally?
> 
> ```
> npm start install -- -d /home/test/src/sample_app_6th_ed/
> ```
> 
> yarn1

Yeah, the problem here is when you start with `yarn start` or `npm start` yarn litters the environment with its version pin. You need to `yarn build` and then run with `node build/cli.js` to avoid this problem.

packages/cli/src/cmds/agentInstaller/projectConfiguration.ts
dividedmind: Good point, thanks.

symwell (approved):
Running with 
```
node built/cli.js install /home/test/src/sample_app_6th_ed
```
didn't work for me. Approved assuming it worked for you.

dividedmind:
> Running with
> 
> ```
> node built/cli.js install /home/test/src/sample_app_6th_ed
> ```
> 
> didn't work for me. Approved assuming it worked for you.

Ah, this is because the sample_app_6th_ed is broken and doesn't have the yarn release committed.

symwell:
fyi the deploy step failed https://app.travis-ci.com/github/getappmap/appmap-js/jobs/596119965

PR #1072, opened 2023-02-24
Seq diagram backup

ahtrotta:
backup branch


ahtrotta:
No longer needed since the branch that this backs up was successfully released.

PR #1076, opened 2023-02-27
feat: Generate sequence diagram as PNG

kgilpin:
Use our own implementation of sequence diagram to provide png output.

Adds `--format png` to the `@appland/appmap sequence-diagram`. 

Rendering is performed using puppeteer in a headless browser.

Previously, to get a PNG it was necessary to generate PlantUML and then use `java -jar plantuml.jar -t(png|svg)` to get an AppMap sequence diagram as an image. 

This way of course, we also preserve all our own styling and formatting.

The diagram is rendered with an `interactive = false` flag, so that features like open/close and hover behaviors are removed.


packages/components/src/components/sequence/CallLabel.vue
```diff
@@ -64,7 +69,14 @@ export default {
   },
 
   computed: {
+    classes(): string[] {
+      const result = ['label'];
+      if (this.interactive) result.push('interactive');
+      return result;
+    },
```
dustinbyrne, line 76: This can also be
```ts
{
  classes(): { [key: string]: boolean } {
    return {
      label: true,
      interactive: this.interactive,
    };
  }
}
```
It'll use every key name where the value is truthy.

packages/components/src/components/DiagramSequence.vue
```diff
@@ -111,6 +132,9 @@ export default {
   },
 
   methods: {
+    loadData(data) {
+      this.serializedDiagram = data;
```
dustinbyrne, line 136: If this is the primary means of loading a diagram, `serializedDiagram` should be a `data` element instead of a `prop`. `prop` data is meant to be controlled by a parent component only.

dustinbyrne (approved):
Looks good. I generated an image from the command line and that looks to be working great.

packages/components/src/components/DiagramSequence.vue
kgilpin: serializedDiagram can also be set directly, to render the diagram without the AppMap wrapping stuff.

PR #1079, opened 2023-02-28
Remove assert and util from vue code

### LINKED ISSUE #1098: Remove assert and util from `components` and `sequence-diagram` packages

ahtrotta:
Fixes #1098 

Remove node library code from Vue components to avoid complex polyfilling.

### DIFF
--- a/packages/components/src/components/DiagramSequence.vue
+++ b/packages/components/src/components/DiagramSequence.vue
@@ -83,7 +83,7 @@ export default {
         const specification = Specification.build(appMapObj, { loops: true });
         result = buildDiagram('<an AppMap file>', appMapObj, specification);
       }
-      if (!result) throw Error;
+      if (!result) throw Error();
       return result;
     },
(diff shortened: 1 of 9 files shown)


packages/components/src/components/DiagramSequence.vue
```diff
@@ -84,7 +83,7 @@ export default {
         const specification = Specification.build(appMapObj, { loops: true });
         result = buildDiagram('<an AppMap file>', appMapObj, specification);
       }
-      assert(result);
+      if (!result) throw Error;
```
dustinbyrne, line 86: These should all be `throw new Error()` otherwise the stack won't be reported

dustinbyrne (changes requested):

packages/components/src/components/DiagramSequence.vue
ahtrotta: Okay, good to know! I fixed this and rebased the branch.

PR #1090, opened 2023-03-04
Add appmapping of some sub-projects

### LINKED ISSUE #1088: Enable mapping in (some) appmap-js test scripts

dividedmind:
Fixes #1088


dividedmind:
To avoid useless overhead in CI, I changed it so that appmapping is not enabled by default, but available by running `yarn appmap`. @lachrist, could you take a look at the config files? I've mostly lifted them from your `dog` branch, but there could be differences.

packages/client/appmap.yml
```diff
@@ -0,0 +1,12 @@
...
+recorder: jest
+exclude:
+- \[anonymous\]
+packages:
+- regexp: (^|/)node_modules/
+  enabled: false
+- regexp: ^../
+  enabled: false
+- regexp: ^
+  enabled: true
```
lachrist, line 12: Note that `enabled: true` is the default if we want to make the config file shorter. I still prefer the explicit `enabled` value though.

packages/client/package.json
```diff
@@ -7,6 +7,7 @@
     "dist"
   ],
   "scripts": {
+    "appmap": "appmap-agent-js -- jest --no-cache",
```
lachrist, line 10: In the next release, `--no-cache` will no longer be necessary.

packages/sequence-diagram/appmap.yml
```diff
@@ -0,0 +1,45 @@
+ordering: chronological
+command: yarn exec jest --no-cache
+recorder: jest
+exclude:
+- \[anonymous\]
+- matchAction
```
lachrist, line 6: Any reason why this is not scoped to a single package?

lachrist (commented):
The config files looks legit. In  my dog branch I used a single config files to share some fields across different sub projects. But @kgilpin did not like it so one appmap config file per sub-project seems to be the way forward. I think this is okay to merge this if this is not part of CI yet. I still have some feedbacks to process.

lachrist:
Now, pretty much all the sub projects can be appmaped with (various degree of success)

brikelly:
@lachrist can we just close this PR now, since it sounds like other changes have addressed these issues in other PRs?

packages/client/appmap.yml
dividedmind: Yeah, agreed.

packages/sequence-diagram/appmap.yml
dividedmind: No reason, thanks.

kgilpin (approved):
Thank you for all this. Reminder - provide `name`, `language`, `appmap_dir` to *appmap.yml*.

PR #1092, opened 2023-03-06
Sequence diagram legend

### LINKED ISSUE #1066: Sequence Diagram has a legend
When a user clicks on the 'i' icon in the lower right hand corner of the Dependency Map or Trace View, they get a legend explaining what the different visual indicators on the map mean.

Image

Image

The Sequence Diagram is not included in this information. Clicking the 'i' icon in the Sequence Diagram view produces the same double legend that is used for the Dependency Map and Trace View.

Quick fix would be to add legend content for the Sequence Diagram. Better solution would be to have the 'i' icon produce a context-specific legend for the visualization that the user is looking at.

ahtrotta:
Fixes #1066 

image


ahtrotta:
The lifeline is the whole vertical piece, including the box and the dashed line and the rectangular lines. See [here](https://sparxsystems.com/resources/tutorials/uml2/sequence-diagram.html#:~:text=complex%20procedural%20logic.-,Lifelines,which%20owns%20the%20sequence%20diagram.) for explanation.

image

The order of these two should be switched, the dashed line should be labeled `Lifeline`, and the rectangle should be `Active lifeline`.

image

This should be `Hide lifeline`.

image

This could be `Call arrow`.

ahtrotta:
Since loop boxes have a grey header, could you make the legend have a similar grey header? Basically make the top border extra thick to more closely resemble the actual loop box.

image

brikelly (approved):
Approved, thanks

PR #1095, opened 2023-03-07
fix alignment for text hover for export & share buttons

typaulhus:
Make the controls for AppMap diagrams styled consistently. The purple styling was initially meant to draw attention to the Share button, but should remain consistent with other control styles.

## Before
Screenshot 2023-03-06 at 10 11 03 AM

## After 
Screenshot 2023-03-07 at 10 47 57 AM


ahtrotta (approved):
Looks good!

PR #1106, opened 2023-03-15
Updates to handle large appmaps

### LINKED ISSUE #1104: Prune CLI command returns excluded events
To better handle large AppMaps, the prune command should return what events were excluded so that we can notify the user when we automatically prune a large AppMap.

ahtrotta:
Fixes #1104 

Alter the `prune` CLI command:

- [x] When using the `--size` option, prune the most expensive functions based on the entire appmap rather than the most expensive functions per "chunk"
- [x] Option to return json of what functions were excluded with `--format json`
- [x] Allow users to specify what to prune using the `--filter` option followed by a base64url-encoded string representing the filter state
- [x] The size argument was changed from positional to an option (`--size`, `-s`). This is necessary because it doesn't make sense to include a size when using the `--filter` option.
- [x] Added the `AppMapFilter#apply` function to apply a filter state object to an instance of `AppMapFilter`. This allows this code to be re-used by Vue components and the CLI.

**Example:**

command: 
```
node built/cli.js prune tests/unit/fixtures/stats/appmap/Microposts_interface_micropost_interface.appmap.json --format json -s 500kb
```

output:
```
{
  "exclusions": [
    "function:logger/Logger::LogDevice#write",
    "function:actionpack/ActionDispatch::Request::Session#[]",
    "function:activerecord/ActiveRecord::Relation#records",
    "function:actionview/ActionView::Resolver#find_all",
    "function:app/helpers/UsersHelper#gravatar_for"
  ]
}
```

**Example (with `--filter`):**

The filter for the command below is:
```
{
  hideName: [
    function:logger/Logger::LogDevice#write
    function:activerecord/ActiveRecord::Relation#records
    function:actionpack/ActionDispatch::Request::Session#[]
  ]
}
```

command:
```
node built/cli.js prune --filter eyJoaWRlTmFtZSI6WyJmdW5jdGlvbjpsb2dnZXIvTG9nZ2VyOjpMb2dEZXZpY2Ujd3JpdGUiLCJmdW5jdGlvbjphY3RpdmVyZWNvcmQvQWN0aXZlUmVjb3JkOjpSZWxhdGlvbiNyZWNvcmRzIiwiZnVuY3Rpb246YWN0aW9ucGFjay9BY3Rpb25EaXNwYXRjaDo6UmVxdWVzdDo6U2Vzc2lvbiNbXSJdfQ tests/unit/fixtures/stats/appmap/Microposts_interface_micropost_interface.appmap.json
```

output:
```
done
```

### DIFF
--- a/packages/cli/src/cmds/prune/prune.ts
+++ b/packages/cli/src/cmds/prune/prune.ts
@@ -48,18 +48,26 @@ function parseSize(size: string) {
 export default {
   command: 'prune <file> <size>',
 
-  describe: 'Prune an appmap file down to the given size',
+  describe: 'Make an appmap file smaller by removing events',
 
   builder: (argv: Yargs.Argv) => {
     argv.option('output-dir', {
-      describe: 'Specifies the output directory. Pruned files will be written here.',
+      describe: 'Specifies the output directory',
       type: 'string',
       default: '.',
       alias: 'o',
     });
+
+    argv.option('format', {
+      describe: 'How to format the output',
+      choices: ['json', 'text'],
+      default: 'text',
+    });
+
     argv.positional('file', {
       describe: 'AppMap to prune',
     });
+
     argv.positional('size', {
       describe: 'Prune input file to this size',
     });
@@ -71,6 +79,15 @@ export default {
 
     const outputPath = `${argv.outputDir}/${basename(argv.file)}`;
     fs.writeFileSync(outputPath, JSON.stringify(appMap));
-    console.log('done');
+
+    let message = 'done';
+    if (argv.format === 'json' && !argv.fqids)
+      message = JSON.stringify(
+        { exclusions: Array.from(appMap.data.exclusions.values()) },
+        null,
+        2
+      );
+
+    console.log(message);
   },
 };
--- a/packages/cli/src/cmds/prune/prune.ts
+++ b/packages/cli/src/cmds/prune/prune.ts
@@ -3,7 +3,7 @@ import JSONStream from 'JSONStream';
 import { basename } from 'path';
 import Yargs from 'yargs';
 import { handleWorkingDirectory } from '../../lib/handleWorkingDirectory';
-import { AppMap, pruneAppMap } from './pruneAppMap';
+import { AppMap, pruneAppMap, removeEventsByFqid } from './pruneAppMap';
 
 async function fromFile(filePath): Promise<AppMap> {
   let data: AppMap = { events: [] };
@@ -47,7 +47,7 @@ function parseSize(size: string) {
 }
 
 export default {
-  command: 'prune <file> <size>',
+  command: 'prune <file>',
 
   describe: 'Make an appmap file smaller by removing events',
 
@@ -69,8 +69,11 @@ export default {
       describe: 'AppMap to prune',
     });
 
-    argv.positional('size', {
+    argv.option('size', {
       describe: 'Prune input file to this size',
+      default: '15mb',
+      type: 'string',
+      alias: 's',
     });
 
     argv.option('directory', {
@@ -78,12 +81,25 @@ export default {
       type: 'string',
       alias: 'd',
     });
+
+    argv.option('fqids', {
+      describe: 'Remove events from the map by fqid',
+      type: 'array',
+    });
   },
 
   handler: async (argv: any): Promise<void> => {
     handleWorkingDirectory(argv.directory);
-    const size = parseSize(argv.size);
-    const appMap = pruneAppMap(await fromFile(argv.file), size);
+    const map = await fromFile(argv.file);
+    let appMap: AppMap;
+
+    if (argv.fqids) {
+      appMap = removeEventsByFqid(map, argv.fqids);
+    } else if (argv.size) {
+      appMap = pruneAppMap(map, parseSize(argv.size));
+    } else {
+      throw Error('Invalid usage');
+    }
 
     const outputPath = `${argv.outputDir}/${basename(argv.file)}`;
     fs.writeFileSync(outputPath, JSON.stringify(appMap));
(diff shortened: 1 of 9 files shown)


packages/cli/src/cmds/prune/prune.ts
```diff
@@ -46,31 +46,60 @@ function parseSize(size: string) {
...
+    });
+
     argv.positional('file', {
       describe: 'AppMap to prune',
     });
-    argv.positional('size', {
+
+    argv.option('size', {
       describe: 'Prune input file to this size',
+      default: '15mb',
```
kgilpin, line 72: By changing this from a positional argument to an option, the change is backwards-incompatible. So, I suggest leaving it as a positional argument - since there's no great reason to break compatibility.

packages/cli/src/cmds/prune/prune.ts
```diff
@@ -46,31 +46,60 @@ function parseSize(size: string) {
...
+      type: 'string',
+      alias: 's',
+    });
+
+    argv.option('events', {
+      describe: 'Events to remove from the map',
```
kgilpin, line 78: How would I use this option?

packages/cli/src/cmds/prune/prune.ts
```diff
@@ -46,31 +46,60 @@ function parseSize(size: string) {
...
 
 export default {
-  command: 'prune <file> <size>',
+  command: 'prune <file>',
 
-  describe: 'Prune an appmap file down to the given size',
+  describe: 'Make an appmap file smaller by removing events',
 
   builder: (argv: Yargs.Argv) => {
     argv.option('output-dir', {
```
kgilpin, line 54: There is a standard `--directory, -d` option that we use for the working directory. Please take a look at some of the other commands and you'll see it. Then you can process that option with `handleWorkingDirectory` which will chdir to that directory. It makes it easy, and consistent, to test all these commands against an external repo.

packages/cli/src/cmds/prune/prune.ts
```diff
@@ -46,31 +46,60 @@ function parseSize(size: string) {
...
-      describe: 'Specifies the output directory. Pruned files will be written here.',
+      describe: 'Specifies the output directory',
       type: 'string',
       default: '.',
       alias: 'o',
     });
+
+    argv.option('json', {
```
kgilpin, line 61: Prefer a `--format, -f` option whose values can be, e.g. `text | json`

kgilpin (changes requested):
Add `--directory, -d`
Maintain backwards compatibility of options, where possible
Describe the `--events` argument and why I would use it
Implement `--format, -f` option, replacing `--json`

packages/models/src/appMapBuilder/index.js
```diff
@@ -190,37 +195,45 @@ class AppMapBuilder extends EventSource {
...
+        // If there's no call event, there's no need to retain this event
+        if (!callEvent) {
+          return false;
+        }
+
+        if (callEvent.http_server_request || callEvent.http_client_request || callEvent.sql_query) {
+          return true;
+        }
+
+        const name = eventName(classMap, e);
```
kgilpin, line 233: `eventName` is

````${callEvent.defined_class}${callEvent.static ? '.' : '#'}${callEvent.method_id}````

Nowadays, we don't use this data any more for filter. A better option is to use `CodeObject#fqid`. I think now is a good time to make that change. 

In fact, probably a better alternative is to allow the command to accept an `appMapFilter` as a command line (e.g. serialized to JSON or Base64) and simply apply that rather than re-implementing a different kind of filtering.

File: packages/cli/src/cmds/prune/prune.ts (hunk 2 above)
Comment (dustinbyrne) [reply packages/cli/src/cmds/prune/prune.ts:78]: Let's expand this to use a `base64url` encoded [`AppMapFilter`](https://github.com/getappmap/appmap-js/blob/main/packages/models/src/appMapFilter.js#L75) - the same as the filters we apply within an AppMap. The logic for applying these filters is already contained within, but it does expect a normalized AppMap as input.

File: packages/cli/src/cmds/prune/prune.ts (hunk 1 above)
Comment (ahtrotta) [reply packages/cli/src/cmds/prune/prune.ts:74]: Specifying a size doesn't make sense when using this command with the `--filter` option, so I think we need to change the size to an option.

File: packages/cli/src/cmds/prune/prune.ts [hunk 6]
```diff
@@ -71,6 +79,15 @@ export default {
...
+
+    let message = 'done';
+    if (argv.format === 'json' && !argv.fqids)
+      message = JSON.stringify(
+        { exclusions: Array.from(appMap.data.exclusions.values()) },
+        null,
+        2
+      );
+
+    console.log(message);
```
dustinbyrne, line 115: Listing the exclusions is a great UX but it's not on by default. I'd recommend listing the exclusions in the text output or defaulting to `json`.

dustinbyrne: These `exclusions` also won't include anything removed via `--filter` which is a little confusing.

The output will say:
```json
{
  "exclusions": []
}
```

Which looks like nothing was applied, even though it was successful.

packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -316,8 +316,7 @@
 </template>
 
 <script>
-import { Buffer } from 'buffer';
```
dustinbyrne, line 319: This `buffer` was a polyfill. I think the front-end distribution should still include the polyfill despite this code moving to `models`, but it's something to keep an eye on.

dustinbyrne (approved):
Approving because I think this is in a good enough state to merge - another round of review is probably overkill.

Though please consider the comment about defaulting `--format` to `json` or including exclusions in `text` output before merging.

packages/components/src/pages/VsCodeExtension.vue
ahtrotta: Yeah, I had the same thought. I think we still need the polyfill either way since we're pulling in `models` as a dependency for `components`.

PR #1107, opened 2023-03-15
Fix missing actors in sequence diagrams

### LINKED ISSUE #1102: Sequence diagrams for Java incorrectly using one lifeline instead of many
This was seen with a large Java application. There were lots of sub-packages under com.techgiant.* but only 1 lifeline appeared 

Image

This surprised the user given that the sequence diagram contained lots of different packages and classes:
Image

The package list was:
Image

dustinbyrne:
If a package contained a class as well as other sub-packages, those sub-packages would not be considered as eligible actors in the sequence diagram. We'll now traverse the entire package tree in all cases. 

Resolves #1102


packages/sequence-diagram/src/specification.ts
```diff
@@ -58,17 +58,16 @@ export default class Specification {
     };
 
     const includeCodeObjects = (co: AppMapCodeObject, ancestors: string[]): void => {
```
kgilpin, line 60: It looks like `ancestors` is unused.

kgilpin (approved):

packages/sequence-diagram/tests/unit/sql.spec.ts
```diff
@@ -32,7 +32,7 @@ describe('Sequence diagram', () => {
     });
     it('sorts to the right side of the diagram', () => {
       const database = findActor(diagram, 'database:Database');
-      assert.strictEqual(database.order, 21000);
+      assert.strictEqual(database.order, 22000);
```
dustinbyrne, line 35: This test had a missing package hidden by this bug. As a result, this had to be bumped.

PR #1117, opened 2023-03-21
Create stats panel in appmap

### LINKED ISSUE #1116: Stats panel in the map

ahtrotta:
Fixes #1116 

This PR creates a Stats panel in the map view. [View it in Storybook here](https://5fe18baf66f2280021e634d0-mjraltrlpw.chromatic.com/iframe.html?args=&id=pages-vs-code--extension&viewMode=story). Everything will work in Storybook except opening the location in source code.

* Each header in the Stats panel can be clicked to change the sorting (click again to switch toggle between ascending and descending order).
* Clicking on the function name will open the Sequence Diagram tab with the function selected and it will pan to the first invocation of that function.
* Clicking on the location will open the source code to that location.
* If a function is currently filtered/hidden then it is grayed out and cannot be clicked on in the Stats panel.
* The Filter modal can be opened while the Stats panel is open. The Filter modal will be displayed on top of the Stats panel.

image


packages/components/src/stories/data/scenario.stats.json
```diff
@@ -0,0 +1,128 @@
+[
```
kgilpin, line 1: In fact, this "stats" is just frequently occurring functions. So, I'd suggest making this data an object that has a single key (for now) like `functions` - anticipating that we may add more stats in the future.

ahtrotta: Okay, makes sense. I passed this off to @typaulhus for now, but I'll change this before it gets merged.

dustinbyrne (approved):

PR #1121, opened 2023-03-22
Robust appmap locating

### LINKED ISSUE #1120: More robust locating of appmaps in CLI

ahtrotta:
Fixes #1120 

- [x] Fixes a bug where there are multiple matches of the same file and the command prompts the user to select between identical files:
image

- [x] When a valid absolute path is specified for `--appmap-file`, the command no longer recursively searches for the file in the current directory or the `--appmap-dir`.
- [x] If no `--appmap-dir` is provided and no appmap directory can be found from the configuration file, default to using the current working directory.
- [x] Prevents recursive searching when an exact match can be found (this was a problem when running the command from VS Code, which automatically sets the current working directory to `~`, so the recursive search was taking a really long time and the command was hanging).

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -74,7 +74,7 @@
     "@appland/client": "workspace:^1",
     "@appland/components": "workspace:^2",
     "@appland/diagrams": "workspace:^1",
-    "@appland/models": "workspace:^2.2.0",
+    "@appland/models": "workspace:^2.4.0",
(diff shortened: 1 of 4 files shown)


kgilpin:
What was the issue?

ahtrotta:
There were a few little things, I updated the PR description with more info. I ran into these problems when trying to execute the `stats` command from within the VS Code extension.

kgilpin (approved):

PR #1141, opened 2023-04-11
feat: Consider loop iterations when diffing sequence diagrams

dustinbyrne:
We'll now consider the number of loop iterations while diffing sequence diagrams. Consider a performance degradation due to an increase in costly iterations. Prior to this change, a sequence diagram diff would report no changes. With this change in place we can accurately describe the change that occurred.

image

There's some additional work to be done to add diff gutters, however this is a more significant change as the way loops and diff channels are rendered are currently incompatible.

### DIFF
--- a/packages/components/src/stories/data/sequence/openai_api_diff.sequence.json
+++ b/packages/components/src/stories/data/sequence/openai_api_diff.sequence.json
@@ -1,0 +1,223 @@
+{
+  "actors": [
+    {
+      "id": "http:HTTP server requests",
+      "name": "HTTP server requests",
+      "order": 0
+    },
+    {
+      "id": "package:app/controllers",
+      "name": "controllers",
+      "order": 1000
+    },
+    {
+      "id": "package:app/models",
+      "name": "models",
+      "order": 2000
+    },
+    {
+      "id": "external-service:api.openai.com",
+      "name": "api.openai.com",
+      "order": 6000
+    },
+    {
+      "id": "database:Database",
+      "name": "Database",
+      "order": 7000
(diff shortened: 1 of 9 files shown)


packages/components/src/stories/data/sequence/openai_api_diff.sequence.json
```diff
@@ -0,0 +1,223 @@
...
+    }
+  ],
+  "rootActions": [
+    {
+      "nodeType": 4,
+      "callee": "http:HTTP server requests",
+      "route": "GET /v1/code/completions",
+      "status": 200,
+      "digest": "ec900b9d1789e528e360fe2044ca7872aaa8c37aa32cc1efc19a5af5031b87a4",
+      "subtreeDigest": "unknown",
```
kgilpin, line 36: Something is wrong here

packages/sequence-diagram/src/mergeWindow.ts
```diff
@@ -50,7 +50,7 @@ const buildLoop = (merge: Merge): Loop => {
   return {
     nodeType: NodeType.Loop,
     count: merge.length,
-    digest: 'loop',
+    digest: ['loop', merge.length].join(':'),
```
kgilpin, line 53: If we report the number of loop iterations as a change to the digest, then this is going to make the diagram digest pretty unstable. 

We may need to think about this as a metric that we track rather than as a structural change.

kgilpin:
If we report loop count changes as a structural diff this might end up making the diagram digest quite unstable.

I don't think we know yet.

PR #1161, opened 2023-04-26
feat: Remove deprecated fingerprints

kgilpin:
* Drop `info`, `trace` and `update` from the AppMap index.
* Drop the `fingerprints` field from the updated metadata.
* Drop the flag that was used to determine whether to print the canonical files or not. Always print them.
* Remove obsolete `diff` command that used these old fingerprints.

Index directory will now look like this (minus sequence.json, which was generated elsewhere):

Screen Shot 2023-04-26 at 10 57 32 AM


apotterri (approved):
LGTM

PR #1163, opened 2023-04-26
fix: SQL parsing

kgilpin:
Fix SQL parsing case where `type` is specified but `variant` is not.

Also:

* Don't try and parse truncated queries
*  Null pointer in sequence diagram diff


apotterri (changes requested):
The changes to individual files look ok here. The history is a mess, though -- there seem to be lots of changes that are unrelated to "SQL parsing", and the same files are touched in multiple commits.

Also, is this really meant to be merged into `feat/remove-deprecated-fingerprints`?

kgilpin:
Not sure how that happened - I’ll clean it up. Thanks for the review :-)

PR #1168, opened 2023-05-02
User can expand sequence diagram packages to classes

### LINKED ISSUE #1166: Sequence Diagram Actors can be expanded from packages to classes

ahtrotta:
Fixes #1166

For some reason the Travis build isn't attached anymore, but you can find it [here](https://app.travis-ci.com/github/getappmap/appmap-js/builds/262948442).

[Here](https://www.chromatic.com/component?appId=5fe18baf66f2280021e634d0&csfId=pages-vs-code--extension-with-default-sequence-view&buildNumber=2186&k=645bdb3bb4699530485605ba-1200-interactive-true&h=10&b=-5) is the interactive chromatic build where you can try out this feature.

- [x] Adds an expand button in the top left of the lifeline box for packages. When clicked, the classes within the package will become lifeline boxes that have a collapse button:

image

- [x] The collapse button will collapse all of the classes back into the package.
- [x] Classes will expand out and will render adjacent to each other to the right of the package that they were expanded from.
- [x] The actor labels are now the appropriate colors.
- [x] The state provided by `getState` now includes the expanded packages.

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -29,7 +29,7 @@
   "dependencies": {
     "@appland/diagrams": "workspace:^1.6.2",
     "@appland/models": "workspace:^2.6.0",
-    "@appland/sequence-diagram": "workspace:^1.5.0",
+    "@appland/sequence-diagram": "workspace:^1.5.2",
     "buffer": "^6.0.3",
     "diff": "^5.1.0",
     "dom-to-svg": "^0.12.2",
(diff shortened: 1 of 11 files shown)


kgilpin:
Hey, glad you got this working. 

We may need to include a way to reverse this action. Otherwise suppose a package expands to many classes and the user wants to reverse it. How does that work?

And/or maybe we need to indicate how many classes are in the package so the user has an idea what’s going to happen.

ahtrotta:
> Hey, glad you got this working.
> 
> We may need to include a way to reverse this action. Otherwise suppose a package expands to many classes and the user wants to reverse it. How does that work?
> 
> And/or maybe we need to indicate how many classes are in the package so the user has an idea what’s going to happen.

I forgot to reply earlier, but there is now an option to collapse the classes back into a package. Also, the number of classes in the package is displayed in parentheses next to the package name.

dustinbyrne (approved):

PR #1173, opened 2023-05-08
feat: 'depends' performance

kgilpin:
* Optimize target file enumeration
* Cache file modified times
* Remove the most verbose logging
* Remove source locations that are blank

Before optimization:
```
node ./built/cli.js depends -d ~/source/land-of-apps/forem > forem.log  11.41s user 6.40s system 169% cpu 10.537 total
```

After optimization:
```
node ./built/cli.js depends -d ~/source/land-of-apps/forem > forem.log  3.86s user 1.59s system 127% cpu 4.296 total
```

### DIFF
--- a/packages/cli/src/utils.ts
+++ b/packages/cli/src/utils.ts
@@ -102,8 +102,6 @@ export async function processFiles(
   if (q.length()) await q.drain();
 }
 
-
-
 /**
  * Finds all occurrances of `fileName` within a base directory. Each match must be a file, not a directory.
  * This is optimized compared to `processFiles` because it does not use `glob`, which is pretty slow for this use case.
@@ -124,11 +122,13 @@ export async function processNamedFiles(
     }
   };
 
+  let matchCount = 0;
   const processDir = async (dir: string) => {
     // If the directory contains the target fileName, add it to the process queue and return.
     const targetFileName = join(dir, fileName);
     const target = await stats(targetFileName);
     if (target && target.isFile()) {
+      matchCount += 1;
       q.push(targetFileName);
       return;
     }
@@ -143,7 +143,7 @@ export async function processNamedFiles(
   };
   await processDir(baseDir);
 
-  if (q.length()) await q.drain();
+  if (matchCount) await q.drain();
 }
 
 /**
--- a/packages/cli/src/utils.ts
+++ b/packages/cli/src/utils.ts
@@ -6,6 +6,7 @@ import { promisify } from 'util';
 import { join } from 'path';
 import assert from 'assert';
 import { Event, ReturnValueObject } from '@appland/models';
+import { readdir, stat } from 'fs/promises';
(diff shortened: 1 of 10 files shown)


packages/cli/src/utils.ts
```diff
@@ -91,18 +92,61 @@ export async function processFiles(
...
+    for (const childName of await readdir(dir)) {
+      const child = await stats(join(dir, childName));
+      if (child && child.isDirectory()) {
+        await processDir(join(dir, childName));
+      }
+    }
+  };
+  await processDir(baseDir);
+
+  if (matchCount) await q.drain();
```
dividedmind, line 147: Perhaps it'd be better to check `q.idle()` instead — I think `q.drain()` might hang if the queue has emptied in the meantime.

packages/cli/tests/unit/depends.spec.ts
```diff
@@ -20,7 +20,7 @@ describe('Depends', () => {
   beforeAll(async () => verbose(process.env.DEBUG === 'true'));
   beforeEach(async () => {
     appMapDir = tmp.dirSync().name;
-    expected = join(appMapDir, 'user_page_scenario').replace(/\\/g, '/');
+    expected = join(appMapDir, 'user_page_scenario');
```
dividedmind, line 23: I assume this was so the tests pass on Windows correctly. I wonder what changed that it's not needed anymore?

dividedmind (approved):
Looks good, but I'd pay attention to that queue so it doesn't end up hanging.

packages/cli/tests/unit/depends.spec.ts
kgilpin: Before, we were using `glob` to enumerate the files. Now we are walking the directory tree ourselves, so the file paths don't require this gsub.

packages/cli/src/utils.ts
kgilpin: Yeah this has always been an issue with these queues. I'll change it `drain unless q.idle()`

kgilpin: I've done this project-wide, let me know what you think.

packages/cli/tests/unit/queue.spec.ts
```diff
@@ -0,0 +1,14 @@
...
+
+describe('async', () => {
+  it('hangs if you drain an idle queue', async () => {
+    const q = queue(async () => {}, 1);
+    expect(q.idle()).toBe(true);
+    let drained = false;
+    q.drain().then(() => (drained = true));
+    setTimeout(() => {
+      assert(!drained, 'drain should not have resolved yet');
+    }, 50);
```
dividedmind, line 12: This timeout will fire after the test case returns. I suggest instead `await`ing a `setTimeout` from `timers/promises` and then asserting after that.

PR #1186, opened 2023-05-16
User can bookmark filter settings

### LINKED ISSUE #1187: User can bookmark appmap filters

ahtrotta:
Fixes #1187 

This feature allows users to save filter settings. Users can save the filter settings with a name and then select from previously saved filter settings from a dropdown menu in the filters menu:

image

Because the saved filters need to propagate across AppMaps in the workspace, the VS Code extension handles some of the logic for this feature (and the JetBrains plugin will have to do the same in the future). Therefore, much of the functionality is not present in Storybook.

Features:

- [x] Initially, only the `AppMap default` filter will be in the dropdown menu.
- [x] After selecting a filter from the menu, clicking the `Apply` button will apply the saved settings to the current AppMap.
- [x] Any of the saved filters can be set to be the default filter, this will make it default for the current workspace in VS Code.
- [x] Any of the saved filters can be deleted except for the `AppMap default` filter.
- [x] There is a `Copy` button that will copy the filter state as a base-64-URL-encoded string to the clipboard.
- [x] The `Root` state is not saved, only the `Filter` state.

There was some discussion about including an option to save a new filter as the default filter. I did a quick mock-up of that and I didn't like how it came out, so if we want to add this feature, we'll need to come up with a better design.

image

### DIFF
--- a/packages/components/src/components/FilterMenu.vue
+++ b/packages/components/src/components/FilterMenu.vue
@@ -120,7 +120,7 @@
         <div class="filters__block-row">
           <div class="filters__block-row-content">
             Filter:
-            <select class="filters__select" v-model="selectedFilter">
+            <select class="filters__select" v-model="selectedSavedFilter">
               <option
                 v-for="savedFilter in savedFilters"
                 :value="savedFilter"
@@ -172,7 +172,7 @@ import {
   REMOVE_ROOT_OBJECT,
   ADD_HIDDEN_NAME,
   REMOVE_HIDDEN_NAME,
-  SET_SELECTED_FILTER,
+  SET_SELECTED_SAVED_FILTER,
   DEFAULT_FILTER_NAME,
 } from '@/store/vsCode';
 import { serializeFilter, base64UrlEncode } from '@appland/models';
@@ -211,12 +211,12 @@ export default {
       return this.$store.state.savedFilters;
     },
 
-    selectedFilter: {
+    selectedSavedFilter: {
       get() {
-        return this.$store.state.selectedFilter;
+        return this.$store.state.selectedSavedFilter;
       },
       set(value) {
-        this.$store.commit(SET_SELECTED_FILTER, value);
+        this.$store.commit(SET_SELECTED_SAVED_FILTER, value);
       },
     },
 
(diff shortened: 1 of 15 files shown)


packages/components/src/components/FilterMenu.vue
```diff
@@ -225,6 +307,19 @@ export default {
...
     },
+
+    defaultButtonClass() {
+      const suffix = this.selectedFilter && this.selectedFilter.default ? '-disabled' : '';
+      return 'filters__button' + suffix;
+    },
+
+    deleteButtonClass() {
+      const suffix =
+        this.selectedFilter && this.selectedFilter.filterName === DEFAULT_FILTER_NAME
```
dustinbyrne, line 318: Is this unique?

packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -973,6 +983,14 @@ export default {
         const { filters, expandedPackages } = state;
         if (filters) this.$store.commit(SET_FILTER, deserializeFilter(filters));
 
+        const base64EncodedFilters = base64UrlEncode(JSON.stringify({ filters }));
+        const selectedFilter =
+          this.$store.state.savedFilters &&
+          this.$store.state.savedFilters.find(
+            (savedFilter) => savedFilter.state === base64EncodedFilters
+          );
+        if (selectedFilter) this.$store.commit(SET_SELECTED_FILTER, selectedFilter);
```
dustinbyrne, line 992: I'm a little confused with the differing plural forms of `filter`, more so now that there's potentially multiple saved filters.

This may be nit-picky, but I'd recommend maybe naming them something different, `filter` vs. `saved filter` or similar. `commit(SET_FILTER, filters)` is also confusing now.

dustinbyrne: It may just be the usage of `SET_SELECTED_FILTER` as `SET_SAVED_FILTERS` is already a thing.

`SET_SELECTED_SAVED_FILTER`?

dustinbyrne (changes requested):
This looks great - what a huge improvement! I left some minor comments below.

One thing I'd like to see before this gets merged is a storybook entry for the filters view. Without it, there's nothing to catch a future regression in Chromatic.

packages/components/src/components/FilterMenu.vue
ahtrotta: Yes, if you try to save a filter with the same name as a previously saved filter, then the old one gets overwritten. I added logic to prevent that from happening with the `AppMap default` filter.

ahtrotta:
> This looks great - what a huge improvement! I left some minor comments below.
> 
> One thing I'd like to see before this gets merged is a storybook entry for the filters view. Without it, there's nothing to catch a future regression in Chromatic.

Thanks! I added a storybook entry.

dustinbyrne (approved):

PR #1189, opened 2023-05-18
Add explicit option to serialize

### LINKED ISSUE #1194: Small updates to `serializeFilter` and `deserializeFilter` in `@appland/models`
These changes are needed for the filter updates [here](https://github.com/getappmap/appmap-js/pull/1186/files).

ahtrotta:
Fixes #1194 

These changes are needed prior to merging [this PR](https://github.com/getappmap/appmap-js/pull/1186).

- [x] Add an `explicit` option to `serializeFilter` that will return the entire filter state, even if some of the declutter settings are the default settings. 
- [x] Correctly handle when an object is passed to `deserializeFilter` with the `hideExternal` field set to `true` or `false`.

### DIFF
--- a/packages/models/src/serialize.js
+++ b/packages/models/src/serialize.js
@@ -106,8 +106,14 @@ export function deserializeFilter(filterState) {
   }
   ['hideExternal', 'hideExternalPaths'].forEach((key) => {
     if (key in filterState) {
-      filter.declutter.hideExternalPaths.on = true;
-      filter.declutter.hideExternalPaths.dependencyFolders = filterState[key];
+      const value = filterState[key];
+
+      if (value && Array.isArray(value)) {
+        filter.declutter.hideExternalPaths.on = true;
+        filter.declutter.hideExternalPaths.dependencyFolders = filterState[key];
+      } else {
+        filter.declutter.hideExternalPaths.on = filterState[key];
+      }
     }
   });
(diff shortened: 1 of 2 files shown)


packages/models/src/serialize.js
```diff
@@ -60,7 +60,7 @@ export function serializeFilter(filter) {
     const filter = declutter[k];
     if (Array.isArray(v) && v.length !== 0) {
       memo[k] = v;
-    } else if (filter && filter.default !== v) {
+    } else if (filter && (options.explicit || filter.default !== v)) {
```
dustinbyrne, line 63: What makes this necessary? In my mind there should be no functional difference between serialized states where the `explicit` flag differs, the full state should be captured either way.

Are the defaults dynamic?

ahtrotta: Yeah, [there is one spot ](https://github.com/getappmap/appmap-js/blob/main/packages/components/src/pages/VsCodeExtension.vue#L791)where we change the default value. But I can make it so that the defaults are not dynamic. Should I do that, or should we merge this PR?

dustinbyrne (approved):

packages/models/src/serialize.js
dustinbyrne: In that case I think this is fine. I'll approve the PR.

PR #1191, opened 2023-05-19
Propagate user identifier(s) via environment variables

dustinbyrne:
This allows code editors to pass along their own identifiers, guaranteeing one single timeline of anonymous identity.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -136,4 +136,4 @@
   "optionalDependencies": {
     "puppeteer": "^19.7.5"
   }
-}
+} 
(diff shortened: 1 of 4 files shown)


kgilpin (commented):
This PR also adds the concept of a Session that's assigned by us? Just checking because that is more than the PR title says.

dustinbyrne:
The session wouldn't be assigned by us, it'd come from the code editor, e.g.
```ts
APPMAP_SESSION_ID: vscode.env.sessionId
```

In its current state we're issuing and maintaining our own session identifiers. This is problematic because it means sessions are split across multiple identifiers.

dustinbyrne:
I'm going to move this into draft for the time being. I'm considering the addition of an authenticated identifier, but I'm not yet sure how that affects this PR.

dividedmind (approved):

PR #1192, opened 2023-05-22
feat: implement basic flame graph diagram

lachrist:
This is a prototype and it is hidden behind `flamegraphEnabled` prop which is false by default.

### DIFF
--- a/packages/components/src/components/DiagramFlamegraph.vue
+++ b/packages/components/src/components/DiagramFlamegraph.vue
@@ -16,11 +16,6 @@ export default {
   name: 'v-diagram-flame',
 
   props: {
-    theme: {
-      type: String,
-      default: 'dark',
-      validator: (value) => ['dark', 'light'].includes(value),
-    },
     appMap: {
       type: Object,
     },
--- a/packages/components/src/components/DiagramFlamegraph.vue
+++ b/packages/components/src/components/DiagramFlamegraph.vue
@@ -1,0 +1,48 @@
+<template>
+  <div class="diagram-flame">
+    <div ref="d3Container"></div>
+  </div>
+</template>
+
+<script>
+// namespace import to better support commonjs
+import * as d3 from 'd3';
+import * as Flamegraph from 'd3-flame-graph';
+import { digestEventArray } from '../lib/flamegraph';
+
+const { flamegraph: createFlamegraph } = Flamegraph;
+
+export default {
+  name: 'v-diagram-flame',
+
+  props: {
+    theme: {
+      type: String,
+      default: 'dark',
+      validator: (value) => ['dark', 'light'].includes(value),
+    },
+    appMap: {
+      type: Object,
+    },
+  },
+
+  data() {
+    return {};
+  },
+
+  mounted() {
+    const flamegraph = createFlamegraph();
+    flamegraph.selfValue(false); // each node's value contains the sum of its children (as well as its own value)
+    flamegraph.width(960);
+    d3.select(this.$refs.d3Container)
+      .datum(digestEventArray(this.appMap.rootEvents()))
+      .call(flamegraph);
+  },
+};
+</script>
+
+<style scoped lang="scss">
+.diagram-flame {
+  padding: 20px 20px 20px 20px;
+}
+</style>
--- a/packages/components/src/components/DiagramFlamegraph.vue
+++ b/packages/components/src/components/DiagramFlamegraph.vue
@@ -16,8 +16,8 @@ export default {
   name: 'v-diagram-flame',
 
   props: {
-    appMap: {
-      type: Object,
+    events: {
+      type: Array,
     },
   },
 
@@ -29,9 +29,7 @@ export default {
     const flamegraph = createFlamegraph();
     flamegraph.selfValue(false); // each node's value contains the sum of its children (as well as its own value)
     flamegraph.width(960);
-    d3.select(this.$refs.d3Container)
-      .datum(digestEventArray(this.appMap.rootEvents()))
-      .call(flamegraph);
+    d3.select(this.$refs.d3Container).datum(digestEventArray(this.events)).call(flamegraph);
   },
 };
 </script>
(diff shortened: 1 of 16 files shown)


packages/components/src/lib/flamegraph.js
```diff
@@ -0,0 +1,66 @@
...
+    );
+    return total;
+  } else {
+    return value;
+  }
+};
+
+const accumulateValue = (sum, { value }) => sum + value;
+
+const validateNode = ({ name, value, children }) => {
```
dustinbyrne: Maybe this is nit-picky, but a comment would be useful here for clarity. This isn't a "validator" in the sense that it's validating some set of criteria on the node, it's literally making the node valid by mutating/reducing the value.

packages/components/tests/unit/flamegraph.spec.js
```diff
@@ -0,0 +1,28 @@
...
+  expect(Number.isInteger(node.value)).toBe(true);
+  expect(node.value).toBeGreaterThanOrEqual(node.children.reduce(accumulateValue, 0));
+  expect(node.value);
+};
+
+describe('digestEventArray', () => {
+  it('produce valid flamegraph tree', () => {
+    const { events } = buildAppMap(scenario).normalize().build();
+    validateNode(digestEventArray(events));
+  });
```
dustinbyrne, line 27: This doesn't assert the flamegraph is valid, just that an exception is not raised. The data within `@/stories/data/scenario.json` should be stable enough to make assertions of specific values within the tree.

packages/components/src/components/DiagramFlamegraph.vue
```diff
@@ -0,0 +1,48 @@
...
+
+export default {
+  name: 'v-diagram-flame',
+
+  props: {
+    theme: {
+      type: String,
+      default: 'dark',
+      validator: (value) => ['dark', 'light'].includes(value),
+    },
```
dustinbyrne, line 23: `theme` isn't used - it can be removed

packages/components/src/components/DiagramFlamegraph.vue
```diff
@@ -0,0 +1,48 @@
...
+    appMap: {
+      type: Object,
+    },
```
dustinbyrne, line 26: I usually prefer to see components require the bare minimum properties. Otherwise we restrict the context in which we can add a flamegraph to other locations (i.e., we're saying here that we'd need an entire AppMap). In this case, `rootEvents` or even a pre-digested event array might be more suitable. I'll leave it up to your discretion.

dustinbyrne (changes requested):

packages/components/src/lib/flamegraph.js
```diff
@@ -0,0 +1,66 @@
...
+
+// TODO aggregate with codeObject.fqid
+
+const digestEvent = (event) => ({
+  name: String(event.toString()),
+  value: Number(event.elapsedTime),
+  children: event.children.map(digestEvent),
+});
+
+const validateValue = (name, total, value) => {
```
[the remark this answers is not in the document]
lachrist: Negative time are definitely a bug. But I don't want to panic and just return 0. I left the explicit `if` instead of `Math.max` to be able to easily log the case were a function's children take more time that the functions itself which also should not happen.

packages/components/src/lib/flamegraph.js
```diff
@@ -0,0 +1,66 @@
...
+const computeCommonFactor = (min) => {
+  let factor = 1;
+  while (min * factor < PRECISION) {
+    factor *= 10;
+  }
+  return factor;
+};
+
+const convertValueSample = (node) => {
+  const factor = computeCommonFactor(computeCommonFactor(Infinity, node));
```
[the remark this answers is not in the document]
lachrist: That was a bug the correct line is `computeCommonFactor(findMinValue(Infinity, node));`

packages/components/src/lib/flamegraph.js
```diff
@@ -0,0 +1,66 @@
...
+  const loop = ({ name, value, children }) => ({
+    name,
+    value: Math.round(value * factor),
```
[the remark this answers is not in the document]
lachrist: `flame-graph-d3` works with samples so it expects positive integers for node's value. It kinda works with float but not always. I prefer to normalised into integers. The factor is the minimal number to multiply all the float values before rounding them and retain precision defined by `PRECISION`.

packages/components/src/lib/flamegraph.js
lachrist: I rename it to `toCompatibleValue`.

packages/components/src/lib/flamegraph.js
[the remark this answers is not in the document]
lachrist: 0a8ef4a94c6a44a54cf0481ae781669948b2f56e

packages/components/src/lib/flamegraph.js
lachrist: 0a8ef4a94c6a44a54cf0481ae781669948b2f56e

packages/components/tests/unit/flamegraph.spec.js
lachrist: 0a8ef4a94c6a44a54cf0481ae781669948b2f56e

packages/components/src/components/DiagramFlamegraph.vue
lachrist: fixup 0e541a34b5dffae101cb8da3ec71c90806c2da8c

packages/components/src/components/DiagramFlamegraph.vue
lachrist: fixup: 36e774af53d38121844c9bd642c85138061999ea and a54ec1c2103bc84bcb427e2dd71e177d3a912590

kgilpin: But in order for the filters to work, isn't the `appMap` parameter required?

kgilpin: We do want to eventually support flame graphs for huge AppMaps, but not in v0.

lachrist: At the moment it is not required, so I think what Dustin said make sense. We can tweak the property later imo.

dustinbyrne (approved):

PR #1208, opened 2023-06-06
fix: update d3 dependencies to dedup d3-selection

lachrist:
Duplicate d3-selection causes issues. Let's use the most recent d3-selection version . As a bonus we get rid of the ugly global d3Event.

Fixes: https://github.com/getappmap/appmap-js/issues/1206

### DIFF
--- a/packages/diagrams/package.json
+++ b/packages/diagrams/package.json
@@ -30,10 +30,10 @@
   ],
   "dependencies": {
     "@appland/models": "workspace:^2.2.0",
-    "d3-interpolate": "^1.4.0",
-    "d3-selection": "^1.4.2",
-    "d3-shape": "^1.3.7",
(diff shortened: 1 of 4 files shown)


lachrist:
It would have been nice to add a test here but this issue only shows up in `vscode-appland`. The zoom worked fine in storybook. So it is the peer PR in `vscode-appland` that should test the zoom.

lachrist:
@dustinbyrne I'm not an expert at d3. Did you have a reason to stick with `d3-selection@1`?

dustinbyrne:
`diagrams` is just old. It hasn't seen active updates in a long time. I think it likely would be too much hassle to be worth our time to run the `components` tests within the extension, but we can definitely add a zoom test.

dustinbyrne (approved):

PR #1210, opened 2023-06-06
feat: Refine instructions for "Add AppMap to your project" and "Record AppMaps"

kgilpin:
* Provide an overview of install instructions for each language + ide.
* Provide manual instructions for all platforms except Java.
* Provide a Next button below each instruction set (automated and manual).
* Explain the advantages of the installer.
* Update Record AppMaps instructions and improve styling.


packages/components/src/components/install-guide/install-instructions/Python.vue
```diff
@@ -1,18 +1,35 @@
...
   <div>
     <section>
-      <h4>Installation via <code class="inline">pip</code></h4>
-      <v-code-snippet clipboard-text="pip install appmap" />
-    </section>
+      Manual installation is a good option if:
 
+      <ol>
+        <li>
+          Your project doesn't use <i>requirements.txt</i>, <i>Pipfile</i>, or <i>Poetry.lock</i>.
```
brikelly, line 8: I think the poetry one is usually all lowercase (i.e. `poetry.lock` rather than `Poetry.lock`)

packages/components/src/components/install-guide/install-instructions/Python.vue
```diff
@@ -1,18 +1,35 @@
...
+        <div class="center-block">or</div>
+        <v-code-snippet clipboard-text="poetry add --dev appmap" />
+      </div>
     </section>
 
     <section>
-      <h4>Installation via <code class="inline">pipenv</code></h4>
-      <v-code-snippet clipboard-text="pipenv install --dev appmap" />
+      <p>
+        <tt>appmap.yml</tt> will be created the first time you run your application with AppMap
```
brikelly, line 30: Should we preface lines like this with something like "A default configuration file, `appmap.yml`, will be created..."

brikelly (commented):
Just a few comments on the wording

packages/components/src/components/install-guide/ProjectPickerRow.vue
```diff
@@ -35,32 +35,104 @@
...
-        </div>
-        <template v-if="manualInstructions">
-          <div class="separator">OR</div>
-          <component :is="manualInstructions" data-cy="manual-install" />
+      <p class="mb20">
+        <template v-if="isJava">
+          Two things are required to make AppMaps of your Java project:
+
+          <ol>
+            <li>The <tt>appmap-agent.jar</tt> must be available on your machine.</li>
```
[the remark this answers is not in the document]
kgilpin: Yeah Java is weird in putting the version number in the file name. It makes it hard to update to new releases. So when we install I think we should just save the JAR file as appmap-agent.jar and drop the version numbers. That way the run configurations will continue to work and just pick up the latest. The JAR manifest has the version number (I think).

Daniel-Warner-X (approved):

PR #1212, opened 2023-06-07
Do not overwrite java home

### LINKED ISSUE #1211: Respect user-selected `JAVA_HOME` when running the CLI installer
Use java.jdt.ls.java.home setting, provided [here](https://github.com/redhat-developer/vscode-java), if available. Otherwise use JDK as specified by this preference order in [Setting the JDK](https://marketplace.visualstudio.com/items?itemName=redhat.java#setting-the-jdk). Otherwise prompt the user to perform manual install. 

There is also a bug where `JAVA_HOME` gets clobbered: [Demo](https://www.loom.com/share/dae63a19436f4c47898f1a2289498815)

ahtrotta:
Fixes #1211 

Removes unnecessary code that was run every time the CLI was executed, no matter what the command was. This code would sometimes overwrite the user's `JAVA_HOME` env variable. Also removes some unused imports.


packages/cli/src/cmds/agentInstaller/javaBuildToolInstaller.ts
```diff
@@ -5,29 +5,8 @@ import chalk from 'chalk';
...
-  }
-
-  const javaHome = path.join(jbHome, 'jbr');
-  const mvnBin = path.join(jbHome, 'plugins/maven/lib/maven3/bin');
-
-  // Make sure we don't override the user's settings: append to path, use the
-  // existing JAVA_HOME if it's set.
-  if (!process.env['JAVA_HOME']) {
-    process.env['JAVA_HOME'] = javaHome;
-  }
```
dividedmind, line 24: It's not clear to me how this could have overwritten JAVA_HOME. It seems pretty explicitly trying not to do that.

dividedmind:
> Fixes #1211

How so? It doesn't seem to do anything to use the Java home set in Visual Studio. Unless the extension ensures correct JAVA_HOME is set up in the terminals?

dustinbyrne:
Stale

PR #1222, opened 2023-06-09
feat: Record instructions styling

Daniel-Warner-X:
Adds styling for the updated Recording instructions.

Ruby
image
Java
image
Python
image


packages/components/src/components/install-guide/record-instructions/IntelliJ.vue
```diff
@@ -1,65 +1,84 @@
...
+            >recommended</span
+          >
+        </h3>
+        <p>
+          When you run your JUnit tests with the AppMap <code class="inline">javaagent</code> JVM
+          argument, an AppMap will be created for each test.
+        </p>
+        <p>
+          Right-click on any test class or package, and choose
+          <component :is="runConfigIcon" class="run-config-icon" /> "Start with AppMap".
```
brikelly, line 29: Might be too much detail, but given the prior sentence, we might need to say here "This will add the required JVM argument to enable AppMap recording."

packages/components/src/components/install-guide/record-instructions/IntelliJ.vue
```diff
@@ -1,65 +1,84 @@
...
+      <section class="recording-method">
+        <h3>
+          <i class="header-icon"><RemoteRecordingIcon /></i>Remote recording
+        </h3>
+        <p>
+          When your application uses {{ this.webFramework.name }}, and you run your application with
+          the AppMap <code class="inline">javaagent</code> JVM argument, remote recording is
+          enabled. To make a remote recording, run your application using
+          <component :is="runConfigIcon" class="run-config-icon" /> "Start with AppMap". Then use
+          the "Record" button to start recording. Interact with your application, through its user
```
brikelly, line 53: This will definitely need a gif of the record button, since that button is just a circle and doesn't even have a hover treatment that says "Record" :(

packages/components/src/components/install-guide/record-instructions/IntelliJ.vue
```diff
@@ -1,65 +1,84 @@
...
+          interface and/or by making API requests using a tool such as Postman. When you are done,
+          click the "Record" button again to stop the recording and view the AppMap.
```
brikelly, line 55: The button would be better named as the "Stop" button here, I think.

packages/components/src/components/install-guide/record-instructions/IntelliJ.vue
```diff
@@ -73,8 +92,11 @@
         for more information.
       </p>
     </section>
-    <section>
-      <h3>Runnable recording</h3>
+    <br />
+    <section class="recording-method">
+      <h3>
+        <i class="header-icon"><PlayIcon /></i>Runnable recording
```
brikelly, line 98: FYI in the docs this is termed "Code block recording" so we should get consistent on that term. https://appmap.io/docs/recording-methods.html#code-block-recording

EDIT: We actually have 3 different terms for this, because Java calls it Runnable Recording, and Python calls it Context Manager Recording: https://appmap.io/docs/reference/appmap-python.html#context-manager-recording

EDIT: and in Ruby we just call it "Block recording" so we have 4 ways to describe it. I think we should settle on "Code block recording" for now and I'll change the docs site to use just that term.

packages/components/src/components/install-guide/record-instructions/Python.vue
```diff
@@ -1,66 +1,78 @@
 <template>
   <section>
-    <p class="mb20 bold">
-      When you run your Python code with the <tt>appmap</tt> package, AppMap will be enabled for
-      recording. The best way to run with AppMap is to add the <tt>appmap</tt> package using Pip,
-      Pipenv, or Poetry.
+    <p>
+      When you run your Python code with the <code class="inline"> appmap </code> package, AppMap
+      will be enabled for recording. The best way to run with AppMap is to add the
+      <code class="inline">appmap</code> package using Pip, Pipenv, or Poetry.
```
brikelly, line 6: Maybe put the pip / pipenv / poetry words in code format?

packages/components/src/components/install-guide/record-instructions/Python.vue
```diff
@@ -1,66 +1,78 @@
...
+        <h3>
+          <i class="header-icon"><RequestsIcon /></i>Requests recording
+        </h3>
+        <p>
+          When your application uses {{ this.webFramework.name }}, and you run your application with
+          AppMap enabled, HTTP server requests recording is enabled. To record requests, first run
+          your application:
+        </p>
+        <p>
+          Start your Django server:
```
brikelly, line 49: Would be nice to only show the instructions for either Django or Flask, and not both. We already know which one it is from `this.webFramework.name`.

packages/components/src/components/install-guide/record-instructions/Python.vue
```diff
@@ -1,66 +1,78 @@
...
+          <v-code-snippet clipboard-text="python manage.py runserver" />
+        </p>
+        <p>
+          Start your Flask server:
+          <v-code-snippet clipboard-text="flask run" />
+        </p>
+        <p>
+          Interact with your application, through its user interface and/or by making API requests
+          using a tool such as Postman. An AppMap will be created for each HTTP server request
```
brikelly, line 58: Maybe "using a tool such as curl or Postman"? I just think curl is more ubiquitous.

packages/components/src/components/install-guide/record-instructions/Python.vue
```diff
@@ -1,66 +1,78 @@
...
+          that's served by your app.
```
brikelly, line 59: Drop the "that's" word, as it makes it sound like they'll get an AppMap for every registered API route. (at least to me!)

packages/components/src/components/install-guide/record-instructions/Ruby.vue
```diff
@@ -1,69 +1,91 @@
...
-      enabled for recording. By default, the <tt>appmap</tt> gem is only enabled in the "test" and
-      "development" environments.
+    <p>
+      When you run your Ruby code with the <code class="inline">appmap</code> gem in your bundle,
+      AppMap will be enabled for recording. By default, the <code class="inline">appmap</code> gem
+      is only enabled in the "test" and "development" environments.
     </p>
+    <br />
     <p>
       Before you run your app with AppMap, be sure and stop or disable Spring. You can stop Spring
```
brikelly, line 10: Change "be sure and stop" to just "stop"

packages/components/src/components/install-guide/record-instructions/Ruby.vue
```diff
@@ -1,69 +1,91 @@
...
+            >recommended</span
+          >
+        </h3>
+        <p>
+          When you run your {{ testFramework.name }} tests with AppMap enabled, an AppMap will be
+          created for each test.
+        </p>
+        <p>
+          <template v-if="testFramework.name.toLowerCase() == 'minitest'">
+            Run Minitest tests:
```
brikelly, line 30: Would be nice to also only show the instructions for the test framework they're using (if they're only using one of them).

packages/components/src/components/install-guide/record-instructions/Ruby.vue
```diff
@@ -1,69 +1,91 @@
...
-          >AppMap docs - Ruby Requests recording</a
-        >.
-      </p>
+      <section class="recording-method">
+        <h3>
+          <i class="header-icon"><RequestsIcon /></i>Requests recording
+        </h3>
+        <p>
+          When your application uses {{ this.webFramework.name }}, and you run your application with
+          AppMap enabled, HTTP server requests recording is enabled. To record requests, first run
```
brikelly, line 59: Change "is enabled" to "will be enabled". Might also need to be "server request recording..." (singular instead of plural).

packages/components/src/components/install-guide/record-instructions/Ruby.vue
```diff
@@ -1,69 +1,91 @@
...
+          your application:
+        </p>
+        <p>
+          Start your Rails server:
+          <v-code-snippet clipboard-text="DISABLE_SPRING=true rails server" />
+        </p>
+        <p>
+          Interact with your application, through its user interface and/or by making API requests
+          using a tool such as Postman. An AppMap will be created for each HTTP server request
+          that's served by your app.
```
brikelly, line 69: See comments on other similar area elsewhere in the PR

packages/components/src/components/install-guide/record-instructions/Ruby.vue
```diff
@@ -1,69 +1,91 @@
...
     </template>
     <template v-else>
       <VWebFrameworkPrompt frameworks="Rails" />
     </template>
-    <section>
-      <h3>Block recording</h3>
+    <br />
+    <section class="recording-method">
+      <h3>
+        <i class="header-icon"><CodeBlockIcon /></i>Block recording
```
brikelly, line 87: Change to "Code Block Recording"

brikelly (commented):
Comments given on mostly language & phrasing

packages/components/src/components/install-guide/record-instructions/IntelliJ.vue
Daniel-Warner-X: @brikelly before standardizing for consistencies sake, I'd like to understand why the different names were chosen. Does the distinct jargon make sense in the context of each language?

kgilpin: It’s named according to the convention for each language. That will feel natural to the user. I don’t think we need to unify that. We are trying to feel language and platform native to the user. Using the terminology of that language and platform is part of how we do that. 

To comment on the instructions you need to put yourself in the shoes of a developer of that language. 

Introducing a term for the purpose of consistency doesn’t help me if I am a developer. I care only about the language I’m using.

packages/components/src/components/install-guide/record-instructions/Ruby.vue
kgilpin: That’s what this section does.

packages/components/src/components/install-guide/record-instructions/Ruby.vue
kgilpin: Don’t though.

packages/components/src/components/install-guide/record-instructions/IntelliJ.vue
kgilpin: I think it’s useful to be explicit.

packages/components/src/components/install-guide/record-instructions/Ruby.vue
Daniel-Warner-X: Not sure what you are pointing out here. Could you be more specific?

packages/components/src/components/install-guide/record-instructions/Ruby.vue
Daniel-Warner-X: done

packages/components/src/components/install-guide/record-instructions/Ruby.vue
Daniel-Warner-X: done

kgilpin:
I welcome feedback to make the instructions better but I’m not reacting positively to the idea of removing language specific details like the JVM flag and language specific names of recoding methods. 

For example, we wouldn’t say “add our Ruby package”. We say “add our Ruby gem”. That is how instructions for users of that language and framework are written.

packages/components/src/components/install-guide/record-instructions/IntelliJ.vue
Daniel-Warner-X: I'm leaving it as 'record' because it is the same button. These instructions are clear to me... that said the UI design itself could be clearer.

packages/components/src/components/install-guide/record-instructions/IntelliJ.vue
Daniel-Warner-X: done

packages/components/src/components/install-guide/record-instructions/Ruby.vue
kgilpin: Should be triple equals though - I think.

packages/components/src/components/install-guide/record-instructions/Python.vue
kgilpin: Is it now though?

packages/components/src/components/install-guide/record-instructions/Ruby.vue
Daniel-Warner-X: reverted to 'Block recording'

packages/components/src/components/install-guide/record-instructions/Ruby.vue
brikelly: Oh I missed the `v-if`, thanks.

packages/components/src/components/install-guide/record-instructions/IntelliJ.vue
brikelly: I'm quite familiar with Java, and calling it Runnable Recording still doesn't make sense to me, even though the feature requires a `Runnable` entry point to record. Consistency aside, it doesn't convey the value or purpose of the feature. It's an implementation detail, but IMO we should be describing things in terms of value to the user so they can decide if they want to use it at all. 

> To comment on the instructions you need to put yourself in the shoes of a developer of that language.

Not every Java programmer is going to be super familiar with `Runnable` unless they are routinely creating threads of their own, which they might not be. Java app servers and frameworks have been trying to shield programmers from doing that kind of stuff for a long time. So if we call it `Runnable Recording` then it will be clear only to a subset of users, but if we call it `code block recording` or `code-driven recording` then 100% of them will understand what it's about.

kgilpin: I hear what you’re saying but I don’t think either of the terms code block recording or code driven recording really works either. 

“block” is a Ruby term.

“code driven” isn’t really clear as a definition - isn’t it all code driven?

Daniel-Warner-X: stet

packages/components/src/components/install-guide/install-instructions/Python.vue
```diff
@@ -1,18 +1,37 @@
...
+      <div>
+        <div class="center-block">or</div>
+        <v-code-snippet clipboard-text="poetry add --dev appmap" />
+      </div>
     </section>
 
     <section>
-      <h4>Installation via <code class="inline">pipenv</code></h4>
-      <v-code-snippet clipboard-text="pipenv install --dev appmap" />
+      <h3>Generate configuration file <i>appmap.yml</i></h3>
```
kgilpin, line 29: This section is a proposal, the Python program below doesn't actually exist yet.

packages/components/src/components/install-guide/record-instructions/IntelliJ.vue
Daniel-Warner-X: I added this screenshot... gif seemed like overkill. 
image

Daniel-Warner-X: @kgilpin heads up that I tweaked the language in the first sentence to describe 'run with AppMap' as a menu item instead of a button.

packages/components/src/components/install-guide/record-instructions/Python.vue
Daniel-Warner-X: removed

packages/components/src/components/install-guide/record-instructions/IntelliJ.vue
kgilpin: Note that for tests it’s different. You right click on the tests package - afaik

packages/components/src/components/install-guide/record-instructions/Python.vue
Daniel-Warner-X: @brikelly I'm leaving this as is for now because part of the intent of these instructions is to inform the user about all our supported frameworks in case they want to switch to another app that might be a better fit for demoing AppMap.

Daniel-Warner-X:
@brikelly @kgilpin all the comments have been addressed. This should be good to merge.

PR #1229, opened 2023-06-14
Replace octokat with octokit

dividedmind:
Octokat hasn't been updated in five years and it's using a dependency with a known security issue:

https://github.com/advisories/GHSA-r683-j2x4-v87g


ahtrotta (approved):
Looks fine and it worked fine when I tested it manually :+1:

PR #1234, opened 2023-06-19
Flamegraph in house rebuild

lachrist:
This PR is meant to put the flamegraph in the hand of the team so it can be iterated upon. Consequently, the flamegraph is hidden behind a configuration flag to not expose the lambda user.

Known area of improvements:

- Performance: The entire event tree is displayed at once. On my setup this is not an issue even on large Appmap. If performance becomes an issue, we can implement a minimal width required to instantiate components. That will cause to only instantiate components on the part of the flamegraph that is zoomed in. While this may seem like a no-brainer, it will make the flamegraph harder to animate because transitions are easier to add on kept-alive components.
- Animation: animations can be improved. First we should not use `transition: all` for performance reasons. Second the rendering of flamegraph from 0px to 1px can cause the animation to be bumpy.
- Styling: this can certainly be tweaked to be made more beautiful.

I think it will be valuable to first focus on bigger design decisions before spending time on the points above.

### DIFF
--- a/packages/components/rollup.config.js
+++ b/packages/components/rollup.config.js
@@ -14,6 +14,7 @@ const argv = minimist(process.argv.slice(2));
 const baseConfig = {
   input: 'src/index.js',
   external: Object.keys(pkg.dependencies),
+  inlineDynamicImports: true,
   plugins: {
     preVue: [
       alias({
--- a/packages/components/rollup.config.js
+++ b/packages/components/rollup.config.js
@@ -80,16 +80,6 @@ if (!argv.format || argv.format === 'es') {
       vue(baseConfig.plugins.vue),
       image({ exclude: ['**/*.svg'] }),
       commonjs(),
-
-      // The Trace component is expected to be loaded dynamically as an asynchronous component.
-      // However, because it's no longer a single module (it's bundled into a much larger module),
-      // we must retain the `__esModule` flag in order for Vue to properly perform dynamic loading.
-      //
-      // See the following source:
-      // https://github.com/vuejs/vue/blob/b51430f598b354ed60851bb62885539bd25de3d8/src/core/vdom/helpers/resolve-async-component.js#L18-L28
-      // This may go away after updating to Vue 3?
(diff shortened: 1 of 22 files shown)


lachrist:
Maybe we should add tests for the new `Slide` component which mimic the the zoom bar from `packages/diagrams` as it is not clearly listed as a flamegraph component.

packages/components/src/components/flamegraph/FlamegraphBranch.vue
```diff
@@ -0,0 +1,132 @@
...
+<script>
+import VFlamegraphRest from './FlamegraphRest.vue';
+import { isEventDurationValid, getEventDuration, add } from '../../lib/flamegraph';
+const getBudget = ({ budget }) => budget;
+const empty = new Set();
+export default {
+  name: 'v-flamegraph-branch',
+  emits: ['selectEvent'],
+  components: {
+    VFlamegraphNode: async () => (await import('./FlamegraphNode.vue')).default,
```
dustinbyrne, line 24: What's the reasoning behind the dynamic import here?

packages/components/rollup.config.js
```diff
@@ -80,16 +80,6 @@ if (!argv.format || argv.format === 'es') {
...
       commonjs(),
-
-      // The Trace component is expected to be loaded dynamically as an asynchronous component.
-      // However, because it's no longer a single module (it's bundled into a much larger module),
-      // we must retain the `__esModule` flag in order for Vue to properly perform dynamic loading.
-      //
-      // See the following source:
-      // https://github.com/vuejs/vue/blob/b51430f598b354ed60851bb62885539bd25de3d8/src/core/vdom/helpers/resolve-async-component.js#L18-L28
-      // This may go away after updating to Vue 3?
-      replace({ 'var Trace': 'var Trace = { __esModule: true, //' }),
```
dustinbyrne, line 91: I'm happy to see this go. The commit referenced here is four years old now and we've since upgraded dependencies many times. Assuming the distribution works in other environments like VS Code, then this looks good to me.

PR #1236, opened 2023-06-20
feat: Add modified date fields to Finding

kgilpin:
Add two fields to the Finding type:

* `scopeModifiedDate?: Date` Latest modified date of any project source file in the finding scope. May be undefined if the scope event does not contain any descendent that comes from a local source file - which is unlikely.
* `eventsModifiedDate?: Date` Latest modified date of the source file of any finding event or ancestor of a finding event. May be undefined if no finding events come from a local source file.

These dates can be used to estimate when a finding was introduced into the code base.

## Definitions

* _Latest modified date_ If the source file has a git status, use the date of the last git commit. Otherwise, use the file mtime.
* _Finding scope_ The `scope` event on the finding. For example, the HTTP server request in which the finding was detected.
* _source file_ File indicated by the code object location of an event.
* _project source file_ A source file located within the filesystem subtree defined by the _appmap.yml_

### DIFF
--- a/packages/scanner/src/lastGitOrFSModifiedDate.ts
+++ b/packages/scanner/src/lastGitOrFSModifiedDate.ts
@@ -1,0 +1,69 @@
+import { debug, warn } from 'console';
+import { verbose } from './rules/lib/util';
+import { exec } from 'child_process';
+import { stat } from 'fs/promises';
+
+const FileModifiedDate = new Map<string, Date>();
+let GitExists: boolean | undefined;
+
+export function resetCache(): void {
+  FileModifiedDate.clear();
+}
+
+export function isCached(file: string): boolean {
+  return FileModifiedDate.has(file);
+}
+
+function detectGitExists(): Promise<boolean> {
+  return new Promise((resolve) => {
+    exec('git --version', (err) => {
+      if (err && err.code && err.code > 0) resolve(false);
+
+      resolve(true);
+    });
+  });
+}
+
+export async function gitExists(): Promise<boolean> {
+  if (GitExists === undefined) {
+    GitExists = await detectGitExists();
+  }
+  return GitExists;
+}
+
+export function gitModifiedDate(file: string): Promise<Date | undefined> {
+  return new Promise<Date | undefined>((resolve) => {
+    exec(`git log -n 1 --pretty=format:%cI ${file}`, (err, stdout) => {
+      if (err && err.code && err.code > 0) resolve(undefined);
+      if (stdout.trim() === '') resolve(undefined);
+
+      resolve(new Date(stdout));
+    });
+  });
+}
+
+export async function fileModifiedDate(file: string): Promise<Date | undefined> {
+  try {
+    const stats = await stat(file);
+    return stats.mtime;
+  } catch (e) {
+    warn(e);
+  }
+}
+
+export default async function lastGitOrFSModifiedDate(file: string): Promise<Date | undefined> {
+  let result = FileModifiedDate.get(file);
+  if (result) {
+    if (verbose()) debug(`Using cached modified date for ${file}`);
+    return result.getTime() === 0 ? undefined : result;
+  }
+
+  if (verbose()) debug(`Computing modified date for ${file}`);
+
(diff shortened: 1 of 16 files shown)


packages/scanner/src/ruleChecker.ts
```diff
@@ -115,7 +130,27 @@ export default class RuleChecker {
...
+    let appmapConfigDir: string | undefined;
+    {
+      let searchDir: string | undefined = dirname(resolve(appMapFileName));
+      while (!appmapConfigDir) {
+        if (await fileExists(join(searchDir, 'appmap.yml'))) {
+          appmapConfigDir = searchDir;
+        } else {
+          if (dirname(searchDir) === searchDir) break;
+
+          searchDir = dirname(searchDir);
```
dividedmind, line 142: This will loop forever if it reaches the filesystem root.

packages/scanner/src/ruleChecker.ts
```diff
@@ -145,19 +180,87 @@ export default class RuleChecker {
...
+        ) as string[];
+        scopeModifiedDate = (
+          await Promise.all(
+            localScopeFiles.map(async (filePath) => {
+              const resolvedPath = await resolvePath(filePath);
+              if (!resolvedPath) return;
+
+              return lastGitOrFSModifiedDate(resolvedPath);
+            })
+          )
```
dividedmind, line 213: This should probably loop instead of using `Promise.all` to avoid potentially spawning dozens of `git` processes simultaneously.

packages/scanner/src/ruleChecker.ts
```diff
@@ -145,19 +180,87 @@ export default class RuleChecker {
...
+
+      const eventsModifiedDate = (
+        await Promise.all(
+          [...relatedEventFiles].map(async (filePath) => {
+            const resolvedPath = await resolvePath(filePath);
+            if (!resolvedPath) return;
+
+            return lastGitOrFSModifiedDate(resolvedPath);
+          })
+        )
```
dividedmind, line 257: This should also loop instead of using `Promise.all`.

packages/scanner/test/util.ts
```diff
@@ -20,7 +20,11 @@ const fixtureAppMapFileName = (file: string): string => {
 };
 
 const fixtureAppMap = async (file: string): Promise<AppMap> => {
-  const appMapBytes = await readFile(fixtureAppMapFileName(file), 'utf8');
+  let fileName: string;
+  if (await fileExists(file)) fileName = file;
+  else fileName = fixtureAppMapFileName(file);
+
+  const appMapBytes = await readFile(fileName, 'utf8');
```
dividedmind, line 27: It's not `Bytes` if you give it an encoding.

dividedmind (changes requested):
Looks ok, my only concern is the potential infinite loop when looking for appmap.yml. Even if we know it was there, it could have been gone in the meantime; we should detect if it reaches the fs root and bail.

packages/scanner/src/lastGitOrFSModifiedDate.ts
```diff
@@ -0,0 +1,89 @@
...
+    const stats = await stat(file);
+    return stats.mtime;
+  } catch (e) {
+    warn(e);
+  }
+}
+
+export async function isFileUnknownToGit(file: string): Promise<boolean> {
+  return new Promise<boolean>((resolve) => {
+    exec(`git ls-files -o --exclude-standard ${file}`, (err, stdout) => {
```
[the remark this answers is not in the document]
kgilpin: I'm definitely open to suggestions on the proper Git commands. Does this need to change?

packages/scanner/src/ruleChecker.ts
kgilpin: According to the Node docs `(dirname(searchDir) === searchDir)` will be true at the FS root, which is why I'm testing that above.

dividedmind: Ah, right, I missed that! Happens reading the code after a sleepless night, I'm so sorry!

PR #1238, opened 2023-06-21
Remove leftovers of deprecated findingsEnabled flag

dividedmind:
Cleanup after https://github.com/getappmap/vscode-appland/issues/699

### DIFF
--- a/packages/components/src/pages/install-guide/InvestigateFindings.vue
+++ b/packages/components/src/pages/install-guide/InvestigateFindings.vue
@@ -135,7 +135,6 @@ export default {
         maintainability: 0,
       }),
     },
-    findingsEnabled: Boolean,
     analysisEnabled: Boolean,
     userAuthenticated: Boolean,
   },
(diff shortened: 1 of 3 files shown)


packages/components/src/pages/install-guide/InvestigateFindings.vue
```diff
@@ -135,7 +135,6 @@ export default {
         maintainability: 0,
       }),
     },
-    findingsEnabled: Boolean,
     analysisEnabled: Boolean,
```
kgilpin, line 138: What's `analysisEnabled`?

ahtrotta: `findingsEnabled` is just the setting that users can check or uncheck, `analysisEnabled` indicates if the user is authenticated and has findings enabled.

ahtrotta: `analysisEnabled` comes from the `AnalysisManager` from [this code](https://github.com/getappmap/vscode-appland/blob/master/src/services/analysisManager.ts#L74-L101)

ahtrotta (approved):

PR #1252, opened 2023-07-05
External routes are yellow in flame graph

### LINKED ISSUE #1249: External HTTP calls in the Flame Graph should be colored yellow
The top frame here should be yellow:

Image

This would match what it is on the other views:

Sequence Diagram:
Image

Dependency Map:
Image

ahtrotta:
Fixes #1249 

Before:

image

After: 

image


dustinbyrne (approved):
Looks good. This might also be a good candidate for a unit test.

PR #1256, opened 2023-07-05
Do not test on node 14

### LINKED ISSUE #1255: Stop testing on Node 14 in CI

ahtrotta:
Fixes #1255 

We can stop supporting Node 14 and so we can stop testing with it in CI to save time and resources.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,11 +27,6 @@ jobs:
 
     - node_js: 16
 
-    - node_js: 14
-      # appmapping doesn't work on node 14
-      # https://github.com/getappmap/appmap-agent-js/issues/229
-      script: yarn test:no-appmap
-
     - os: windows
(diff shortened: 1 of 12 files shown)


dividedmind (changes requested):
Could you also remove `test:no-appmap` altogether? It was added just for node 14. I think you can pretty much just revert 1898f113abe237447d1c2f6a0ee3acae686fc0cf to do that (except for https://github.com/getappmap/appmap-js/commit/1898f113abe237447d1c2f6a0ee3acae686fc0cf#diff-0d43c70cbc6a9d4bc14487388a77cd45ce912f5fb30e8aba690c1fa6f9ab3afcR7 — this line should be kept as it is now, I'm not sure what happened there). Thanks!

dividedmind (approved):

PR #1261, opened 2023-07-07
Highlight findings events for N+1 queries

### LINKED ISSUE #1251: Events selected in the search bar should also be highlighted in the flame graph
This would be especially helpful for analysis findings related to performance. For example, this map has an N+1 query with 60 occurrences. Showing those in the Trace View as a set of IDs to cycle through is OK, but seeing them all highlighted right in the flame graph would help the developer see the scale of the problem and help them fix it. 

Image

We might not want to use the typical "selected event" border color, because this color would indicate something different, namely which events are germane to an analysis finding, of which there might be many.

Also, it would be great if an N+1 query finding in the side panel (or perhaps any performance finding) could open the flame graph with the events pre-selected. For example, see where my mouse clicks in this gif. We could have a "Show in Flame Graph" that would reveal all the N queries, highlighted with a specific color.
Image

ahtrotta:
Fixes #1251


packages/components/src/components/DetailsPanel.vue
```diff
@@ -185,6 +193,21 @@ export default {
...
     },
+    analysisFindingSelection(findingObject) {
+      const finding = findingObject.resolvedFinding && findingObject.resolvedFinding.finding;
+      if (!finding) return;
+
+      if (finding.checkId === 'n-plus-one-query') {
+        this.$store.commit(SET_HIGHLIGHTED_EVENTS, finding.relatedEvents);
+        this.$store.commit(SET_VIEW, VIEW_FLAMEGRAPH);
+
+        const eventToFocus = finding.relatedEvents.find((relatedEvent) => !relatedEvent.sql_query);
```
kgilpin, line 204: Every finding can have relatedEvents. So does this need to be hard coded to N + 1 query?

packages/components/src/components/DetailsSearch.vue
```diff
@@ -197,6 +204,24 @@ export default {
...
       }
+
+      if (type === 'analysis-finding') this.$emit('analysisFindingSelection', object);
+    },
+
+    analysisFindingSelection(findingObject) {
+      const finding = findingObject.resolvedFinding && findingObject.resolvedFinding.finding;
+      if (!finding) return;
+
+      if (finding.checkId === 'n-plus-one-query') {
```
kgilpin, line 215: Is this just for prototype purposes? Hard coding diagram behavior for specific findings is probably not necessary if we just base the behavior on the Finding data / type.

packages/components/src/components/DiagramFlamegraph.vue
```diff
@@ -42,10 +41,6 @@ export default {
       type: Array,
       required: true,
     },
-    selectedEvents: {
-      type: Array,
```
kgilpin, line 46: The diagram is already being passed an array of events to select / highlight / focus (the terminology used in the diagram is very confusing - I couldn’t quite understand it). We may not need to create a new highlight mechanism here. We can just use the list that’s already passed in. @dustinbyrne

packages/components/src/components/DetailsPanel.vue
ahtrotta: @dustinbyrne @brikelly and I met on Friday and thought that N+1 query findings would be a good place to start improving how we display findings in AppMaps. We figured that the optimal way to display a finding in the map would be different for different types of findings. There will probably be overlap, though. For example, we talked about how other types of performance findings might be displayed in a similar way to N+1 query findings. But for now this seems like a good place to start making changes.

packages/components/src/components/DiagramFlamegraph.vue
ahtrotta: The selected object is what is displayed in the sidebar, and it gets centered and highlighted in some way in the diagram by default. 

The focused object is not displayed in the sidebar, but it gets centered and highlighted. The little eyeball icon uses this.

Both of these highlight a single event/code object, whereas I'm trying to highlight multiple events in a distinct way with the new highlight mechanism.

packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -1209,6 +1214,23 @@ export default {
     updateFilters(updatedFilters) {
       this.$store.commit(SET_SAVED_FILTERS, updatedFilters);
     },
+
+    analysisFindingSelection(findingObject) {
+      const finding = findingObject.resolvedFinding && findingObject.resolvedFinding.finding;
+      if (!finding) return;
+
+      if (finding.checkId === 'n-plus-one-query') {
```
dustinbyrne, line 1222: I'd avoid the special case and just enable it across the board.

packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -1209,6 +1214,23 @@ export default {
...
+        this.$store.commit(SET_HIGHLIGHTED_EVENTS, finding.relatedEvents);
+        this.$store.commit(SET_VIEW, VIEW_FLAMEGRAPH);
```
dustinbyrne, line 1224: If it's a `Performance` finding go to the flame graph, otherwise go to the trace view.

packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -1209,6 +1214,23 @@ export default {
...
+
+        const eventToFocus =
+          finding.participatingEvents && finding.participatingEvents.commonAncestor;
```
dustinbyrne, line 1227: The nullish operator works in this project now.

I'd also fall back to the finding event if `commonAncestor` isn't present.

```
const eventToFocus = finding.participatingEvents?.commonAncestor || finding.event;
```

packages/components/src/components/flamegraph/FlamegraphMain.vue
```diff
@@ -51,11 +53,6 @@ export default {
       type: Array,
       required: true,
     },
-    focusEvent: {
-      type: Object,
-      default: null,
-      validator: (value) => value === null || value instanceof Event,
-    },
```
dustinbyrne, line 58: I'd keep the prop drilling intact and pass this in as far down as from the `v-vs-code-extension` element. I've been slowly pulling out our usage of the store as it makes this kind of thing more difficult to embed in other contexts.

packages/components/src/components/DiagramFlamegraph.vue
dustinbyrne: See my other comment regarding prop drilling

dustinbyrne (changes requested):
Looks good. I tested it out locally with some of the changes suggested and it feels like a big improvement when viewing a finding.

packages/components/src/pages/VsCodeExtension.vue
ahtrotta: Done

packages/components/src/pages/VsCodeExtension.vue
ahtrotta: Good idea :+1:

packages/components/src/components/flamegraph/FlamegraphMain.vue
ahtrotta: As we discussed, I'll keep using the state, but we can re-evaluate later if needed.

packages/components/src/pages/VsCodeExtension.vue
dustinbyrne: Let's try it

dustinbyrne (approved):

PR #1265, opened 2023-07-11
fix: select correct code object

### LINKED ISSUE #1264: The code object tree opens an appmap with the code object selected
Currently in the VS Code extension, when using the Code Objects tree view, if you click on a query or HTTP request, the correct map is opened, but the relevant code object is not selected.

ahtrotta:
Fixes #1264 

An example: 
`fqid` sent from VS Code: `route:GET /`
`fqid` in Vue : `route:HTTP server requests->GET /`

Another example:
`fqid` from VS Code: `query:DELETE FROM "active_storage_attachments" WHERE "active_storage_attachments"."id" = ?`
`fqid` in Vue: `query:Database->DELETE FROM "active_storage_attachments" WHERE "active_storage_attachments"."id" = ?`

As you can see, the strings are not exact, so this PR changes the condition from an exact match to a conditional using `includes`.

### DIFF
--- a/packages/components/src/pages/VsCodeExtension.vue
+++ b/packages/components/src/pages/VsCodeExtension.vue
@@ -980,7 +980,9 @@ export default {
                 }
               }
             } else {
-              selectedObject = classMap.codeObjects.find((obj) => obj.fqid === fqid);
+              selectedObject = classMap.codeObjects.find(
+                (obj) => obj.fqid.includes(type) && obj.fqid.includes(object)
+              );
             }
 
             if (selectedObject) {


dustinbyrne:
Shouldn't the FQID be the same regardless of where it's used? Why does it matter whether you're in Vue or VS Code? I'm still looking into this.

kgilpin:
> Shouldn't the FQID be the same regardless of where it's used? Why does it matter whether you're in Vue or VS Code? I'm still looking into this.

I agree thanks for saying this. fqid is definitive. If it’s not working then I think that’s a bug somewhere but we don’t need to match different “forms” of fqid.

ahtrotta:
Okay, I'll make the `fqid`s consistent

PR #1271, opened 2023-07-12
Consistent `fqid`s for queries and (external) routes

### LINKED ISSUE #1269: `fqid` is consistent in VS Code and `@appland/models`
Currenlty, the `fqid` generated in VS Code and `@appland/models` is inconsistent, which is causing problems when comparing them.

For example, when selecting an HTTP request or SQL query from the Code Objects tree in VS Code, the correct AppMap gets opened, but the code object is not selected in the AppMap. This PR fixes that.

An example:
`fqid` sent from VS Code: `route:GET /`
`fqid` in Vue : `route:HTTP server requests->GET /`

ahtrotta:
Fixes #1269 

We do not need to include the parent name for queries and routes because the relationship between a query/route and its parent is always the same and can be inferred from the type.

For more information, see discussion [here](https://github.com/getappmap/appmap-js/pull/1265).

### DIFF
--- a/packages/components/tests/e2e/specs/appmap/componentDiagram.spec.js
+++ b/packages/components/tests/e2e/specs/appmap/componentDiagram.spec.js
@@ -156,10 +156,7 @@ context('AppMap component diagram', () => {
 
       // Go back to the component diagram
       cy.get('.tabs .tab-btn').first().click();
-      cy.get(`.node[data-id="HTTP server requests->GET /admin/orders"]`).should(
-        'have.class',
-        'highlight'
-      );
+      cy.get(`.node[data-id="GET /admin/orders"]`).should('have.class', 'highlight');
 
       // Go back to Trace view and select event with exceptions
       cy.get('.tabs .tab-btn').contains('Trace View').click();
--- a/packages/models/src/codeObjectId.js
+++ b/packages/models/src/codeObjectId.js
@@ -1,8 +1,19 @@
 import { CodeObjectType } from './codeObjectType';
 
+function isExcludedParentType(type) {
+  return [CodeObjectType.HTTP, CodeObjectType.DATABASE, CodeObjectType.EXTERNAL_SERVICE].includes(
+    type
+  );
+}
+
 export default function codeObjectId(codeObject, tokens = []) {
-  if (codeObject.parent) {
-    codeObjectId(codeObject.parent, tokens);
+  const { parent } = codeObject;
+
+  // If it's a route, query, or external service we don't need to include the parent name
+  //  because it's always the same ('HTTP server requests' for route and 'Database' for queries).
+  // This mirrors the VS Code implementation.
+  if (parent && !isExcludedParentType(parent.type)) {
+    codeObjectId(parent, tokens);
 
     let separator;
     switch (codeObject.parent.type) {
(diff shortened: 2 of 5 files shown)


packages/models/src/codeObjectId.js
```diff
@@ -1,8 +1,12 @@
 import { CodeObjectType } from './codeObjectType';
 
 export default function codeObjectId(codeObject, tokens = []) {
-  if (codeObject.parent) {
-    codeObjectId(codeObject.parent, tokens);
+  const { parent } = codeObject;
+
+  // If it's a route or query we don't need to include the parent name because it's always the same
+  // ('HTTP server requests' for route and 'Database' for queries). This mirrors the VS Code implementation.
+  if (parent && parent.type !== CodeObjectType.HTTP && parent.type !== CodeObjectType.DATABASE) {
```
dustinbyrne, line 8: External services as well. The most accurate definition would be any parent that's not a package or class.

packages/models/src/codeObjectId.js
```diff
@@ -1,8 +1,12 @@
 import { CodeObjectType } from './codeObjectType';
 
 export default function codeObjectId(codeObject, tokens = []) {
-  if (codeObject.parent) {
-    codeObjectId(codeObject.parent, tokens);
+  const { parent } = codeObject;
+
+  // If it's a route or query we don't need to include the parent name because it's always the same
+  // ('HTTP server requests' for route and 'Database' for queries). This mirrors the VS Code implementation.
```
dustinbyrne, line 7: Can you add a unit test to assert these are as expected?

dustinbyrne (changes requested):

packages/models/src/codeObjectId.js
ahtrotta: I tried changing to definition to be any parent that's not a package or class and it made [this unit test](https://github.com/getappmap/appmap-js/blob/main/packages/models/tests/unit/codeObjectId.spec.js#L85-L95) fail, so I specifically checked for queries, routes, and external services.

packages/models/src/codeObjectId.js
ahtrotta: Yup, good idea :+1:

packages/models/src/codeObjectId.js
dustinbyrne: I'd argue the assertion made in that unit test is invalid and the VS Code implementation is also wrong by including anything separated by `->`. I don't believe those separators were intended to make their way into the FQID.

In any case, this implementation is at least consistent with VS Code now.

packages/components/tests/e2e/specs/appmap/componentDiagram.spec.js
```diff
@@ -156,10 +156,7 @@ context('AppMap component diagram', () => {
 
       // Go back to the component diagram
       cy.get('.tabs .tab-btn').first().click();
-      cy.get(`.node[data-id="HTTP server requests->GET /admin/orders"]`).should(
-        'have.class',
-        'highlight'
-      );
+      cy.get(`.node[data-id="GET /admin/orders"]`).should('have.class', 'highlight');
```
dustinbyrne, line 159: I don't think this is important at the moment, but these attributes should probably switch to use FQID in the future.

dustinbyrne (approved):

PR #1281, opened 2023-07-20
fix: Close SQL warning file handle after use

kgilpin:


kgilpin:
This problem (process hanging) was due to git status hanging open. It’s been resolved by other means.

PR #1164, opened 2023-04-26
feat: Analysis CLI commands - archive, restore, compare

kgilpin:

PR #1286, opened 2023-07-24
Archive restore update compare commands

ahtrotta:
This is a copy of #1164, but I've re-worked the git commits to make it easier to review.


kgilpin:
Aside from the `update` command, this code has been merged via other PRs.

PR #1287, opened 2023-07-25
fix: Deserialize filter

kgilpin:


packages/models/src/serialize.js
```diff
@@ -92,8 +92,16 @@ export function deserializeFilter(filterState) {
...
+  for (const property in filterState) {
+    if (
+      Object.prototype.hasOwnProperty.call(filterState, property) &&
+      filterState[property] === undefined
+    )
+      delete filterState[property];
+  }
+
+  if ('rootObjects' in filterState && filterState.rootObjects !== false) {
+    filter.rootObjects = filterState.rootObjects;
```
ahtrotta, line 104: Changing it from `filter.declutter.rootObjects` to `filter.rootObjects` is going to break some other code that's looking for `rootObjects` in the `declutter` property. It looks like that's why the build failed. Should we just leave it as `filter.declutter.rootObjects`?

ahtrotta: Oh true, I had assumed that the test assertions were reflective of actual code, but that wasn't a good assumption.

kgilpin:
So we should probably update the Components for this as well?

ahtrotta (approved):
It turns out that the problem was that `serializeFilter` was looking for `rootObjects` in the `declutter` property, so I fixed it.

PR #1290, opened 2023-07-26
Feat check missing function

lachrist:
This will help prevent regression of: https://github.com/getappmap/appmap-agent-js/issues/219

### DIFF
--- a/packages/validate/lib/index.js
+++ b/packages/validate/lib/index.js
@@ -29,7 +29,7 @@ const versions = new Map([
 
 const keys = Array.from(versions.keys());
 
-const makeDesignator = JSON.stringify;
+const digest = JSON.stringify;
 
 /* c8 ignore start */
 const hasOwnProperty =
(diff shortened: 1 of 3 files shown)


packages/validate/lib/index.js
```diff
@@ -190,6 +168,25 @@ const updateEventArray = (events, updates) => {
   );
 };
 
+const normalizeVersion = (version) => {
+  if (/^[0-9]$/.test(version)) {
+    return `${version}.0.0`;
+  } else if (/^[0-9]+\.[0-9]$/.test(version)) {
+    return `${version}.0`;
+  } else if (/^[0-9]+\.[0-9]+\.[0-9]+$/.test(version)) {
```
dustinbyrne, line 176: The first regex doesn't allow multiple digits as in the major version.
The second regex doesn't allow multiple digits in the minor version.

packages/validate/lib/index.js
```diff
@@ -29,7 +29,7 @@ const versions = new Map([
 
 const keys = Array.from(versions.keys());
 
-const makeDesignator = JSON.stringify;
+const digest = JSON.stringify;
```
dustinbyrne, line 32: This is nit-picky, but why not indicate that this is actually `JSON.stringify` and not some other implementation? Anybody will be familiar with `JSON.stringify` and not `digest`. The alias makes this a little harder to read.

packages/validate/lib/index.js
```diff
@@ -83,29 +83,15 @@ const checkSchema = (data, options) => {
...
-      //   entity
-      // );
-      const designator = makeDesignator([
-        parent.name,
-        path2,
-        parseInt(lineno),
-        entity.static,
-        entity.name,
-      ]);
+      const designator = digest([entity.name, path2, parseInt(lineno), entity.static]);
```
dustinbyrne, line 94: Is there a `path1`? :)

dustinbyrne (approved):

packages/validate/lib/index.js
lachrist: rebased

packages/validate/lib/index.js
lachrist: 1bf8410d38e289fd8de9d043aa10622f7fa847c6

packages/validate/lib/index.js
lachrist: rebased

PR #1291, opened 2023-07-26
feat: Compare filter

kgilpin:
Configure filters that trim noise from the canonical sequence diagrams used for CI analysis.

Filters can be configured in two ways:

1) Via a `compare.filter` option in appmap.yml
2) Via `--filter` arguments to the `appmap archive` command

The filter type is defined as:

```typescript
export type HideOption =
  | 'pragma'
  | 'savepoint'
  | 'pg_metadata'
  | 'sqlite_metadata'
  | 'selenium'
  | 'ruby_included';

// This type is similar to FilterState in @appland/models, but only contains the
// fields that are likely to be used for creating a canonical sequence diagram.
export type CompareFilter = {
  hide_external?: boolean;
  dependency_folders?: Array<string>;
  hide_name?: Array<string>;
  hide?: Array<HideOption>;
  reveal?: Array<HideOption>;
};
```

These options are related to the AppMapFilter and FilterState types used in the diagram filter UI. They are different in a couple of ways:

1) Not all FilterState options are available in CompareFilter - for example, FilterState provides `hide_elapsed_time_under` and CompareFilter doesn't.
2) CompareFilter provides some named "hide options" - pragma, savepoint, etc. Each of these is a symbolic name assigned to hide_name regular expressions which are built into the analyzer. At this time, those filters are:

```typescript
export const HIDE_OPTIONS: Record<HideOption, RegExp[]> = {
  pragma: [/^query:PRAGMA\b/, /^query:[\s\S]*\bPRAGMA\b/],
  savepoint: [/^query:SAVEPOINT\b/, /^query:[\s\S]*\bSAVEPOINT\b/],
  selenium: [/^external-route:.*\bhttp:\/\/127\.0\.0\.1:\d+\/session\/[a-f0-9]{32,}\//],
  pg_metadata: [/^query:[\s\S]*\bpg_attribute\b/],
  sqlite_metadata: [/^query:[\s\S]*\bsqlite_master\b/],
  ruby_included: [/^function:.*\.included$/],
};
```

Note that **all** hide options are enabled by default. To disable a hide option, either:

a) Pass a `reveal` option. The indicated hide option will be remove the corresponding regular expressions from the hide_name list.
b) Pass a `hide` option. The indicated hide option(s) will be the only one(s) applied. 

Both `reveal` and `hide` option can be repeated, as can `hide_name` and `dependency_folders`.

`hide_name` is a more specific way to add hide name filters, as compared to `hide` and `reveal`.

`dependency_folders` is, by default `[ 'vendor', 'node_modules' ]`. It's used to configure local directories in CI that contain bundle external dependencies. It's unlikely that it will need to be changed, but it's available in any case.

### DIFF
--- a/packages/cli/src/lib/loadAppMapConfig.ts
+++ b/packages/cli/src/lib/loadAppMapConfig.ts
@@ -1,9 +1,26 @@
-import { FilterState } from '@appland/models';
 import { readFile } from 'fs/promises';
 import { load } from 'js-yaml';
 
+export type HideOption =
+  | 'pragma'
+  | 'savepoint'
+  | 'pg_metadata'
+  | 'sqlite_metadata'
+  | 'selenium'
+  | 'ruby_included';
+
+// This type is similar to FilterState in @appland/models, but only contains the
+// fields that are likely to be used for creating a canonical sequence diagram.
+export type CompareFilter = {
+  hide_external?: boolean;
+  dependency_folders?: Array<string>;
+  hide_name?: Array<string>;
+  hide?: Array<HideOption>;
+  reveal?: Array<HideOption>;
+};
+
 export interface CompareConfig {
-  filter?: FilterState;
+  filter?: CompareFilter;
 }
 
 export interface UpdateConfig {
--- a/packages/cli/jest.config.js
+++ b/packages/cli/jest.config.js
@@ -5,4 +5,7 @@ module.exports = {
   testTimeout: parseInt(process.env.TEST_TIMEOUT, 10) || 5000,
   silent: true,
   restoreMocks: true,
+  // There are test cases that change the process working directory, and that does
+  // not work with multiple workers.
+  maxWorkers: 1,
 };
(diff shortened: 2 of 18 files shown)


packages/cli/src/lib/loadAppMapConfig.ts
```diff
@@ -1,9 +1,26 @@
...
 import { readFile } from 'fs/promises';
 import { load } from 'js-yaml';
 
+export type HideOption =
+  | 'pragma'
+  | 'savepoint'
+  | 'pg_metadata'
+  | 'sqlite_metadata'
+  | 'selenium'
+  | 'ruby_included';
```
dustinbyrne, line 10: If these are meant to be tightly coupled consider the following:
```ts
export const HIDE_OPTIONS = {
  pragma: [/^query:PRAGMA\b/, /^query:[\s\S]*\bPRAGMA\b/],
  savepoint: [/^query:SAVEPOINT\b/, /^query:[\s\S]*\bSAVEPOINT\b/],
  selenium: [/^external-route:.*\bhttp:\/\/127\.0\.0\.1:\d+\/session\/[a-f0-9]{32,}\//],
  pg_metadata: [/^query:[\s\S]*\bpg_attribute\b/],
  sqlite_metadata: [/^query:[\s\S]*\bsqlite_master\b/],
  ruby_included: [/^function:.*\.included$/],
} as const;

export type HideOption = keyof typeof HIDE_OPTIONS;
```

dustinbyrne (approved):

PR #1297, opened 2023-07-28
scrollfix-flamegraph

typaulhus:
adds fix for flame graph scrolling to IDE extension, so fixes will not be on [app.land only](https://github.com/getappmap/appmap-server/pull/1261)

Before:
Screenshot 2023-07-28 at 3 48 38 PM

After:
Screenshot 2023-07-28 at 3 47 31 PM

### DIFF
--- a/packages/components/src/components/flamegraph/FlamegraphMain.vue
+++ b/packages/components/src/components/flamegraph/FlamegraphMain.vue
@@ -237,13 +237,22 @@ export default {
 </script>
 
 <style lang="scss">
+.github-artifact {
+  background-color: #000;
+
+  #app {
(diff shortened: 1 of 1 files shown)


packages/components/src/components/flamegraph/FlamegraphMain.vue
```diff
@@ -237,13 +237,22 @@ export default {
 </script>
 
 <style lang="scss">
+.github-artifact {
```
dustinbyrne, line 240: This concept doesn't exist in this repo, only in app.land

typaulhus:
styles added here arent needed for the IDE, closing the PR

PR #1302, opened 2023-08-04
feat: `appmap index` aggregates event usage

dustinbyrne:
This change adds:
- Client interfaces to the usage APIs (`POST v1/usage`, `GET /v1/usage`)
- `appmap index` now uses the usage API to notify event/appmap/contributor usage
- Fixes an issue where `appmap index --watch` may have exited before finishing the `EventAggregator` callback due to flushing on the process `exit` event.
- Removes some timers from CLI tests which were causing tests to hang until they resolved


kgilpin:
It’s also important to metric the non-watch index. This is used in CI eg as a precondition to analysis.

kgilpin:
So is an API key required to run the indexer now?

packages/client/src/usage.ts
```diff
@@ -0,0 +1,51 @@
...
+    const responseJson = JSON.parse(response.body.toString('utf-8')) as UsageReport | ErrorResponse;
+    if (!response.ok) {
+      if ('message' in responseJson) throw new Error(responseJson.message);
+      throw new Error(`Got unexpected status code ${response.statusCode}`);
+    }
+
+    return responseJson as UsageReport;
+  }
+
+  static async update(events: number, orgId?: number): Promise<void> {
```
kgilpin, line 39: I don’t follow org id here. The user may belong to multiple orgs. What we do know is the app name as reported by appmap.yml. Also the account identified by the api key.

packages/client/test/usage.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+        .query({})
+        .matchHeader('authorization', `Bearer ${ApiKey}`)
+        .reply(200, usageReport);
+      return expect(Usage.get()).resolves.toEqual(usageReport);
+    });
+
+    it('issues an expected API call when an org identifier is provided', () => {
+      const orgId = 1;
+      nock('http://localhost:3000')
+        .get('/v1/orgs/usage')
```
kgilpin, line 50: Query to this data will be restricted by org id so I’d expect that to be a path parameter?

packages/client/src/usage.ts
```diff
@@ -0,0 +1,51 @@
...
+    await makeRequest({
+      service: ServiceEndpoint.ServiceApi,
+      path: this.usagePath,
+      method: 'POST',
+      body: JSON.stringify({
+        orgId,
```
kgilpin, line 45: How does the client know which org id to send?

dustinbyrne:
At the moment this will silently fail if the API returns a non-2xx response. So with that being said, an API key is not yet required. We can turn that on when we're ready.

packages/cli/src/fingerprint/fingerprintWatchCommand.ts
```diff
@@ -32,15 +35,20 @@ export default class FingerprintWatchCommand {
...
-    new EventAggregator((events) => {
+    this.eventAggregator = new EventAggregator(async (events) => {
       const indexEvents = events.map(({ args: [event] }) => event) as FingerprintEvent[];
-      this.pendingTelemetry = this.sendTelemetry(
-        indexEvents.map(({ metadata }) => metadata),
-        directory
+      const { metadata, numEvents } = indexEvents.reduce(
+        (acc, { metadata, numEvents }) => ({
+          metadata: [...acc.metadata, metadata],
+          numEvents: acc.numEvents + numEvents,
```
kgilpin, line 43: We may as well count AppMaps too.

packages/cli/src/lib/emitUsage.ts
```diff
@@ -0,0 +1,77 @@
+import { Usage } from '@appland/client';
+import { Metadata } from '@appland/models';
+import { Git } from '../telemetry';
+import type { UsageReport, UsageUpdateDto } from '@appland/client/dist/src/usage';
+import sanitizeURL from './repositoryInfo';
+
+function serializeMetadata(metadata?: Metadata): string | undefined {
+  if (!metadata) return;
```
kgilpin, line 8: How could this be falsey?

packages/cli/src/lib/emitUsage.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  if (!usageReport) {
+    // We don't have a cached usage report, so fetch one from the server
+    try {
+      usageReport = await Usage.get();
+    } catch (e) {
+      // Do nothing, we'll try again later.
+      // At the time being, we don't need to rate limit the client.
+      // The client will either only emit usage once per execution, or
+      // at a maximum, once per 10 seconds.
```
kgilpin, line 42: Log something? Completely swallowed exceptions freak me out.

packages/client/src/loadConfiguration.ts
```diff
@@ -1,15 +1,21 @@
...
 class Settings {
   baseURL = DefaultURL;
+  apiURL = DefaultApiURL;
   apiKey?: string;
 }
 
 function loadFromEnvironment(): Settings {
   const settings = new Settings();
 
+  ['APPLAND_API_URL', 'APPMAP_API_URL'].forEach((key) => {
```
kgilpin, line 15: Do we want new env vars for these? If `APPMAP_URL` is `getappmap.com` then `API_URL` can be `api.getappmap.com`

packages/cli/src/lib/emitUsage.ts
dustinbyrne: I added a message when running in verbose mode.

packages/cli/src/lib/emitUsage.ts
dustinbyrne: As it's implemented now it should never be undefined. This was more of a code style decision to satisfy the type checker when returning an array element. E.g.

As implemented I can:
```ts
serializeMetadata(metadataArray[0]);
```
vs. without:
```ts
if (metadataArray.length) serializeMetadata(metadataArray[0]);
```

After collecting and adding [git information to the metadata](https://github.com/getappmap/appmap-js/pull/1302/commits/64b4e4439f464c7065e987ff4668c274b8c19f89#diff-19be78c4bb538809c1cbfc6fa61ff897e0db6f959d0ec2bd791675fb740706c8R8-R25) this will now always return a value.

PR #1303, opened 2023-08-04
fix: New-root selection color

### LINKED ISSUE #1289: First item in 'add new root' select is highlighted but not selected
When a user clicks on the 'Add new root...' select box in the AppMap filter the first option has a 'selected' style but is not actually selected.

Image

(In the above screenshot `package:actionpack` is not actually selected).

Additionally, the *selected* state color is the same as the background color of the group box. This makes it hard to understand what is being selected. The *selected* state and *hover* states should be clearly different. The purple that is used in the rest of the group might be a good choice.

typaulhus:
When selecting a New root from the filter menu, the first item is selected. This selection color was incorrect, should be just white text to be consistent.

## Before:
Screenshot 2023-08-04 at 4 33 23 PM

## After: 
Screenshot 2023-08-04 at 4 29 20 PM

### DIFF
--- a/packages/telemetry/telemetry.spec.ts
+++ b/packages/telemetry/telemetry.spec.ts
@@ -196,14 +196,16 @@ describe('telemetry', () => {
     });
 
     describe('state', () => {
-      it('returns NotInstalled on error', () => {
-        jest.spyOn(child_process, 'spawn').mockImplementation(() => {
+      it('returns NotInstalled on error', async () => {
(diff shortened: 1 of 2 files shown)


dustinbyrne:
I rebased this and the issue was automatically closed and 'merged'. Apparently this was already shipped via another PR: https://github.com/getappmap/appmap-js/commit/04aab6ec71b29b6f7bfcab6040ea02a588986b8f

PR #1305, opened 2023-08-07
feat: Export scanner types and "scan" function

kgilpin:
Export index.ts from the scanner, so that the project can be imported by other projects.

This will enable us to scan from any other project (such as the CLI). It will also enable us to remove some nasty hacking of types in other projects, and redefinition of scanner types such as Findings in the CLI and Components.

The primary reason for this change is to allow the `analysis` CLI command to scan directly, rather than invoking through `npx`. That way it has more control over the scan, and can implement parallel scanning via worker threads.

Externally facing types are moved into [index.ts](https://github.com/getappmap/appmap-js/pull/1305/files#diff-e1bd45839d343ae9e2c0d432c1366078ed586ab3c33b2f2b10c8fb0423cfce75).

Implementation details are not exported through this file. A `scan` function is also exported.

See also: https://github.com/getappmap/appmap-js/pull/1306

### DIFF
--- a/packages/scanner/bin/front-matter.ts
+++ b/packages/scanner/bin/front-matter.ts
@@ -3,8 +3,8 @@
 import { readFile, writeFile } from 'fs/promises';
 import { dump } from 'js-yaml';
 
-import { Rule } from '../src/types';
 import { allRules } from './util';
+import RuleInstance from '../src/ruleInstance';
 
 function toTitleWord(word: string): string {
   if (['http', 'rpc', 'sql'].includes(word.toLowerCase())) {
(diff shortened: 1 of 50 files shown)


dustinbyrne (approved):
Looks good :+1:. This actually fixes some import weirdness in `vscode-appland` as well. In the longer term I think it's worth turning on `declarations` in tsconfig and moving away from type declarations in `index.ts`, but the need is not immediate.

kgilpin:
Whoops, I meant to squash the fix ups. Oh well.

PR #1306, opened 2023-08-07
feat: Analyze faster

### LINKED ISSUE #1292: CPU-intensive analysis operation are accelerated using multi-threading
On a particular data set:

```
Indexing AppMaps...
Indexed 310 AppMaps in 9469ms
Generating sequence diagrams...
Generated 310 sequence diagrams in 31448ms
Generating OpenAPI...
Generated OpenAPI in 5134ms
Scanning...
Scanned 310 AppMaps in 186645ms
```

Indexing is less than 10 seconds, so in this situation, it probably can't be accelerated further. OpenAPI generation is similar.

Sequence diagrams takes 30s. 

Scanning takes 186s, it's the slowest by a factor of 6.

kgilpin:
Archive command accepts a new flag:

* `thread-count [number]` Number of Node.js worker threads to use when analyzing AppMaps 

If this option is not provided, the thread count will equal [os.cpus().length](https://nodejs.org/api/os.html#oscpus). 

Worker threads are used to generate sequence diagrams and to perform scanning on the AppMaps. The analyzer will start out with 2 worker threads (or 1, if only one is requested), and add a new worker thread every 5 seconds until the max workers is reached or the analysis is completed.

## Review instructions

To review this, I might recommend reviewing these small commits:

* 452d04efa628a24e51712e83603a66d686a61677
* 460ffba4efe53731271b6fe1740328d5949dd8e1
* 720e42fe40f7f08d6de1891bae452f635f98d50e
* 68841b8cbc8a7eb4cffbe8b5cb73e09b830693b8

Then look at [analyzeWorker.ts](https://github.com/getappmap/appmap-js/pull/1306/files#diff-72ea99c627e6a9e55d60bb334042b46eba9e9ccc0f2032fe76c69c0d6869d75a) and [analyze.ts](https://github.com/getappmap/appmap-js/pull/1306/files#diff-7c0ce7e602d963bab7ee004b75473f191036edb0ef52274f5308d6479ef6b275).

## Sample data

Data from running `archive` on a sample (private) repo using a 2021 MacBook Pro with Apple M1 Max. OS reports 10 cpus, but this seems nowhere close to reality - at least for optimal scaling of the analysis. 

The 310 AppMaps are indexed already, so the index phase takes only 326ms. OpenAPI gen takes about 5 seconds.

| Worker threads | Sequence diagrams | Scanning | Total |
|--|--|--|--|
| 1 |  84s | 34s | 2:09 |
| 2 | 47s | 22s | 1:12 |
| 4 | 32s | 21s | 1:04 |
| 8 | 30s | 33s | 1:14 |
 

Fixes #1292


packages/scanner/src/appMapIndex.ts
```diff
@@ -3,7 +3,7 @@ import { QueryAST } from './types';
 import LRUCache from 'lru-cache';
 
 const NormalizedSQLBySQLString = new LRUCache<string, string>({ max: 10000 });
-const ASTBySQLString = new LRUCache<string, QueryAST>({ max: 1000 });
+const ASTBySQLString = new LRUCache<string, QueryAST | 'parse-error'>({ max: 1000 });
```
kgilpin, line 6: Because the LRUCache could contain `null` as a value, unparseable SQL would be parsed over and over.

packages/cli/src/cmds/archive/analyze.ts
```diff
@@ -1,42 +1,119 @@
...
+import { SequenceDiagramResult } from '../sequenceDiagram/SequenceDiagramResult';
+import { ScanResult } from '../scan/ScanResult';
+import { ScanTask } from '../scan/ScanTask';
+import { warn } from 'console';
+import { existsSync } from 'fs';
+import assert from 'assert';
+
+const WORKER_FILE = [
+  join(__dirname, 'analyzeWorker.js'),
+  join(__dirname, '../../../built/cmds/archive', 'analyzeWorker.js'),
```
kgilpin, line 20: In testing, it might be here.

packages/cli/src/cmds/archive/analyzeWorker.ts
```diff
@@ -0,0 +1,62 @@
...
+  if (task.verbose) verbose(task.verbose);
+
+  const writeIndexFile = async (data: string, fileName: string) => {
+    const indexDir = join(dirname(task.appmapFile), basename(task.appmapFile, '.appmap.json'));
+    await writeFile(join(indexDir, fileName), data);
+  };
+
+  try {
+    const stats = await stat(task.appmapFile);
+    if (stats.size > task.maxSize) {
```
kgilpin, line 32: It's handy to centralize oversize checking here.

packages/cli/src/cmds/archive/indexAppMaps.ts
```diff
@@ -15,7 +15,7 @@ export async function indexAppMaps(
   options.fileCountFn = counter.setCount();
   options.errorFn = reportAppMapProcessingError('Index');
   await processFiles(
-    '**/*.appmap.json',
+    '.appmap.json',
```
kgilpin, line 18: You'll see a bunch of this in the PR. We are using an old version of Glob. It's given me fits when the `appmap_dir` is `.`, as it is in some tests and may possibly be in the real world? The only feature of Glob we were using is recursive finding by file name or file extension, so I have just written that and refactored Glob out.

packages/cli/src/cmds/archive/analyze.ts
```diff
@@ -1,42 +1,119 @@
...
     const elapsed = new Date().getTime() - startTime;
-    console.log(`Scanned ${result} AppMaps in ${elapsed}ms`);
+    console.log(
+      `Scanned ${result.numProcessed} AppMaps in ${elapsed}ms, found ${findingsCount} findings`
+    );
+    for (const file of result.oversized) oversizedAppMaps.add(file);
+  };
+
+  assert(WORKER_FILE);
+  const workerPool = new WorkerPool(WORKER_FILE, threadCount);
```
kgilpin, line 110: By using one worker pool for both sequence diagrams and scanning, we save a bunch of time that would be consumed by starting the worker threads twice. It takes somewhere between 0.5 and 2 seconds to start a worker thread and process a job.

packages/cli/src/lib/workerPool.ts
```diff
@@ -0,0 +1,152 @@
...
+
+  constructor(
+    public taskFile: string | URL,
+    public numThreads: number,
+    public addWorkerDelay = DEFAULT_ADD_WORKER_DELAY
+  ) {
+    super();
+
+    this.addNewWorker();
+    if (numThreads > 1) this.addNewWorker();
```
kgilpin, line 55: We start one or two workers, then add another every 5 seconds. That way, if there aren't many jobs to process, we don't start a bunch of workers that we don't need.

packages/cli/src/lib/workerPool.ts
```diff
@@ -0,0 +1,150 @@
...
+    if (numThreads > 2) {
+      this.addWorkerInterval = setInterval(() => {
+        this.addNewWorker();
+
+        if (this.addWorkerInterval && this.workers.length === numThreads)
+          clearInterval(this.addWorkerInterval);
+      }, addWorkerDelay);
```
dustinbyrne, line 62: This should be [unref](https://nodejs.org/api/timers.html#immediateunref)'d, otherwise the process won't exit early if all jobs have finished processing early.

packages/cli/src/cmds/archive/analyzeWorker.ts
```diff
@@ -0,0 +1,62 @@
...
+      parentPort.postMessage({ oversized: true });
+      return;
+    }
+
+    let result: ScanResult | SequenceDiagramResult;
+    if (isScanTask(task)) {
+      const scanResults = await scan(task.appmapFile, 'appmap-scanner.yml');
+      writeIndexFile(JSON.stringify(scanResults, null, 2), 'appmap-findings.json');
```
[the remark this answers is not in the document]
dustinbyrne: Every disk access is going to be a potential point of contention, so I wouldn't be surprised if moving IO to the main thread resulted in a boost in performance. That's assuming we can read, transport and write that data faster than we can process it. We might also just be IO bound.

packages/cli/src/utils.ts
```diff
@@ -168,24 +160,61 @@ export async function processNamedFiles(
...
+  const traverseDirectory = async (
+    dir: string,
+    result = new Array<string>()
+  ): Promise<string[]> => {
+    let files: string[];
+    try {
+      files = await readdir(dir);
+    } catch (err) {
+      const code = (err as any).code;
+      if (code === 'ENOENT') return result;
```
dustinbyrne, line 192: We get a fair amount of `EPERM` in the wild as well. To resolve this, there are cases where we set the glob option `strict: false` to silently ignore permission errors. It doesn't look like any of those cases are changed in this commit though, so this is really more of an FYI.

dustinbyrne (approved):

PR #1310, opened 2023-08-10
Compare commands

kgilpin:
Does not include the `update` command, because it's not currently used in the preflight workflow.

The `compare` and `compare-report` commands are not tested on Windows since they are meant to be used in our GitHub Action, which runs on Linux.


packages/cli/src/cmds/compare-report/MarkdownReport.ts
```diff
@@ -106,6 +106,11 @@ export default class MarkdownReport implements Report {
         changeReport.apiDiff.breakingDifferences?.length || 0;
       changeReport.apiDiff.nonBreakingDifferenceCount =
         changeReport.apiDiff.nonBreakingDifferences?.length || 0;
+      Object.keys(changeReport.findingDiff).forEach((key) => {
+        changeReport.findingDiff[key].forEach(
+          (finding: Finding) => ((finding as any).appmap = finding.appMapFile)
+        );
+      });
```
ahtrotta, line 113: @kgilpin I'm not sure why you removed this code but it caused a test failure so I put it back in.

kgilpin:
It's in the wrong place; guarded by the wrong statement.

This is correct:

https://github.com/getappmap/appmap-js/pull/1316/files#diff-2999745164a31bff8495e19a264899f857961607c14734abe587484e58b5b9ccR128

ahtrotta (approved):

PR #1320, opened 2023-08-18
Drop OpenApi from the onboarding flow

### LINKED ISSUE #1319: Remove OpenAPI from onboarding flow

ahtrotta:
Fixes #1319


kgilpin:
What do you want to do with the “pie”? 

Should we add a step at the end to add our GitHub action?

kgilpin:
> > What do you want to do with the “pie”?
> > Should we add a step at the end to add our GitHub action?
> 
> I think we should remove OpenAPI and shrink the pie to 4 pieces, and then we can add something later if we want to as a follow up batch of work.

OK that's fine. I wasn't sure how to change the pie :-)

dustinbyrne (changes requested):
See the Chromatic feedback. It looks like the 'next' button isn't disappearing once all steps are complete. We should add a test case for that as well.

Other than that this is looking good.

dustinbyrne (approved):

kgilpin:
@ahtrotta looks like you can squash fixups and merge.

PR #1321, opened 2023-08-18
fix: `appmap compare` retains new AppMaps

dustinbyrne:
Previously new AppMaps were not being referenced during compare and thus marked for deletion. When they'd later appear in a report, the links would be broken.

### DIFF
--- a/packages/cli/tests/unit/compare/ChangeReporter.spec.ts
+++ b/packages/cli/tests/unit/compare/ChangeReporter.spec.ts
@@ -1,0 +1,46 @@
+import type { AppMapName } from '../../../src/cmds/compare/ChangeReport';
+import type { AppMapMetadata } from '../../../src/cmds/compare/ChangeReporter';
+import ChangeReporter, {
+  ChangeReportOptions,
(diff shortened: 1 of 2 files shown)


packages/cli/tests/unit/compare/ChangeReporter.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+        ]),
+      };
+      changeReporter.baseAppMaps = new Set<AppMapName>();
+      changeReporter.headAppMaps = new Set<AppMapName>(['example']);
+      changeReporter.failedAppMaps = new Set<AppMapName>();
+
+      const result = await changeReporter.report(new ChangeReportOptions());
+
+      expect(result.newAppMaps).toEqual(['example']);
+      expect(changeReporter.referencedAppMaps.test(RevisionName.Head, 'example')).toBeTruthy();
```
dustinbyrne, line 43: I've confirmed this fails on `main`

kgilpin (approved):
Thanks!

PR #1327, opened 2023-08-22
fix: Deduplicate findings in `change-report.json`

### LINKED ISSUE #1312: Findings have different hashes but look identical
From https://github.com/getappmap/appmap-server/pull/1268

Screen Shot 2023-08-11 at 12 12 54 PM

Investigate why these are considered separate findings.

dustinbyrne:
This removes findings with duplicate `hash_v2` properties. I considered doing the duplication in the markdown report, but I don't see any value in retaining the duplicated findings to begin with.

Fixes #1312.


kgilpin (approved):
Seems reasonable. I guess we can revert the related commit from the analyze-action?

dustinbyrne:
We can. It also doesn't hurt, especially if there's a chance for a version conflict given analyze-action doesn't explicitly set a version of the CLI to run with.

PR #1333, opened 2023-08-25
Scanner sometimes does not re-scan

ahtrotta:


kgilpin:
See https://github.com/getappmap/appmap-js/pull/1334

ahtrotta:
Fixed by #1334

PR #1339, opened 2023-08-31
feat: Sequence diagram diff in appmap

### LINKED ISSUE #1096: Diffs can be generated and viewed without PlantUML

kgilpin:
Fixes #1096 

Display sequence diagram diff in AppMap view. This feature is designed to be used to display the "diff" diagrams that are generated by AppMap Analysis in CI / GitHub Actions. 

When the AppMap diagram `loadData` method is called with a `sequenceDiagram` argument, that diagram will be used in place of the auto-generated sequence diagram that would otherwise be created. This "precomputed" sequence diagram provided can be a "diff" diagram. In this mode, the AppMap behavior differs from the default behavior in the following ways:

* The Filters functionality is disabled, because there is no way to change the sequence diagram to reflect the filters. 
* Diagram actions that do not contain any "diff" content are collapsed by default.
* The diagram scrolls to show the first "diff" action.

For an interactive example, see http://localhost:6006/?path=/story/pages-vs-code--extension-with-sequence-diff.

We will need to update the AppMap view in getappmap.com to fetch both the head AppMap and the sequence diagram diff, and construct the AppMap in accordance with these instructions. 

Related: https://github.com/getappmap/vscode-appland/pull/794

Screen Shot 2023-08-31 at 3 39 43 PM

### DIFF
--- a/packages/components/src/components/DetailsPanelFilters.vue
+++ b/packages/components/src/components/DetailsPanelFilters.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="details-panel-filters">
+  <div v-if="!isPrecomputedSequenceDiagram" class="details-panel-filters">
     <div class="details-panel-filters__head">
       <FilterIcon class="details-panel-filters__head-icon" />
       <span class="details-panel-filters__head-text">Filters</span>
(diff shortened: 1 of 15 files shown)


ahtrotta:
I opened this in storybook and then I clicked on `Extension with Sequence Diff` and it looked good, but then I navigated to `Extension with Default Sequence View` and then back to `Extension with Sequence Diff` and I got this error:

image

kgilpin:
This isn't a new problem as far as I know.

dustinbyrne:
Yeah there's an existing issue with state changes across stories. A page reload should resolve it.

dustinbyrne (approved):

PR #1341, opened 2023-08-31
Select finding when map opened from findings webview

### LINKED ISSUE #1340: Findings should open to the correct view

ahtrotta:
Fixes #1340 

Previously, when opening an AppMap from the findings webview:

image

It would open to the `Trace View`:

image

This change (along with a followup change in `vscode-appland`) makes it so that it opens the AppMap to the finding:

image

### DIFF
--- a/packages/components/src/pages/VsCodeExtension.vue
+++ b/packages/components/src/pages/VsCodeExtension.vue
@@ -357,6 +357,7 @@ import VStatsPanel from '../components/StatsPanel.vue';
 import VTabs from '../components/Tabs.vue';
 import VTab from '../components/Tab.vue';
 import VTraceFilter from '../components/trace/TraceFilter.vue';
+import toListItem from '@/lib/finding';
 import {
   store,
   SET_APPMAP_DATA,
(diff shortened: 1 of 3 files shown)


packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -979,6 +980,10 @@ export default {
                   });
                 }
               }
+            } else if (type === 'analysis-finding') {
+              const hash_v2 = object;
+              const finding = this.findings.filter(({ finding }) => finding.hash_v2 === hash_v2)[0];
```
dustinbyrne, line 985: Use `this.findings.find(...)` instead. It'll only look for the first match instead of iterating the entire list. Also, `toListItem` looks to expect that it's input is never undefined, so it'd be best to handle that gracefully as well.

dustinbyrne (changes requested):

dustinbyrne:
The data for this is readily available so please add a test for this as well.

packages/components/src/pages/VsCodeExtension.vue
ahtrotta: Yup, good points, I made these changes :+1:

dustinbyrne (approved):

PR #1345, opened 2023-09-02
feat: Sequence diagram validation

### LINKED ISSUE #1344: A helpful error message is returned when invalid files are passed to sequence-diagram-diff
A user attempts to diff two AppMaps, e.g.:
```
$ npx @appland/appmap@latest sequence-diagram-diff base.appmap.json head.appmap.json
```
A stacktrace is reported instead of a clear error message:
```
TypeError: Cannot read properties of undefined (reading 'forEach')
    at unparseDiagram (/Users/brikelly/.npm/_npx/93d874c56accaa6c/node_modules/@appland/sequence-diagram/dist/unparseDiagram.js:7:20)
    at readDiagramFile (/Users/brikelly/.npm/_npx/93d874c56accaa6c/node_modules/@appland/appmap/built/cmds/sequenceDiagram/readDiagramFile.js:8:50)
    at async compareFiles (/Users/brikelly/.npm/_npx/93d874c56accaa6c/node_modules/@appland/appmap/built/cmds/sequenceDiagramDiff.js:75:29)
    at async Object.handler (/Users/brikelly/.npm/_npx/93d874c56accaa6c/node_modules/@appland/appmap/built/cmds/sequenceDiagramDiff.js:131:9)
```
The desired behavior is that the user get a more helpful error message indicating that they should pass diagram files instead of AppMap files. Specifically, if we know that both files passed were both AppMap format, we should suggest they run the following commands to generate diagram files for each and then re-run the original command with the new diagrams:
```
appmap sequence-diagram -f json base.appmap.json
appmap sequence-diagram -f json head.appmap.json
appmap sequence-diagram-diff base.sequence.json head.sequence.json
```

This zip contains 2 AppMaps which can be used to test the current behavior, and to create diagram files if needed. [2-appmaps.zip](https://github.com/getappmap/appmap-js/files/12502764/2-appmaps.zip)

moxw:
Fixes #1344. This PR adds validation steps in the sequence diagram file reading and processing. It ensures that invalid or improperly formatted diagrams are flagged and logged, thus resolving the problem outlined in issue #1344.

**Changes**
Modified readDiagramFile to validate diagram files before parsing, returning a Result type that contains the diagram or validation outcome.
Updated sequenceDiagramDiff.ts to adapt to the new Result type, and added proper error messages for invalid or mismatched diagram types.
Added ValidationResult enum for categorizing validation results.
Created a new function validateDiagram for validating diagram data.


kgilpin:
@moxw this seems like a reasonable approach - thanks for your contribution! Please add a test for each of the three validation cases.

kgilpin:
PS if this is still a work in progress, you can indicate that using the Draft PR status.

moxw:
> @moxw this seems like a reasonable approach - thanks for your contribution! Please add a test for each of the three validation cases.

Have added the test cases.

packages/cli/src/cmds/sequenceDiagram/readDiagramFile.ts
```diff
@@ -1,7 +1,20 @@
...
-import { Diagram, unparseDiagram } from '@appland/sequence-diagram';
+import {
+  unparseDiagram,
+  validateDiagram,
+  ValidationResult,
+  Result,
+} from '@appland/sequence-diagram';
 
-export async function readDiagramFile(fileName: string): Promise<Diagram> {
+export async function readDiagramFile(fileName: string): Promise<Result> {
```
kgilpin, line 9: Thanks for creating the validate diagram function. We don’t want readDiagramFile to always validate first, because in some contexts the input is already known to be a valid diagram and we don’t want to spend the extra computation to validate it. 

So, readDiagramFile and validateDiagram can be two separate functions. 

The sequence diagram CLI can call validate followed by read. In other contexts, read can be called without validate. 

The CLI should have a command option to skip validation, but it should be enabled by default.

moxw: Addressed.

packages/cli/src/cmds/sequenceDiagramDiff.ts
```diff
@@ -50,6 +53,12 @@ export const builder = (args: yargs.Argv) => {
     describe: 'code objects to exclude from the diagram',
   });
 
+  args.option('skip-validation', {
```
kgilpin, line 56: This can be `validate` and when the option is `--no-validate`, the option will be `false`. I think this is called a `flag`, when the option does not require an argument because it's either off or on. The default of course will be `true`.

kgilpin (changes requested):

packages/cli/src/cmds/sequenceDiagramDiff.ts
moxw: Addressed

packages/sequence-diagram/src/validateDiagram.ts
```diff
@@ -0,0 +1,90 @@
...
+  isQuery,
+} from './types';
+
+export default async function validateDiagram(diagramData: any): Promise<ValidationResult> {
+  // Check if it's likely an AppMap object
+  if (
+    diagramData &&
+    diagramData.hasOwnProperty('metadata') &&
+    diagramData.hasOwnProperty('classMap') &&
+    diagramData.hasOwnProperty('events')
```
ahtrotta, line 19: If you run `yarn lint` from `packages/sequence-diagram`, you'll see that the linter throws an error. Instead of using `hasOwnProperty`, please just directly attempt to access the property (i.e. `diagramData.metadata`).

packages/sequence-diagram/src/validateDiagram.ts
```diff
@@ -0,0 +1,90 @@
...
+  Action,
+  isLoop,
+  NodeType,
+  isFunction,
+  isServerRPC,
+  isClientRPC,
+  isQuery,
+} from './types';
+
+export default async function validateDiagram(diagramData: any): Promise<ValidationResult> {
```
ahtrotta, line 13: Instead of the `any` type, please use `Diagram`.

packages/sequence-diagram/src/validateDiagram.ts
```diff
@@ -0,0 +1,90 @@
...
+  ) {
+    return ValidationResult.AppMap;
+  }
+
+  if (!Array.isArray(diagramData.actors) || !Array.isArray(diagramData.rootActions)) {
+    console.error('Invalid Diagram: actors and rootActions must be arrays.');
+    return ValidationResult.Invalid;
+  }
+
+  const validateActor = (actor: any): actor is Actor => {
```
ahtrotta, line 29: Instead of the `any` type please use `Actor`.

packages/sequence-diagram/src/validateDiagram.ts
```diff
@@ -0,0 +1,90 @@
...
+      typeof actor.name !== 'string' ||
+      typeof actor.order !== 'number'
+    ) {
+      console.error('Invalid Actor:', actor);
+      return false;
+    }
+    return true;
+  };
+
+  const validateAction = (action: any): action is Action => {
```
ahtrotta, line 41: Instead of the `any` type please use `Action`.

ahtrotta (changes requested):
This is looking good, thanks! I would like to see the changes I requested in the comments. Could you please move the fixtures from `fixtures/app/tmp/appmap` to `fixtures/sequenceDiagrams`?

Also, there are many different inputs for `validateDiagram` that will cause it to return `Invalid`. Can you add tests to cover more of these cases?

Specifically, add a test for each of the following:
* The input is an AppMap
* The input is missing `rootActions`
* One of the `actors` is not valid
* One of the `actions` is not valid
* One of the `actions` children is not valid

Could you also add tests to make sure that `validateDiagram` correctly returns `Valid` for `fixtures/sequenceDiagrams/listVsListWithPrefetch.sequence.json` and `fixtures/sequenceDiagrams/userFoundVsNotFound.sequence.json`?

packages/sequence-diagram/src/validateDiagram.ts
moxw: Directly trying to access leads to typescript errors. Will instead use `Object.prototype.hasOwnProperty.call(foo, "bar");`

moxw:
> This is looking good, thanks! I would like to see the changes I requested in the comments. Could you please move the fixtures from `fixtures/app/tmp/appmap` to `fixtures/sequenceDiagrams`?
> 
> Also, there are many different inputs for `validateDiagram` that will cause it to return `Invalid`. Can you add tests to cover more of these cases?
> 
> Specifically, add a test for each of the following:
> 
> * The input is an AppMap
> * The input is missing `rootActions`
> * One of the `actors` is not valid
> * One of the `actions` is not valid
> * One of the `actions` children is not valid
> 
> Could you also add tests to make sure that `validateDiagram` correctly returns `Valid` for `fixtures/sequenceDiagrams/listVsListWithPrefetch.sequence.json` and `fixtures/sequenceDiagrams/userFoundVsNotFound.sequence.json`?

Addressed.

kgilpin (approved):

ahtrotta (approved):
@moxw looks good! Thank you!

PR #1348, opened 2023-09-04
fix: Apply button in filter dialog is confusing

### LINKED ISSUE #1347: The "apply" button in the filter dialog is confusing
This [loom](https://www.loom.com/share/260df56d87d5448ab08a03041d82e08e?sid=8a37167b-e5c7-4ab4-8323-e87500a79124) demonstrates the issue. 

Users change some settings, then click "Apply" thinking it is needed to make those settings go into effect, but it actually reverts their changes back to the settings in the currently selected "saved filter" instead.

zermelo-wisen:
Fixes #1347. 

Changed FilterMenu.vue:
- apply button title renamed from "Apply" to "Load"
- moved the button to the right of the filter select
- spacing between "Load" button and select on the left made consistent with Save Filter button and the input box on its left.

Before (Filter Menu storybook)
Before

After (Filter Menu storybook)
After

### DIFF
--- a/packages/components/src/components/FilterMenu.vue
+++ b/packages/components/src/components/FilterMenu.vue
@@ -130,13 +130,13 @@
                 {{ savedFilter.filterName }}
               </option>
             </select>
+            <button class="filters__button" data-cy="apply-filter-button" @click="applyFilter">
+              Load
+            </button>
           </div>
         </div>
         <div class="filters__block-row">
           <div class="filters__block-row-content">
-            <button class="filters__button" data-cy="apply-filter-button" @click="applyFilter">
-              Apply
-            </button>
             <button :class="deleteButtonClass" data-cy="delete-filter-button" @click="deleteFilter">
               Delete
             </button>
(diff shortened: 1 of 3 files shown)


ahtrotta (changes requested):
Thanks for your PR! I like the change from `Apply` to `Load`, I think that's more clear. However, I think moving `Load` to the side of the select bar is a bit confusing because the `Delete`, `Set as Default`, and `Copy` buttons also apply to the selected filter. I'd like to see the following changes:

1. Move the `Load` button back to its original position.
2. Make the `Load` button disabled if the current filter state matches the selected saved filter state.
3. Please add a simple Cypress test in the `packages/components/tests/e2e/specs` folder.

zermelo-wisen:
- Made the changes above.
- Added a test file: `filterMenu.spec.js`.
- Fixed related tests in `appmap/viewFilter.spec.js` which are dependent on the order of disabled and enabled buttons, and "Load" (previously "Apply") button title.

ahtrotta (approved):
Looks good! Thank you!

PR #1349, opened 2023-09-04
Update FilterMenu for Comfortable Use

### LINKED ISSUE #1347: The "apply" button in the filter dialog is confusing

ultra1008:
Fixes #1347 
```
- "Apply" Button is disabled at first. When the "Saved Filter" is changed, then the "Apply" button becomes enabled.
- "Save Filter" Button and Input component is disabled at first. When the new filters are generated, then they become enabled.
```

### DIFF
--- a/packages/components/src/store/vsCode.js
+++ b/packages/components/src/store/vsCode.js
@@ -40,6 +40,14 @@ export const SET_HIGHLIGHTED_EVENTS = 'setHighlightedEvents';
 export const DEFAULT_VIEW = VIEW_COMPONENT;
 export const DEFAULT_FILTER_NAME = 'AppMap default';
 
+export const SET_LIMIT_ROOT = 'setLimitRoot';
+export const SET_HIDE_MEDIA = 'setHideMedia';
+export const SET_HIDE_EXTERNAL_CODE = 'setHideExternalCode';
+export const SET_HIDE_UNLABELED = 'setHideUnlabeled';
+export const SET_HIDE_ELAPSED_TIME_UNDER = 'setHideElapsedTimeUnder';
+export const SET_HIDE_NAME = 'setHideName';
+export const SET_CLEAR_FILTERS = 'setClearFilters';
+
 // Always have the AppMap default filter as the top option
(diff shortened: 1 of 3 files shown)


ultra1008:
```
- "Apply" Button is disabled at first. When the "Saved Filter" is changed, then the "Apply" button becomes enabled.
- "Save Filter" Button and Input component is disabled at first. When the new filters are generated, then they become enabled.
```

packages/components/src/store/vsCode.js
```diff
@@ -67,6 +75,13 @@ export function buildStore() {
       selectedSavedFilter: null,
       defaultFilter: null,
       highlightedEvents: [],
+
+      limitRoot: false,
+      hideMedia: false,
+      hideExternalCode: false,
+      hideUnlabeled: false,
+      hideElapsedTimeUnder: false,
+      hideName: false,
```
ahtrotta, line 84: Instead of creating these new state variables in the store, take a look at the `state.filters.declutter` property to keep track of the filter state.

ahtrotta (changes requested):
Thanks for your PR! I like your strategy for dealing with this problem. There are a couple of things I'd like to see changed, though:

1. Instead of adding several new properties to the store state, please look at `state.filters.declutter` to keep track of the filter state.
2. Please add a simple Cypress test in `packages/components/tests/e2e/specs`.
3. When a user changes the filter settings so that they are different from the currently selected saved filter (e.g. they click on `hide unlabeled`), the `Apply` button should become re-enabled.

packages/components/src/store/vsCode.js
ultra1008: Thanks for your review. I got what you mean and I will update the code asap.

Issue #1350, opened 2023-09-05
API changes in PR report contain extra confusing "paths." strings


brikelly:
Our API changes in PR reports look like this:
`paths./v1/usage`
But they should really look something like this:
`GET /v1/usage`

For example:
<img width="620" alt="Screenshot 2023-09-03 at 11 30 22 PM" src="https://github.com/getappmap/appmap-js/assets/573108/e9aea00c-7403-483e-a3a5-c8fcc5608d04">

Since we use [this library](https://www.npmjs.com/package/openapi-diff) for performing the OpenAPI diff, we might want to consider just not printing the `paths.` substring in the summary section to at least make it more readable.

kgilpin:
The format of that string is the sequence of keys in the openapi, so it’s
very predictable. We can reformat that string to match the format we want.

PR #1352, opened 2023-09-05
Improve paths in compare-report

moxw:
feat(ChangeReporter.ts): add import statements for DiffOutcome and DiffResult
feat(ChangeReporter.ts): add formatEndpoint and processDifferences methods
feat(ChangeReporter.ts): extend apiDiff logic for processing differences

Addresses #1350

### DIFF
--- a/packages/cli/src/cmds/compare/ChangeReporter.ts
+++ b/packages/cli/src/cmds/compare/ChangeReporter.ts
@@ -1,5 +1,11 @@
 import { mkdir, readFile, rm, writeFile } from 'fs/promises';
-import { default as openapiDiff } from 'openapi-diff';
+import {
+  DiffOutcomeFailure,
+  DiffOutcomeSuccess,
+  DiffResult,
+  DiffResultType,
+  default as openapiDiff,
+} from 'openapi-diff';
 import { dirname, isAbsolute, join, relative, resolve } from 'path';
 import { ClassMap, Metadata } from '@appland/models';
 import { FormatType, format } from '@appland/sequence-diagram';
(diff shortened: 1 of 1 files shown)


packages/cli/src/cmds/compare/ChangeReporter.ts
```diff
@@ -489,6 +495,36 @@ export class ReportFieldCalculator {
     };
   }
 
+  async formatEndpoint(input: string) {
+    if (input.startsWith('paths.')) {
+      const parts = input.split('.');
+
+      // Check if the input only has the 'paths.' and endpoint
+      if (parts.length === 2) {
+        return parts[1].replace('.', '/');
```
dustinbyrne, line 504: Hi @moxw, thank you for your PR! Out of curiosity, what's the condition in which you need to replace `.` with `/`?

moxw: Thank you for your review, @dustinbyrne. Upon further examination, the replacement of '.' with '/' was unnecessary in this context. I have removed it in the latest commit.

brikelly: @dustinbyrne do you have anything else you'd like addressed in this before we approve it?

dustinbyrne (approved):

packages/cli/src/cmds/compare/ChangeReporter.ts
dustinbyrne: Nope! This change looks to be well tested so I've approved 👍

PR #1355, opened 2023-09-06
fix: Reduce max memory usage of OpenAPI gen

kgilpin:
For OpenAPI generation, stage requests data into files organized by request path, then process one path at a time. This approach will greatly reduce the max memory usage.

Fixes https://github.com/getappmap/appmap-js/issues/1342

* [x] Create pre-release build of the CLI
* [x] Test pre-release on an existing build job that generates OpenAPI. Ensure that the diff report is the same.

Tested in https://github.com/land-of-apps/rails_tutorial_sample_app_7th_ed/pull/19/files#diff-a28198b36e9b27792ef763b3349cc505969aaff75e15e4bd5f43c3ced91117c3R37


kgilpin:
Verified as working correctly via manual testing with large repos.

PR #1358, opened 2023-09-07
feat: Emit run stats after index

dustinbyrne:
Following indexing, a JSON file is output to `./appmap/run-stats` (assuming `.appmap` exists). This file contains a sample AppMap metadata blob, the number of events and AppMaps processed and indicates whether the index occurred in CI.

Currently, this is not enabled in `--watch` mode, only during one shot indexing of a specific directory or via the `analyze` command.

### DIFF
--- a/packages/cli/src/lib/emitUsage.ts
+++ b/packages/cli/src/lib/emitUsage.ts
@@ -3,7 +3,8 @@ import { Metadata } from '@appland/models';
 import { Git } from '../telemetry';
 import type { UsageUpdateDto } from '@appland/client';
 import sanitizeURL from './repositoryInfo';
-import { verbose } from '../utils';
+import { mkdir, stat, writeFile } from 'node:fs/promises';
+import { join } from 'node:path';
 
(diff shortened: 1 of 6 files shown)


packages/cli/src/lib/emitUsage.ts
```diff
@@ -29,25 +30,27 @@ export default async function emitUsage(
...
-    }
-  });
+  try {
+    const stats = await stat('.appmap');
+    if (stats.isDirectory()) {
+      const runStatsDirectory = join('.appmap', 'run-stats');
+      await mkdir(runStatsDirectory, { recursive: true });
 
-  return dto;
+      const statsFilePath = join(runStatsDirectory, `${Date.now().toString()}.json`);
```
kgilpin, line 47: Is this time as number or time as string?

dustinbyrne: `Date.now()` returns the milliseconds since epoch, so a number. The `toString` is probably superfluous given this context.

kgilpin (approved):

PR #1359, opened 2023-09-07
Changing ~ to $HOME in instructions for location of Java agent

brikelly:
Raised by a user: using `~` causes problems with IntelliJ, so we will use $HOME instead.

### DIFF
--- a/packages/components/src/components/install-guide/ProjectPickerRow.vue
+++ b/packages/components/src/components/install-guide/ProjectPickerRow.vue
@@ -49,12 +49,12 @@
 
           <ol>
             <li>
-              The AppMap agent JAR (<code class="inline">~/.appmap/lib/java/appmap.jar</code>) must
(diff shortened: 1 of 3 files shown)


ahtrotta:
I'm assuming that the user was trying to modify their Run Config to add the VM argument themselves, right? If I modify the Run Config for `PetClinicApplication.java` to use `$HOME`:

image

I get this error:

image

Which is essentially the same error that I get when I use `~` instead of `$HOME`:

image

So either I'm doing something wrong (very possible!), or this doesn't fix the issue.

kgilpin:
I think what they mean is that when they copy this instruction to their
terminal, it doesn't work.

brikelly:
Maybe a better intent is to convey that they should expand $HOME and use that path explicitly. I guess we can expand it for them and just show the exact full path in these instructions, wdyt?

PR #1360, opened 2023-09-08
fix: Long call labels are cut off

### LINKED ISSUE #1063: Long call labels are cut off and can't be scrolled to in the sequence diagram
### Issue

When viewing a sequence diagram in an AppMap, if the call label is very long (as is often the case for a DB query), then it gets truncated:

Image

Seen using this map: 

[1676598908_493438_http_localhost_3000_cart_link.appmap.json.zip](https://github.com/getappmap/appmap-js/files/10815966/1676598908_493438_http_localhost_3000_cart_link.appmap.json.zip)

### Details

* Truncate the label (using a `...` at the end of the string to indicate truncation) when it goes past the arrow and add a hover state that shows the entire call label.
* Code for the call label is [here](https://github.com/getappmap/appmap-js/blob/main/packages/components/src/components/sequence/CallLabel.vue).
* Please make Cypress tests in `packages/components/tests/e2e/specs`.

zermelo-wisen:
Fixes #1063 

Before
image

After
image

### DIFF
--- a/packages/components/src/components/sequence/CallLabel.vue
+++ b/packages/components/src/components/sequence/CallLabel.vue
@@ -5,7 +5,35 @@
     </template>
     <div :class="nameClasses">
       <template v-for="text in name">
-        <span @click="selectEvent" :class="text.class" :key="text.text">{{ text.text }}</span>
+        <div
+          :key="text.text"
+          :style="{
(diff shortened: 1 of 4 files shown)


dustinbyrne (approved):
Looks great!

PR #1362, opened 2023-09-08
feat: Disable diagram tabs in 'diff' mode

kgilpin:
When showing a precomputed sequence diagram, disable the other tabs since they don't indicate any "diff". Also, disable the options to `Show in Trace`, `Show in Sequence`, `Show in Flame Graph`. This whole section of the LHS panel can be disabled.

In the future, it would be great to bring "diff" to the dependency map as well.

# Progress

When using a precomputed AppMap:

- [x] Tabs to switch between views are disabled
- [x] `Show in X` in the LHS panel is disabled

### DIFF
--- a/packages/components/src/components/DetailsButton.vue
+++ b/packages/components/src/components/DetailsButton.vue
@@ -43,7 +43,7 @@ export default {
   align-items: center;
   transition: $transition;
 
-  &:hover {
+  &:hover:not(&--disabled) {
     background-color: $gray4;
     border-color: $gray4;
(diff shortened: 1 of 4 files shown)


ahtrotta (dismissed):

kgilpin:
I've realized that we also need to disable `Show in Trace`, `Show in Flame Graph` so I'm going to move this back to Draft for now.

ahtrotta (approved):
Looks good, thanks!

Issue #1353, opened 2023-09-05
Hovering over a return value in a sequence diagram shows the value being returned


ahtrotta:
### Issue

When hovering over a return label in the sequence diagram, the value should be displayed. For example, the value of `array` should be displayed on hover:

![Image](https://user-images.githubusercontent.com/573108/220173746-3b30648e-08d5-406e-8afa-3fd94c9029a5.png)

### Details

* Add a hover state to the return label so that the return value is displayed on hover.
* The code for the return label is [here](https://github.com/getappmap/appmap-js/blob/main/packages/components/src/components/sequence/ReturnLabel.vue).
* Please add Cypress tests in `packages/components/tests/e2e/specs`.

kgilpin:
@ahtrotta @brikelly is this a duplicate of https://github.com/getappmap/appmap-js/pull/1360?

ahtrotta:
No, 1360 is about a hover state for the call label and this is about a hover state for the return label

PR #1363, opened 2023-09-08
Return value tooltip on hover

moxw:
Fixes: #1353 

feat(components/DiagramSequence.vue): add appMap prop to DiagramSequence
feat(sequence/ReturnAction.vue): integrate appMap prop to VReturnLabel instances
feat(sequence/ReturnLabel.vue): add tooltip to display return values or exceptions on hover
feat(sequenceDiagram.spec.js): add e2e test to verify tooltip visibility on hover

### DIFF
--- a/packages/components/src/components/DiagramSequence.vue
+++ b/packages/components/src/components/DiagramSequence.vue
@@ -35,6 +35,7 @@
               :actionSpec="action"
               :collapsed-actions="collapsedActions"
               :key="actionKey(action)"
+              :return-value="returnValue(action)"
             />
           </template>
         </template>
@@ -198,6 +199,26 @@ export default {
           selected.firstElementChild.scrollIntoView(SCROLL_OPTIONS);
         }
       }, 16);
+    },
+    returnValue(action: ActionSpec): string {
+      if (!action.eventIds || !this.appMap || !this.appMap.eventsById) {
+        return '';
+      }
+      const eventId = action.eventIds[0];
+      const event = this.appMap.eventsById[eventId];
+
+      if (!event) return '';
+
+      if (event.returnValue?.value) {
+        return event.returnValue.value;
+      }
+
+      if (event.exceptions?.length) {
+        const exception = event.exceptions[0];
+        return `${exception.class} ${exception.message}`;
+      }
+
+      return '';
     },
   },
 
--- a/packages/components/src/components/sequence/ReturnAction.vue
+++ b/packages/components/src/components/sequence/ReturnAction.vue
@@ -19,7 +19,7 @@
           },
         }"
       >
-        <VReturnLabel :action-spec="actionSpec" />
+        <VReturnLabel :action-spec="actionSpec" :return-value="returnValue" />
       </div>
     </template>
     <template v-else-if="actionSpec.callArrowDirection === 'right'">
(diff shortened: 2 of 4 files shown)


dustinbyrne:
I'm going to close and re-open this PR to trigger CI.

packages/components/src/components/sequence/ReturnLabel.vue
```diff
@@ -10,21 +11,53 @@ import { ActionSpec } from './ActionSpec';
...
     },
+
+    value() {
+      if (!this.actionSpec.eventIds || !this.appMap || !this.appMap.events) {
+        return false;
+      }
+      const eventId = this.actionSpec.eventIds[0];
+      const event = this.appMap.events.find(
+        (e) => e.parent_id && e.event == 'return' && e.parent_id === eventId
+      );
```
dustinbyrne, line 46: You can retrieve the event without iterating by accessing the event via the `eventsById` map. E.g.:
```ts
this.appMap.eventsById[eventId]
```
This will give you the call event. You can get the return value via the `returnValue` accessor. If you need the return event, `returnEvent` is also available.

packages/components/src/components/sequence/ReturnLabel.vue
```diff
@@ -10,21 +11,53 @@ import { ActionSpec } from './ActionSpec';
...
     actionSpec: {
       type: ActionSpec,
       required: true,
       readonly: true,
     },
+    appMap: {
+      type: Object,
+      required: true,
+      readonly: true,
+    },
```
dustinbyrne, line 30: We try to include as little data in each component as possible. I'd prefer it if we passed only the return value string, as that's the most we need to render this component.

This will mean you'll need to set up some sort of mapping in the parent component. Something as simple as a method to exchange an `ActionSpec` for a return value string will suffice.

packages/components/src/components/sequence/ReturnAction.vue
```diff
@@ -163,6 +163,11 @@ export default {
       type: Array,
       required: true,
     },
+    appMap: {
+      type: Object,
+      required: true,
+      readonly: true,
+    },
```
dustinbyrne, line 170: Same comment here regarding too much data passed to this component. A return value string would be ideal.

packages/components/src/components/DiagramSequence.vue
```diff
@@ -35,6 +35,7 @@
               :actionSpec="action"
               :collapsed-actions="collapsedActions"
               :key="actionKey(action)"
+              :appMap="appMap"
```
dustinbyrne, line 38: At this top level, passing an AppMap seems fine.

dustinbyrne (changes requested):
This is looking good! I've left a couple of comments.

packages/components/src/components/sequence/ReturnLabel.vue
moxw: Updated.

packages/components/src/components/sequence/ReturnLabel.vue
moxw: Updated.

packages/components/src/components/DiagramSequence.vue
moxw: Moved everything to parent.

packages/components/src/components/sequence/ReturnAction.vue
moxw: updated.

dustinbyrne (approved):
Looks great!

PR #1366, opened 2023-09-09
Stats command relative path support

### LINKED ISSUE #1200: Stats command should work with relative path
### Issue

Currently, the stats command only works with absolute file paths, like this:
```
npx @appland/appmap@latest stats --appmap-file /Users/kgilpin/source/land-of-apps/sample_app_6th_ed/tmp/appmap/minitest/Users_index_index_as_admin_including_pagination_and_delete_links.appmap.json
```

### Details

* The `stats` command code can be found [here](https://github.com/getappmap/appmap-js/blob/main/packages/cli/src/cmds/stats/stats.ts).
* The tests should be added to [this test file](https://github.com/getappmap/appmap-js/blob/main/packages/cli/tests/unit/stats/stats.spec.ts).
* The `stats` command should work with relative file paths.
* The `stats` command should work with file names (the command should just check the current directory).
* Please add a test to make sure that it works with a file name where the file is in the current working directory.
* Please add a test to make sure that it does not analyze a file when a file name is passed but the file is in a child directory.
* Please add a test to make sure it works with relative file paths.

moxw:
Fixes #1200 

refactor(cli/src/cmds/stats/stats.ts): simplify appmap file resolution logic
test(cli/tests/unit/stats/stats.spec.ts): adjust and add tests for appmap file resolution

### DIFF
--- a/packages/cli/src/cmds/stats/stats.ts
+++ b/packages/cli/src/cmds/stats/stats.ts
@@ -26,32 +26,19 @@ async function locateAppMapFile(userInput: string, appmapDir: string): Promise<s
   }
   const mapFile = userInput + extension;
 
-  const appmapDirIsCwd = (): boolean => path.resolve(appmapDir) === process.cwd();
-
   // if --appmap-file is a valid exact path, then return it
-  if (path.isAbsolute(mapFile) && existsSync(mapFile)) return mapFile;
+  // or resolve cwd and --appmap-file and check for an exact match
+  const absoluteUserInput = path.resolve(process.cwd(), mapFile);
+  if (existsSync(absoluteUserInput)) return absoluteUserInput;
 
-  // resolve --appmap-dir and --appmap-file and check for an exact match
-  const searchPath = appmapDirIsCwd() ? mapFile : path.join(appmapDir, mapFile);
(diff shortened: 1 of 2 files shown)


dustinbyrne (changes requested):
Thanks @moxw! The change looks great.

We use [`semantic-release`](https://semantic-release.gitbook.io/semantic-release/) to automatically version, update CHANGELOG entries and release the packages in this repo. Only `fix` or `feat` commits will trigger that to happen. Do you mind rebasing this to mark this change as a `fix`? No need to include a scope.

Otherwise, the code looks great. We can merge this after the rebase. :+1:

moxw:
> Thanks @moxw! The change looks great.
> 
> We use [`semantic-release`](https://semantic-release.gitbook.io/semantic-release/) to automatically version, update CHANGELOG entries and release the packages in this repo. Only `fix` or `feat` commits will trigger that to happen. Do you mind rebasing this to mark this change as a `fix`? No need to include a scope. If you're interested in reading more, we follow the conventional commits specification outlined here: https://conventionalcommits.org
> 
> Otherwise, the code looks great. We can merge this after the rebase. 👍

Done, let me know if this is good.

dustinbyrne (approved):
Perfect! Thank you.

Issue #1051, opened 2023-02-20
Long function call labels are obscured by lifelines in the sequence diagram


brikelly:
### Issue

When the call label is long and overlaps with an active lifeline (i.e. a thick lifeline), then the label is obscured by the lifeline.

![Image](https://user-images.githubusercontent.com/573108/220190724-3a99279a-004e-41d0-90bf-15595f987e32.png)

### Details

* [Here](https://github.com/getappmap/appmap-js/files/10787513/1676598908_493438_http_localhost_3000_cart_link.appmap.json.zip) is an AppMap that has this issue.
* The label should be in front of the lifeline, not behind it.
* The code for the label is [here](https://github.com/getappmap/appmap-js/blob/main/packages/components/src/components/sequence/CallLabel.vue).
* Please write Cypress tests in `packages/components/tests/e2e/specs`.

brikelly:
AppMap is attached, it's from Solidus

[1676598908_493438_http_localhost_3000_cart_link.appmap.json.zip](https://github.com/getappmap/appmap-js/files/10787513/1676598908_493438_http_localhost_3000_cart_link.appmap.json.zip)

PR #1367, opened 2023-09-09
Long function call labels on top of lifelines

moxw:
Fixes: #1051 
style(CallAction.vue): add position:relative to .connecting-span for better alignment
test(sequenceDiagram.spec.js): verify .call-line-segment.connecting-span has position relative

### DIFF
--- a/packages/components/src/components/sequence/CallAction.vue
+++ b/packages/components/src/components/sequence/CallAction.vue
@@ -441,6 +441,10 @@ export default {
   fill: $sequence-call-line-color;
 }
 
+.connecting-span {
+  position: relative;
+}
+
 .call-right {
   .single-span {
     left: calc(((var(--caller-lifecycle-depth)) * $sequence-activation-gutter-width));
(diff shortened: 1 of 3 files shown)


dustinbyrne (changes requested):
Thanks for this. Do you mind adding a transparent background to the labels? I think it helps clear up any confusion about which element is in front of the other. E.g., here's a `background-color` of `rgba(0,0,0,.7)` on [`.label .name`](https://github.com/getappmap/appmap-js/blob/main/packages/components/src/components/sequence/CallLabel.vue#L169)
image

It's just a small change I'd usually just add it myself but I can't push to your fork :smile:

moxw:
> Thanks for this. Do you mind adding a transparent background to the labels? I think it helps clear up any confusion about which element is in front of the other. E.g., here's a `background-color` of `rgba(0,0,0,.7)` on [`.label .name`](https://github.com/getappmap/appmap-js/blob/main/packages/components/src/components/sequence/CallLabel.vue#L169) image
> 
> It's just a small change I'd usually just add it myself but I can't push to your fork 😄

Done, added.

dustinbyrne (approved):
Perfect, thank you!

PR #1398, opened 2023-09-21
feat: Hide unused actors when collapsing actions

### LINKED ISSUE #1391: Hide unused Actors when collapsing Actions
The only Actors which should be displayed in the sequence diagram are those which are utilized by a displayed Action. 

When collapsing an Action, hide any Actors that are no longer referenced by a non-collapsed Action. 

Hiding behavior applies to:
- Actor packages
- Actor classes when a package has been expanded

When expanding an Action, add the child/to Actor if it was not previously referenced.

zermelo-wisen:
Fixes #1391

### DIFF
--- a/packages/components/src/components/DiagramSequence.vue
+++ b/packages/components/src/components/DiagramSequence.vue
@@ -163,7 +163,7 @@ export default {
       return this.collapsedActionState;
     },
     diagramSpec(): DiagramSpec {
-      this.collapsedActionState = this.collapsedActionState ?? [];
+      this.collapsedActionState = this.collapsedActionState ?? []; // eslint-disable-line vue/no-side-effects-in-computed-properties
 
       const result = new DiagramSpec(this.diagram);
       if (this.collapsedActionState.length == 0) {
--- a/packages/components/src/components/DiagramSequence.vue
+++ b/packages/components/src/components/DiagramSequence.vue
@@ -68,7 +68,7 @@ import VReturnAction from '@/components/sequence/ReturnAction.vue';
 import VActor from '@/components/sequence/Actor.vue';
 import DiagramSpec from './sequence/DiagramSpec';
 import { ActionSpec } from './sequence/ActionSpec';
-import { SET_COLLAPSED_ACTIONS, SET_FOCUSED_EVENT } from '../store/vsCode';
+import { SET_FOCUSED_EVENT } from '../store/vsCode';
 
 const SCROLL_OPTIONS = { behavior: 'smooth', block: 'center', inline: 'center' };
 
@@ -95,6 +95,12 @@ export default {
     selectedEvents: {
       type: Array,
     },
+  },
+
+  data() {
+    return {
+      collapsedActionState: [],
+    };
   },
 
   computed: {
@@ -154,13 +160,13 @@ export default {
       return result;
     },
     collapsedActions() {
-      return this.$store?.state.collapsedActions ?? [];
+      return this.collapsedActionState;
     },
     diagramSpec(): DiagramSpec {
-      let currentCollapsedActions = this.$store?.state.collapsedActions;
+      this.collapsedActionState = this.collapsedActionState ?? [];
 
       const result = new DiagramSpec(this.diagram);
-      if (currentCollapsedActions && currentCollapsedActions.length == 0) {
+      if (this.collapsedActionState.length == 0) {
         // If a Diagram contains any actions in diff mode, expand all ancestors of every diff action,
         // and collapse all other actions.
         const expandedActions = new Set<number>();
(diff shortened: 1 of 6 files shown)


kgilpin:
I'm getting this error in Storybook: http://localhost:6006/?path=/story/appland-diagrams-sequence--list-users

```
index.js:56 TypeError: Cannot read properties of undefined (reading 'commit')
    at VueComponent.collapseOrExpand (CallLabel.vue:164:1)
    at invokeWithErrorHandling (vue.common.dev.js:3636:1)
    at HTMLDivElement.invoker (vue.common.dev.js:1483:1)
    at original._wrapper (vue.common.dev.js:7931:1)
```

```javascript
  methods: {
    collapseOrExpand() {
      this.$set(this.collapsedActionState, this.actionSpec.index, !this.collapsed);
      this.$store.commit(SET_COLLAPSED_ACTIONS, this.collapsedActionState);
    },
```

Is there a way to do this without using the `store` and rather having an array of collapsed actions that's used by the UI elements?

kgilpin:
It's working nicely when it's in the 'VSCode' container - http://localhost:6006/?path=/story/pages-vs-code--extension-with-default-sequence-view&args=scenario:appland2

zermelo-wisen:
Done. I changed this to not use store (vuex state management).

By the way, there are some interactions which does not work on storybooks above, like:
- hiding an actor with (x) button gives an error.
- actors cannot be selected by clicking them.
I'm not sure how important these are, but it looks like they are all related to vuex store.

kgilpin:
Thanks! Yes there are some other features that don’t work without the store. But that is not to do with this PR so, not your problem :-)

packages/components/src/components/DiagramSequence.vue
```diff
@@ -97,6 +97,12 @@ export default {
     },
   },
 
+  data() {
+    return {
+      collapsedActionState: [],
```
kgilpin, line 102: For my edification - how does this component know to redraw when the set of collapsed actions changes?

zermelo-wisen: Though I'm not an expert in Vue, this is my understanding. 

For `collapsedActionState` to be reactive, it must be present in data function's return object. Since it's an array, settting elements like `this.collapsedActionState[5] = true`, without recreating the array reference, is not detected by Vue. Changes are detected - and a new render request is triggered - when we use `this.set$(collapsedActionState, 5, true)` instead.

packages/components/src/components/DiagramSequence.vue
```diff
@@ -153,52 +159,64 @@ export default {
       }
       return result;
     },
+    collapsedActions() {
+      return this.collapsedActionState;
+    },
     diagramSpec(): DiagramSpec {
-      const result = new DiagramSpec(this.diagram);
-      this.collapsedActions = []; // eslint-disable-line vue/no-side-effects-in-computed-properties
+      this.collapsedActionState = this.collapsedActionState ?? [];
```
ahtrotta, line 166: The [build](https://app.travis-ci.com/github/getappmap/appmap-js/jobs/610404554) is currently failing because the linter is throwing an error:

```
➤ YN0000: [@appland/components]: /home/travis/build/getappmap/appmap-js/packages/components/src/components/DiagramSequence.vue
➤ YN0000: [@appland/components]:   166:7  error  Unexpected side effect in "diagramSpec" computed property  vue/no-side-effects-in-computed-properties
➤ YN0000: [@appland/components]: 
➤ YN0000: [@appland/components]: ✖ 1 problem (1 error, 0 warnings)
➤ YN0000: [@appland/components]: 
➤ YN0000: [@appland/components]: Process exited (exit code 1), completed in 12s 328ms
```

It looks like you'll need to add `// eslint-disable-lin vue/no-side-effects-in-computed-properties` back to this line.

packages/components/src/components/sequence/DiagramSpec.ts
```diff
@@ -100,4 +102,30 @@ export default class DiagramSpec {
...
+        break;
+
+      // If non of the ancestors of this action is collapsed
+      // then this action is visible (not swallowed) and its actors
+      // need to be visible.
+      if (!actionSpec.isCollapsed(collapsedActions)) {
+        const [calle, caller] = actionActors(actionSpec.action);
+
+        if (calle && !result.has(calle))
+          result.add(calle);
```
ahtrotta, line 122: This is a bit nit-picky but it should be `callee` instead of `calle`. Can you change it just so that it's easier to read?

packages/components/src/components/sequence/DiagramSpec.ts
```diff
@@ -100,4 +102,30 @@ export default class DiagramSpec {
...
+  determineVisuallyReachableActors(collapsedActions: boolean[]) {
+    const result = new Set<Actor>();
+
+    for (const actionSpec of this.actions.filter(a => a.nodeType !== 'return')) {
+      // If all of the actors are visible at this point
+      // no need to check remaining actions.
+      if (result.size == this.actions.length)
+        break;
+
+      // If non of the ancestors of this action is collapsed
```
ahtrotta, line 115: Should be `If none of the ancestors of this action are collapsed`.

ahtrotta (changes requested):
This looks great! There are just a couple of small fixes I'd like to see before we merge. Thanks!

zermelo-wisen:
All done.

ahtrotta (approved):

PR #1400, opened 2023-09-21
feat: Diff once

kgilpin:
Don't run `git diff` for every appmap - run it only when a new set of root files is encountered in the classMap.

Parse the diff output to figure out what the diff is for each file.

## Reviewer's guide 🎉 

**How it used to work**

For every AppMap, if that AppMap contained source paths (most do...), [`git diff` was executed](https://github.com/getappmap/appmap-js/pull/1400/files#diff-c437d6ceecd503d4d812155930024411ffbdeba29439ec53f4efc2c9f26d88ecL97) to find the source diff for that AppMap.

When a PR contained a lot of changed AppMaps, this was a lot of work to perform.

**How it works now**

* For each AppMap [find the root source paths](https://github.com/getappmap/appmap-js/pull/1400/files#diff-c437d6ceecd503d4d812155930024411ffbdeba29439ec53f4efc2c9f26d88ecR95) (filename, if the path is not in a directory, or the top-level directory)
* Determine [if there are any new root paths](https://github.com/getappmap/appmap-js/pull/1400/files#diff-209a5b7d3452f008f594d316e58f77edc72a36f3624a4f9bfbea84e189848b9dR93) in the AppMap.
* If so, [load the diff for the new root paths](https://github.com/getappmap/appmap-js/pull/1400/files#diff-209a5b7d3452f008f594d316e58f77edc72a36f3624a4f9bfbea84e189848b9dR101).
* Parse the diff and [store the diff for each source file path](https://github.com/getappmap/appmap-js/pull/1400/files#diff-209a5b7d3452f008f594d316e58f77edc72a36f3624a4f9bfbea84e189848b9dR110).

Other notes:

* Only [one `diff` command is performed at a time](https://github.com/getappmap/appmap-js/pull/1400/files#diff-209a5b7d3452f008f594d316e58f77edc72a36f3624a4f9bfbea84e189848b9dR25). This just makes it easier to avoid race conditions as AppMaps load in parallel.
* The output is the [unified diff format](https://github.com/getappmap/appmap-js/pull/1400/files#diff-209a5b7d3452f008f594d316e58f77edc72a36f3624a4f9bfbea84e189848b9dR77); some of the information output by `git diff` is discarded; but I don't see how this will matter in practice.

### DIFF
--- a/packages/cli/src/cmds/compare/DiffLoader.ts
+++ b/packages/cli/src/cmds/compare/DiffLoader.ts
@@ -2,6 +2,7 @@ import parseDiff from 'parse-diff';
 import { executeCommand } from '../../lib/executeCommand';
 import { warn } from 'console';
 import assert from 'assert';
+import { exists } from '../../utils';
 
 export const MAX_DIFF_LENGTH = 1000 * 1000;
 
@@ -12,9 +13,30 @@ export default class DiffLoader {
 
   constructor(public baseRevision: string, public headRevision: string) {}
 
+  lookupDiff(path: string): string | undefined {
+    assert(this.diffByFile);
+
+    const diffFiles = this.diffByFile.get(path);
+    if (!diffFiles) return;
+
+    return diffFiles
+      .map((diffFile) =>
+        diffFile.chunks
+          .map((chunk) =>
+            [chunk.content, chunk.changes.map((change) => change.content).join('\n')].join('\n')
+          )
+          .join('\n')
+      )
+      .join('\n');
+  }
+
   async update(sourcePathRoots: Set<string>) {
     let reloadDiff = false;
     for (const root of sourcePathRoots) {
+      if (['vendor', 'node_modules'].includes(root)) continue;
+
+      if (!(await exists(root))) continue;
+
       if (!this.sourcePathRoots.has(root)) {
         this.sourcePathRoots.add(root);
         reloadDiff = true;
@@ -29,12 +51,19 @@ export default class DiffLoader {
   }
 
   private async reloadDiff() {
+    const diffCommands = new Set<string>();
+
     while (this.reloadQueue.length > 0) {
       const sourcePathRoots = this.reloadQueue.shift()!.sort();
+      const diffCommand = `git diff ${this.baseRevision}..${
+        this.headRevision
+      } -- ${sourcePathRoots.join(' ')}`;
+      if (diffCommands.has(diffCommand)) continue;
+
+      diffCommands.add(diffCommand);
+
       try {
-        const diffStr = await executeCommand(
-          `git diff ${this.baseRevision}..${this.headRevision} -- ${sourcePathRoots.join(' ')}}`
-        );
+        const diffStr = await executeCommand(diffCommand, true);
         if (diffStr.length > MAX_DIFF_LENGTH) {
           warn(`Diff is too large to parse. Skipping...`);
           continue;
@@ -42,11 +71,14 @@ export default class DiffLoader {
 
         const diffFiles = parseDiff(diffStr);
         this.diffByFile = diffFiles.reduce((memo, diffFile) => {
-          [diffFile.to, diffFile.from].filter(Boolean).forEach((path) => {
-            assert(path);
-            if (!memo.has(path)) memo.set(path, []);
-            memo.get(path)?.push(diffFile);
-          });
+          [...new Set([diffFile.to, diffFile.from])]
+            .filter(Boolean)
+            .sort()
+            .forEach((path) => {
+              assert(path);
+              if (!memo.has(path)) memo.set(path, []);
+              memo.get(path)?.push(diffFile);
+            });
           return memo;
         }, new Map<string, parseDiff.File[]>());
       } catch (err) {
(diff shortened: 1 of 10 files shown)


packages/cli/src/cmds/compare/DiffLoader.ts
```diff
@@ -0,0 +1,125 @@
...
+    // Process only one diff at a time.
+
+    if (this.diffQueue.length()) await this.diffQueue.drain();
+
+    const roots = [...sourcePathRoots].sort().join(' ');
+
+    this.diffQueue.push(roots);
+    await this.diffQueue.drain();
+
+    return this.diffByRoots.get(roots);
```
dividedmind, line 34: I feel this would be much cleaner and more convincingly correct using [p-limit](https://github.com/sindresorhus/p-limit). This queue stuff is fiddly.

packages/cli/src/cmds/compare/DiffLoader.ts
```diff
@@ -0,0 +1,125 @@
...
+    const diff = await this.diffLoaderQueue.diff(diffRoots);
+    if (diff) this.parseDiffStr(diff);
+  }
+
+  static isEligibleFile(root: string): boolean {
+    return !['vendor', 'node_modules'].includes(root) && existsSync(root);
+  }
+
+  private parseDiffStr(diff: Diff) {
+    if (!this.diffByFile) this.diffByFile = new Map<string, Diff>();
```
dividedmind, line 110: Why not initialize this field directly in the class definition?

packages/cli/src/cmds/compare/DiffLoader.ts
```diff
@@ -0,0 +1,125 @@
...
+        .sort()
+        .forEach((path) => {
+          assert(path);
+          if (!memo.has(path)) memo.set(path, { command: diff.command, files: [] });
+
+          memo.get(path)!.files.push(diffFile);
+        });
+      return memo;
+    }, this.diffByFile);
+  }
```
dividedmind, line 124: Using `reduce` here is quite convoluted. How about using a simple loop like this:
```suggestion
    for (const file of diff.files)
      for (const path of [file.to, file.from]) {
        if (!path) continue;
        const entry = this.diffByFile.get(path);
        if (entry) entry.files.push(file);
        else this.diffByFile.set(path, { command: diff.command, files: [file] });
      }
```
Note I removed the Set, can they really be the same? And what was the `sort` even for?

packages/cli/src/cmds/compare/DiffLoader.ts
```diff
@@ -0,0 +1,125 @@
...
+  constructor(public baseRevision: string, public headRevision: string) {
+    this.diffQueue = queue(this.executeDiffCommand.bind(this), 1);
+  }
+
+  async diff(sourcePathRoots: Set<string>): Promise<Diff | undefined> {
+    // Process only one diff at a time.
+
+    if (this.diffQueue.length()) await this.diffQueue.drain();
+
+    const roots = [...sourcePathRoots].sort().join(' ');
```
dividedmind, line 29: What if a root contains spaces?

packages/cli/src/cmds/compare/DiffLoader.ts
```diff
@@ -0,0 +1,125 @@
...
+            [chunk.content, chunk.changes.map((change) => change.content).join('\n')].join('\n')
+          )
+          .join('\n');
+        return [fromLine, toLine, content, ''].join('\n');
+      })
+      .join('\n');
+    return [content].join('\n');
+  }
+
+  async update(sourcePathRoots: Set<string>) {
```
dividedmind, line 92: Requiring a `Set` is quite a cumbersome interface (as the test code shows). There's even nothing in the code that requires it to be a set (or, indeed, the items to be unique). How about just `Iterable<string>`? (Ditto for `DiffLoaderQueue.diff`.)

dividedmind (approved):

packages/cli/src/cmds/compare/DiffLoader.ts
kgilpin: To make the diff command stable.

packages/cli/src/cmds/compare/DiffLoader.ts
kgilpin: I thought about this and thought, who is putting spaces in the root directory of their source tree?

kgilpin:
Thanks for the comments @dividedmind I will keep these in mind.

packages/cli/src/cmds/compare/DiffLoader.ts
kgilpin: Thanks for the pointer. I did take a look at that - it's also a queue so, while I don't love the `queue` we are using, it's a known quantity to me at this point. At least the concept of a queue with a concurrency of 1 is easy to understand, even if the implementation isn't especially so.

packages/cli/src/cmds/compare/DiffLoader.ts
kgilpin: Like some of the other things, it's something of a relic of a previous incarnation of the code.

packages/cli/src/cmds/compare/DiffLoader.ts
kgilpin: Yes `to` and `from` are the same when the file is modified, so actually they are often the same. `git diff` prints from and to like `a/app/models/user.rb`, `b/app/models.user.rb` - I am not sure why, but these are reported by parse-diff simply as `app/models/user.rb`, which makes a lot more sense to me.

PR #1401, opened 2023-09-25
Feat/sequence diagram auto collapses actions on startup

### LINKED ISSUE #1390: Sequence diagram auto-collapses Actions on startup display
**Note** This issue depends on the successful completion of https://github.com/getappmap/appmap-js/issues/1391. Do `1391` first.

When initially opening a sequence diagram, collapse any Actions that satisfy the following conditions:

- Greater than 4 levels deep in the stack
- All descendants are function calls (as opposed to other types such as sql queries and external service calls)
- Does not contain the Selected Action (see below). 

Ensure that:
- When an Action is selected, the ancestors of that action are expanded as necessary to reveal the selected Action.

zermelo-wisen:
Fixes #1390.

Built on top of #1391 which has this PR:
https://github.com/getappmap/appmap-js/pull/1398


packages/components/src/components/DiagramSequence.vue
```diff
@@ -283,6 +285,58 @@ export default {
...
+          actionSpec.nodeType == 'call';
+
+        if (actionShouldCollapse) this.$set(this.collapsedActionState, actionSpec?.index, true);
+
+        return descendantPreventingCollapseFound;
+      };
+
+      const headsToVisit = this.diagramSpec.actions.filter(
+        (a) =>
+          a.ancestorIndexes.length == 4 && a.action.children?.length > 0 && a.nodeType === 'call'
```
ahtrotta, line 335: After messing around with this in various appmaps, I think it's best if we change this requirement to 2 instead of 4. Also, it should be `>=` instead of `==`. So:
```
 a.ancestorIndexes.length >= 2
```

packages/components/src/components/DiagramSequence.vue
```diff
@@ -283,6 +285,58 @@ export default {
 
       return '';
     },
+    collapseActionsForCompactLook() {
+      // Collapse any Actions that satisfy the following conditions:
+      //
+      // -Greater than 4 levels deep in the stack
```
ahtrotta, line 291: After messing around with this in various appmaps, I think it's best if we change this from 4 to 2.

packages/components/tests/e2e/specs/appmap/sequenceDiagram.spec.js
```diff
@@ -191,4 +191,25 @@ context('AppMap sequence diagram', () => {
...
+      // div containing [+].
+      cy.contains('span', 'decode')
+        .first()
+        .parent()
+        .parent()
+        .parent()
+        .find('.collapse-expand')
+        .should('have.class', 'collapsed');
+    });
+  });
```
ahtrotta, line 214: Can you add tests to cover all possible combinations?

- If the action is 2+ levels deep, has only function descendants, but it is the selected action, then it shouldn't be collapsed.
- If the action is 2+ levels deep, is not the selected action, but has a sql query as a descendant, then it shouldn't be collapsed.
- etc

ahtrotta (changes requested):
This is working pretty well! There are a few changes I'd like to see:

- I'd like to see more test coverage, since there are many possible scenarios that we want to test. 
- I'd like to change the requirement from 4 levels deep in the stack to 2 levels deep in the stack. 
- I found a strange bug that doesn't appear when running Storybook, but it does appear if I [build a local version of the VS Code extension](https://github.com/getappmap/vscode-appland/blob/develop/DEVELOPMENT.md#running-in-development-with-a-local-version-of-applandcomponents) with this change. It occurs when opening an appmap. You can see it in the console of the inspector (you can open it with `Ctrl + Shift + I` when running the development extension in VS Code). Can you fix this?

image

Here is a larger screenshot for context:

image

packages/components/src/components/DiagramSequence.vue
zermelo-wisen: Sure.

By the way I'm grabbing heads with `a.ancestorLength == n`. All deeper nodes (`> n`, descendants of heads) are visited since I'm visiting heads recursively. I'll leave it as `==` if you agree.

ahtrotta: Yes, good point, you can leave it as an equality check, but can you use the strict equality operator instead?

zermelo-wisen:
I found and fixed the bug above.

@ahtrotta, Regarding tests, since we are creating this collapsed compact look only in initial render of the sequence diagram, is there a scenario in which we have a selected action before the first render? If a selected action can be selected only by the user than we won't encounter a selected action while we create this collapsed compact look. Maybe I misunderstood something here. I won't create a test for selected action if you agree.

zermelo-wisen:
I think we don't need to create this collapased compact look if we are in diff mode. What do you think?

Diff mode has its own collapse logic:
```
// If a Diagram contains any actions in diff mode, expand all ancestors of every diff action,
// and collapse all other actions.
```

kgilpin:
I agree with that.

ahtrotta:
> I found and fixed the bug above.
> 
> @ahtrotta, Regarding tests, since we are creating this collapsed compact look only in initial render of the sequence diagram, is there a scenario in which we have a selected action before the first render? If a selected action can be selected only by the user than we won't encounter a selected action while we create this collapsed compact look. Maybe I misunderstood something here. I won't create a test for selected action if you agree.

Yeah, that's a good point, don't worry about tests for selected actions.

ahtrotta:
> > I found and fixed the bug above.
> > @ahtrotta, Regarding tests, since we are creating this collapsed compact look only in initial render of the sequence diagram, is there a scenario in which we have a selected action before the first render? If a selected action can be selected only by the user than we won't encounter a selected action while we create this collapsed compact look. Maybe I misunderstood something here. I won't create a test for selected action if you agree.
> 
> Yeah, that's a good point, don't worry about tests for selected actions.

Actually, it is possible to open an appmap with an action selected. So please do make a test for this.

kgilpin:
We are doing a lot of iteration on this, and I think it would be really helpful to display the collapse depth in the UI and make it user controllable. What do you think? 

Are we going to continue to always reveal SQL, or remove that, or make it also configurable?

ahtrotta:
> We are doing a lot of iteration on this, and I think it would be really helpful to display the collapse depth in the UI and make it user controllable. What do you think?

I think that this is a good idea but it will require some design and alignment, so I think we should do this as a follow up to this PR.

> Are we going to continue to always reveal SQL, or remove that, or make it also configurable?

I just experimented with this and I think we should stop trying to always reveal SQL.

ahtrotta:
Also, I found a bug. If an action is hidden and then you try to pan to it using the eyeball, it doesn't work. When you click the eyeball it should uncollapse whatever is necessary to pan to the action. For example, if you run Storybook, visit this map, then select `SAVEPOINT active_record_1` from the SQL queries, you'll see this in the sidebar. And if you click the first eyeball, nothing happens:

image

So @zermelo-wisen, please:

- [ ] Finish adding tests as specified above, including tests that ensure that a selected action is not hidden.
- [ ] Remove the `isCollapsePreventingNodeType` function
- [ ] Fix the bug above

kgilpin:
> I think that this is a good idea but it will require some design and alignment, so I think we should do this as a follow up to this PR.

But keep in mind this change is going to affect ALL AppMaps opened so we need to get it right the first time.

kgilpin:
So how about if we stop revealing SQL but go with a deeper threshold for now and then add a UI control. 

We don’t want to “over-hide”. Under-hiding is probably better for now. Then we can use the UI control to experiment in the field with a lot of AppMaps and reconsider the default.  

Of course it’s also essential for selection / unhiding to work properly.

ahtrotta:
Okay, so now we want to add a control in the toolbar to change the depth of auto-collapsing. It should only be present when viewing the sequence diagram, and it should look approximately like this:

image

But please make the `+` and `-` in the buttons a little larger and more prominent add a light border around the number and the buttons so that it's clear that they are grouped together.

So the current checklist is:

- [ ] Finish adding tests as specified above, including tests that ensure that a selected action is not hidden.
- [ ] Remove the isCollapsePreventingNodeType function
- [ ] Fix the bug about clicking on the eyeballs
- [ ] Add a user control for auto-collapse depth (see above)

zermelo-wisen:
Ok, I'm going with these subtasks. 

To clarify user control, this is what I understand:

-Each appmap will initially open with depth level of 3.
-User can collapse or expand actions manually. 
-If the user control value is changed (clicking - or +) we re-apply the compact view with the new depth as if the AppMap is reopened with the new depth, losing manual collapse and expand interaction effects of the user.
-When opening a new AppMap or reopening the current one, it will always start with a depth level of 3, regardless of the previous AppMap's last setting.

ahtrotta:
> Ok, I'm going with these subtasks.
> 
> To clarify user control part, this is what I understand:
> 
> -Each AppMap will initially open with depth level of 3. -User can collapse or expand actions manually. -If the user control value is changed (clicking - or +) we reapply the compact view with the new depth as if the AppMap is reopened with the new depth, losing manual collapse and expand interaction effects of the user. -When opening a new AppMap or reopening the current one, it will always start with a depth level of 3, regardless of the previous AppMap's last setting.

Yup, you got it. The only edge case I would note is:

- User selects an action (it is highlighted and shown in the sidebar)
- User clicks + or - to change depth level
- We re-apply the compact view with the new depth, but we make sure that the selected action is visible

zermelo-wisen:
All done.

zermelo-wisen:
I had to change the appearance of depth control a bit. 

Toolbar's overflow is hidden, so when the space is not enough due to the space taken by new depth control, the filter button gets invisible. At least one test failed because the filter button was not clickable.

Now I'm hiding (-) and (+) buttons. I'm showing them only when the user hovers over text showing current depth (3 for example). I'm showing them with absolute positioning with high z-index, to prevent them to take extra space. I'm using mouseenter, mouseleave and click events to make it useful in desktop and mobile.

[3]
_mouseenter_
[-][3][+]
_mouseleave_
[3]
_click depth text_
[-][3][+]
_click depth text_
[3]

ahtrotta:
@zermelo-wisen thanks for your work on this, it works well! I pushed a few changes based on discussions that we've been having.

ahtrotta (approved):

PR #1409, opened 2023-10-03
feat: Inventory command, v2

kgilpin:
Update the `inventory` command, add `inventory-report` to provide a report on the "state of a repo", aka the technical debt report of a repository.

This command is integrated into the [archive-action](https://github.com/getappmap/archive-action/pull/28). When the `archive-action` runs in a PR context during setup, the welcome message is written to the PR.

The report can be improved and updated in the future. The goal for this PR is just to get basic approval and get the command released so that the downstream GitHub Actions can be merged and start working as well.


packages/cli/src/cmds/compare/detectRevisions.ts
```diff
@@ -2,13 +2,17 @@ import { executeCommand } from '../../lib/executeCommand';
 import { prominentStyle } from './ui';
 import { ValidationError } from './ValidationError';
 
+export async function detectRevision(): Promise<string> {
+  return (await executeCommand(`git rev-parse HEAD`)).trim().split('\n')[0];
```
dustinbyrne, line 6: Unreleated - but I was just looking at [simple-git](https://www.npmjs.com/package/simple-git). Might simplify things a bit for us.

dustinbyrne (approved):

packages/cli/src/cmds/compare/detectRevisions.ts
kgilpin: Looks nice, we can migrate to that.

PR #1412, opened 2023-10-04
Improve sequence diagram load time

### LINKED ISSUE #1411: Improve sequence diagram load time
When loading the sequence diagram, we call `buildDiagram` twice, which is expensive and shouldn't be necessary. 

You can see in this flame graph that `buildDiagram` is called twice and represents a lot of the processing time (each call is > 700 ms):

Image

ahtrotta:
Fixes #1411 

PRs #1416 and #1418 will need to be merged and released before this one.

- Simplify actor/action key computation (~15 % improvement)
- Avoid calling `buildDiagram` twice when loading sequence diagram (~46% improvement)
- Cache parsed SQL in Event class (~24% improvement)

### DIFF
--- a/packages/models/src/event.js
+++ b/packages/models/src/event.js
@@ -1,3 +1,4 @@
+/* eslint-disable */
 import { addHiddenProperty, hasProp, transformToJSON } from './util';
 import analyzeSQL, { abstractSqlAstJSON } from './sql/analyze';
 import normalizeSQL from './sql/normalize';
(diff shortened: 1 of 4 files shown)


packages/models/src/event.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable */
```
dustinbyrne, line 1: This looks like it's been removed upstream already, but please make sure it doesn't make it to `main` :-)

ahtrotta:
These changes are all in different PRs now.

PR #1414, opened 2023-10-04
fix: set appmap_dir for maven, gradle project

apotterri:
Set `appmap_dir` to `tmp/appmap` when creating the default configs for maven and gradle projects.


brikelly (approved):
Good catch!

PR #1418, opened 2023-10-05
Cache parsed SQL in `buildDiagram` Function

### LINKED ISSUE #1417: Improve sequence diagram load time (`@appland/models`)
This is to release a version of `@appland/models` to use in `@appland/components`.

ahtrotta:
Fixes #1417 

When loading a sequence diagram, the `buildDiagram` function is one of the most expensive function calls. When it uses the [`hash` getter](https://github.com/getappmap/appmap-js/blob/main/packages/sequence-diagram/src/buildDiagram.ts#L62) on an instance of `Event`, the [`buildStableHash` function](https://github.com/getappmap/appmap-js/blob/main/packages/models/src/event.js#L370) is called, which ultimately ends up parsing SQL if the event has any. It turns out that parsing SQL is a relatively expensive operation, so this can add up to a lot of computation if there are a lot of events with SQL in the sequence diagram. In this flame graph you can see that `peg$parseRule` accounts for 41% of `buildDiagram`:

image

This PR creates an LRU cache in the `buildDiagram` method, and then passes it into the `getHashWithSqlCache` method on the `Event` class so that we don't repeat the computation needlessly. 

This will need to be split up into two PRs so that we can release a version of `@appland/models` that will get used in `@appland/sequence-diagram`.

### DIFF
--- a/packages/models/src/event.js
+++ b/packages/models/src/event.js
@@ -362,14 +362,14 @@ export default class Event {
 
   get identityHash() {
     if (!this.$hidden.identityHash) {
-      this.$hidden.identityHash = this.buildIdentityHash(this).digest();
+      this.$hidden.identityHash = this.buildIdentityHash().digest();
     }
     return this.$hidden.identityHash;
   }
 
   get hash() {
     if (!this.$hidden.hash) {
-      this.$hidden.hash = this.buildStableHash(this).digest();
+      this.$hidden.hash = this.buildStableHash().digest();
     }
     return this.$hidden.hash;
   }
@@ -379,6 +379,13 @@ export default class Event {
       this.$hidden.stableProperties = this.gatherStableProperties();
     }
     return this.$hidden.stableProperties;
+  }
+
+  getHashWithSqlCache(parsedSqlCache) {
+    if (!this.$hidden.hash) {
+      this.$hidden.hash = this.buildStableHash(parsedSqlCache).digest();
+    }
+    return this.$hidden.hash;
   }
 
   callStack() {
@@ -501,7 +508,7 @@ export default class Event {
 
   // Collects properties of an event which are not dependent on the specifics
   // of invocation.
-  gatherStableProperties() {
+  gatherStableProperties(parsedSqlCache) {
     const { sqlQuery } = this;
 
     // Convert null and undefined values to empty strings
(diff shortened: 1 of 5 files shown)


kgilpin:
Take a look at https://github.com/getappmap/appmap-js/blob/f0ede4844dc4eb51ec20a46a99d68236c8bfbc1f/packages/scanner/src/appMapIndex.ts#L6 which also caches the parse tree.

packages/models/src/event.js
```diff
@@ -19,6 +19,8 @@ function alias(obj, prop, alias) {
 // This class supercedes `CallTree` and `CallNode`. Events are stored in a flat
 // array and can also be traversed like a tree via `parent` and `children`.
 export default class Event {
+  static parsedSqlCache = {};
```
kgilpin, line 22: Consider a Map instead of Object.

kgilpin (commented):
It’s problematic to keep a cache of every SQL query in memory because tools that process many AppMaps will end up consuming a lot of memory doing this. That’s why the scanner uses an LRU cache. 

Adding a cache to the Event will impact a lot of code in addition to the code we are trying to optimize. 

Can we add the cache we need in a way that’s more specific to the use case. Eg in the DiagramComponent, or an LRU cache in buildDiagram.

kgilpin (changes requested):
It’s problematic to keep a cache of every SQL query in memory because tools that process many AppMaps will end up consuming a lot of memory doing this. That’s why the scanner uses an LRU cache.

Adding a cache to the Event will impact a lot of code in addition to the code we are trying to optimize.

Can we add the cache we need in a way that’s more specific to the use case. Eg in the DiagramComponent, or an LRU cache in buildDiagram.

ahtrotta:
> It’s problematic to keep a cache of every SQL query in memory because tools that process many AppMaps will end up consuming a lot of memory doing this. That’s why the scanner uses an LRU cache.
> 
> Adding a cache to the Event will impact a lot of code in addition to the code we are trying to optimize.
> 
> Can we add the cache we need in a way that’s more specific to the use case. Eg in the DiagramComponent, or an LRU cache in buildDiagram.

I just changed how this works, so now I'm creating an LRU cache in `buildDiagram` [here](https://github.com/getappmap/appmap-js/pull/1418/files#diff-64d7dc670f0b84c9db079ad5c6df8e001e2d9b9590365b46c2af59962f10f6b5R23) and then passing the cache into the new `getHashWithSqlCache` method on the `Event` class that can be used instead of the `hash` getter. This won't impact any of the other tools.

PR #1423, opened 2023-10-11
Change gradle and maven installers to use the latest versions of those plugins

brikelly:
Upgrading each one to the latest (trying to specify a range was not fruitful).


apotterri (changes requested):
There are still several gradle configs in the fixtures that refer to the plugin version as `[1.0,2)`.  These should be changed to 1.2.0, right?

brikelly:
> There are still several gradle configs in the fixtures that refer to the plugin version as `[1.0,2)`. These should be changed to 1.2.0, right?

Yes, thanks - I have fixed those now.

apotterri (approved):

PR #1434, opened 2023-10-16
Improve appmap performance in components

ahtrotta:
This is the third and largest PR in a series of PRs that improve AppMap viewer performance.

There are several categories of changes in this PR.
- **Functional components**: They have improved performance over standard components because they have less overhead.
- **Stabilize props**: Whenever a prop changes the component and all children re-render. We were passing `selectedEvent` and `focusedEvent` as props down through many levels of components. These values change often and were causing unnecessary re-renders. They were already in the Vuex store, so I removed them as props and changed it so that components that actually use them have computed properties that pull them from the store.
- **Decrease repeated access to computed properties**: There is a small amount of overhead from Vue when accessing computed properties, so it's more efficient to store the computed property as a local variable than repeatedly accessing it directly (i.e. when it is accessed in a `forEach` loop).
- **Remove unnecessary prop validators**: The prop validators in the flame graph were computationally expensive for large AppMaps.

This PR also simplifies the zooming function in the flame graph so that it doesn't have a "smooth" zoom effect. The "smooth" zoom was causing several re-renders, which was problematic for medium-sized or large AppMaps.

### DIFF
--- a/packages/components/src/components/sequence/CallAction.vue
+++ b/packages/components/src/components/sequence/CallAction.vue
@@ -213,6 +213,7 @@
 
 <script lang="ts">
 // @ts-nocheck
+import { Event } from '@appland/models';
 import Arrow from '@/assets/sequence-action-arrow.svg';
 import { ActionSpec } from './ActionSpec';
 import VCallLabel from './CallLabel.vue';
@@ -234,17 +235,10 @@ export default {
       type: Array,
       required: true,
     },
-    focusedEvent: {
-      type: Object,
-      default: null,
-    },
     interactive: {
       type: Boolean,
       required: true,
       readonly: true,
-    },
-    selected: {
-      type: Boolean,
     },
     appMap: {
       type: Object,
@@ -258,6 +252,16 @@ export default {
   },
 
   computed: {
+    selected() {
+      return (
+        this.$store.getters.selectedObject &&
+        this.$store.getters.selectedObject instanceof Event &&
+        this.actionSpec.eventIds.includes(this.$store.getters.selectedObject.id)
+      );
+    },
+    focusedEvent() {
+      return this.$store.state.focusedEvent;
+    },
     containerClasses(): string[] {
       const result = [
         'call',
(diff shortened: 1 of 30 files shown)


packages/components/src/components/sequence/CallAction.vue
```diff
@@ -230,42 +231,41 @@ export default {
...
-  },
-
   computed: {
+    selected() {
+      return (
+        this.$store.getters.selectedObject &&
+        this.$store.getters.selectedObject instanceof Event &&
+        this.actionSpec.eventIds.includes(this.$store.getters.selectedObject.id)
+      );
+    },
```
dustinbyrne, line 251: Does this pattern measurably improve performance? We're still bound by one single observer for each event. What we really need is one observer per event.

packages/components/src/stories/DiagramSequence.stories.js
```diff
@@ -5,6 +5,16 @@ import list_users from '@/stories/data/sequence/list_users.sequence.json';
 import list_users_prefetch from '@/stories/data/sequence/list_users_prefetch.sequence.json';
 import show_user from '@/stories/data/sequence/show_user.sequence.json';
 import user_not_found from '@/stories/data/sequence/user_not_found.sequence.json';
+import { buildStore, SELECT_CODE_OBJECT } from '@/store/vsCode';
+import { Event } from '@appland/models';
+
+const store = buildStore();
```
dustinbyrne, line 11: I do want to be moving away from the store as much as possible. It creates an opaque layer between our components and the data they require to render. Maybe this is something that we can drop in the future along with the single observer, N event pattern.

dustinbyrne (approved):

packages/components/src/components/sequence/CallAction.vue
ahtrotta: I originally created `watch` functions on the relevant values in the store and a new `data` property in the component (e.g. `CallAction`) instead of `computed` properties and that had a noticeable performance increase. I changed it to this pattern because the `View in Sequence` button wasn't working if the sequence diagram hadn't been previously viewed/rendered.

packages/components/src/stories/DiagramSequence.stories.js
ahtrotta: I agree and that's probably why prop drilling is best practice. I'm not sure what the alternative is, though. I suppose we could do `provide`/`inject`. Do you think that is less opaque than the store?

PR #1438, opened 2023-10-16
Finding is selected in appmaps from PR report

### LINKED ISSUE #1437: Opening an AppMap from a finding in the PR report should select that finding

ahtrotta:
Fixes #1437 

When opening an AppMap from the findings section of a PR report, the relevant finding is selected in the AppMap viewer. This is accomplished by adding a `state` query param that has a base64-url-encoded string as its value.

### DIFF
--- a/packages/cli/src/cmds/compare-report/ReportSection.ts
+++ b/packages/cli/src/cmds/compare-report/ReportSection.ts
@@ -2,12 +2,13 @@ import Handlebars, { SafeString } from 'handlebars';
 import { isAbsolute, join, relative } from 'path';
 
 import { readFile } from 'fs/promises';
-import ChangeReport, { AppMap, FindingDiff } from './ChangeReport';
+import ChangeReport, { AppMap } from './ChangeReport';
 import { existsSync } from 'fs';
 import assert from 'assert';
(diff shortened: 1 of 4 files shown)


packages/cli/src/cmds/compare-report/ReportSection.ts
```diff
@@ -219,29 +219,50 @@ export default class ReportSection {
...
+          const stateObject = { selectedObject: `analysis-finding:${findingHash}` };
+          const state = base64UrlEncode(JSON.stringify(stateObject));
+          searchParams.state = state;
+        } catch (e) {
+          // do not add state
+        }
+        url = buildUrlString(searchParams);
       }
+      return new SafeString(url);
     };
```
dustinbyrne, line 266: This looks good. Could you add a test for this to ensure this helper is constructing URLs as expected?

dustinbyrne (changes requested):

packages/cli/src/cmds/compare-report/ReportSection.ts
ahtrotta: Yup, added tests for this helper.

dustinbyrne (approved):

PR #1444, opened 2023-10-20
Changing installer to add our gem with double quotes

brikelly:
Done to satisfy standardrb linters that require double quoted strings.


kgilpin:
I believe there are also linters that want single quotes if double quotes aren’t needed.

brikelly:
> I believe there are also linters that want single quotes if double quotes aren’t needed.

Yeah this was mostly me just trying things out in a draft PR. Sounds like we need to come at this more holistically with the kind of interception step that allows for users to influence things like quote styles, linebreaks, etc.

kgilpin:
Conforming to standardrb sounds like a good default though.

Issue #1447, opened 2023-10-24
`Show in Sequence` does not work if collapse level is too low


ahtrotta:
To re-create the issue:

1. Run `yarn storybook` and then visit [this appmap](http://localhost:6006/iframe.html?args=&id=pages-vs-code--extension&viewMode=story).
2. Navigate to the Sequence Diagram tab and decrease the collapse level to `0`.

![Image](https://github.com/getappmap/appmap-js/assets/45714532/44103880-7e56-491b-a0e1-a4976a21ac35)

3. Navigate to `Dependency Map` tab, click on the database and select the first query.

![Image](https://github.com/getappmap/appmap-js/assets/45714532/b481671f-94ad-41c3-b883-29f5ae279a4a)

4. Then click `Show in Sequence`. Note that the selected item is not visible.

![Image](https://github.com/getappmap/appmap-js/assets/45714532/13a48e3b-a33c-4cc4-91f0-5018b5039e13)

## Solution

When clicking `Show in Sequence`, expand the relevant parts of the sequence diagram so that the selected object can be viewed.

PR #1451, opened 2023-10-29
Show in sequence when collapse level is low

moxw:
Fixes: #1447 

feat(DetailsPanelEvent.vue): add SET_FOCUSED_EVENT in viewEventInSequence
feat(DiagramSequence.vue): emit 'updateCollapseDepth' on collapse change
feat(VsCodeExtension.vue): add handler for new collapse depth
test(sequenceDiagram.spec.js): test action expansion in sequence view

### DIFF
--- a/packages/components/tests/e2e/specs/appmap/sequenceDiagram.spec.js
+++ b/packages/components/tests/e2e/specs/appmap/sequenceDiagram.spec.js
@@ -298,5 +298,28 @@ context('AppMap sequence diagram', () => {
       cy.get('div.depth-text').first().contains('2');
       getCollapseExpandElementOfActionLabel('index').should('have.class', 'expanded');
     });
+
+    it('action is expanded when showed in sequence', () => {
+      cy.get('div.depth-text').first().contains('3');
+
+      cy.get('button[data-cy="decrease-collapse-depth"]').click({ force: true });
+      cy.get('button[data-cy="decrease-collapse-depth"]').click({ force: true });
+      cy.get('button[data-cy="decrease-collapse-depth"]').click({ force: true });
+      cy.get('div.depth-text').first().contains('0');
+
+      cy.get('.tabs .tab-btn').contains('Dependency Map').click();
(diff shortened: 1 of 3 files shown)


packages/components/src/components/DetailsPanelEvent.vue
```diff
@@ -240,6 +246,7 @@ export default {
 
   methods: {
     viewEventInSequence() {
+      this.$store.commit(SET_FOCUSED_EVENT, this.object);
```
ahtrotta, line 249: A focused event and a selected event are different. The focused event is temporarily highlighted with teal and is not selected in the sidebar. A selected event is permanently highlighted with pink and is selected in the sidebar. 

We don't want to make the selected event the focused event, because it creates a confusing scenario where we use the focused event highlighting for the selected event. Specifically, notice how the event is highlighted in teal in the sequence diagram after clicking on the `Show in Sequence` button:

image

ahtrotta (changes requested):
Thanks for the PR @moxw , it solves the issue but we don't want to set the selected event as the focused event when a user clicks on the `Show in Sequence` button.  Can you make this fix without using `SET_FOCUSED_EVENT` on the selected event?

packages/components/tests/e2e/specs/appmap/sequenceDiagram.spec.js
```diff
@@ -298,5 +298,28 @@ context('AppMap sequence diagram', () => {
       cy.get('div.depth-text').first().contains('2');
       getCollapseExpandElementOfActionLabel('index').should('have.class', 'expanded');
     });
+
+    it.only('action is expanded when showed in sequence', () => {
```
ahtrotta, line 302: Please remove `only`

moxw: Done

ahtrotta (approved):
Looks good, thanks!

Issue #1420, opened 2023-10-06
AppMap code objects in side panel can be sorted by different criteria


brikelly:
On the left side panel, it can be difficult to get the right information when such a long list is present.

![Image](https://github.com/getappmap/appmap-js/assets/573108/d9a273a0-c8af-462d-91d6-f495a57ba41f)

We need to add a `Sort By` visual control to the side panel that offers these options:

1. Name
2. Execution order
3. Occurrences
4. Elapsed time

PR #1454, opened 2023-11-02
AppMap code objects sorting

moxw:
Fixes: #1420 

Screenshot_20231103_032651
Screenshot_20231103_032701


ahtrotta (changes requested):
Hi @moxw, this is looking great! I changed the CSS a little bit so that it's more clear to the user what is being used to sort. I see that you have tests for sorting by name and by occurrences, but can you add similar tests for sorting by execution order and elapsed time?

When you make changes, please use `fixup!` commits. To do this, make your changes and stage them, then `git commit --fixup=e0650699e`, where `e0650699e` is the commit hash. I'll squash the `fixup!` commits before merging. Thanks!

moxw:
> Hi @moxw, this is looking great! I changed the CSS a little bit so that it's more clear to the user what is being used to sort. I see that you have tests for sorting by name and by occurrences, but can you add similar tests for sorting by execution order and elapsed time?
> 
> When you make changes, please use `fixup!` commits. To do this, make your changes and stage them, then `git commit --fixup=e0650699e`, where `e0650699e` is the commit hash. I'll squash the `fixup!` commits before merging. Thanks!

Added the test cases.

ahtrotta (approved):

PR #1459, opened 2023-11-06
fix: Ignore missing or invalid HTTP status

kgilpin:
If the HTTP status in a recorded response is missing or invalid, ignore the response. Failure to do this results in invalid OpenAPI, and errors downstream.


packages/openapi/src/method.ts
```diff
@@ -4,6 +4,7 @@ import Response from './response';
 import { headerValue, RPCRequest } from './rpcRequest';
 import ObjectSchema from './objectSchema';
 import { messageToOpenAPISchema, parseScheme } from './util';
+import { warn } from 'console';
```
dustinbyrne, line 7: The build is failing because this code is also loaded into a browser context where this module doesn't exist. The easy fix here would be to use `console.warn`, while the non-trivial fix would be to create a separate distribution artifact for `@appland/openapi` in a browser context. The only code in use in the browser (AFAIK) is a utility function: [`classNameToOpenAPIType`](https://github.com/getappmap/appmap-js/blob/aa25ac9b19e27556f46a2cfe698ecf6984497d97/packages/openapi/src/util.ts#L42)

kgilpin:
Yeah, I am working on it.

PR #1462, opened 2023-11-09
feat: Recognize certain labeled functions as roots

kgilpin:
The following labels are recognized as being "roots" (aka commands):

* `cli.command`
* `job.perform`
* `message.handle`

These labels extend the existing definition of a "root", which is currently only HTTP server requests.

A follow-on to this will change "Limit root events to HTTP" to something like "Limit root events to commands":

Screen Shot 2023-11-10 at 5 33 35 PM

PS "Hide unlabeled" seems kind of useless, maybe we should drop it.


ahtrotta (approved):
Cool, looks like a useful feature :+1:

PR #1463, opened 2023-11-09
Export sequence diagram page with the Vuex store

ahtrotta:
Export the sequence diagram as a page that includes the `store`. This is required to use the sequence diagram as a standalone component.


packages/components/src/pages/SerializedSequence.vue
```diff
@@ -0,0 +1,30 @@
+<template>
+  <VDiagramSequence :serialized-diagram="serializedDiagram" :interactive="interactive" />
+</template>
+
+<script>
+import VDiagramSequence from '@/components/DiagramSequence.vue';
+import { store } from '@/store/vsCode';
+
+export default {
+  name: 'v-serialized-sequence',
```
dustinbyrne, line 10: I'm not sure about the name, `v-sequence-diagram` seems fine even if it's similar to `v-diagram-sequence`, at least it's a page vs. component.

The fact that the data (?) is serialized seems irrelevant to me.

packages/components/src/componentExports.js
```diff
@@ -9,3 +9,4 @@ export { default as VAnalysisFindings } from '@/pages/AnalysisFindings.vue';
 export { default as VFindingDetails } from '@/pages/FindingDetails.vue';
 export { default as VOpenAppmaps } from '@/pages/install-guide/OpenAppMaps.vue';
 export { default as VSidebarSignIn } from '@/components/SignIn.vue';
+export { default as VSerializedSequence } from '@/pages/SerializedSequence.vue';
```
dustinbyrne, line 12: Is the `VDiagramSequence` export used anywhere? Can we drop it?

packages/components/src/pages/SerializedSequence.vue
ahtrotta: Yeah, I had `v-sequence-diagram` originally, but it seemed confusing to have that and `v-diagram-sequence`, but it is a better name so I'll change it back.

packages/components/src/componentExports.js
ahtrotta: I'll look into it

packages/components/src/pages/SerializedSequence.vue
kgilpin: Please change the file name too if you do that.

packages/cli/src/html/sequenceDiagram.js
```diff
@@ -16,7 +16,7 @@ async function initializeApp() {
   return new Vue({
     el: '#app',
     render: (h) =>
-      h(VDiagramSequence, {
```
dustinbyrne, line 19: I'd split this into a separate commit due to the fact this is a separate package.

dustinbyrne (approved):

packages/cli/src/html/sequenceDiagram.js
ahtrotta: :heavy_check_mark:

PR #1464, opened 2023-11-15
feat: Ask OpenAI

kgilpin:
This is old, and basically a placeholder issue now for the "Fix Wrench" - which may get rolled into Navie?


packages/cli/src/cmds/ask/ask.ts
```diff
@@ -0,0 +1,307 @@
...
+      const specification = Specification.build(filteredAppMap, { loops: true });
+      const diagram = buildDiagram(appmapFile, filteredAppMap, specification);
+
+      const diagramText: string[] = [];
+      const maxDepth = 2;
+      const collectAction = (action: Action, depth = 0) => {
+        if (depth > maxDepth) return;
+
+        const indent = '  '.repeat(depth);
+        diagramText.push(`${indent}${nodeName(action)}(events:${action.eventIds.join(', ')}})`);
```
dustinbyrne: FYI There's a stray curly brace at the end of this string. Not sure how that may affect accuracy.

PR #1469, opened 2023-11-22
feat: context filter

kgilpin:
Filter AppMaps by `context`.

The `context` filter is a list of code object names or patterns. An event is included by the filter if:

* It's code object matches a context name.
* It's an ancestor of a matching code object.
* It's a descendant within a configurable `depth` of a matching code object.


packages/models/src/appMapFilter.js
```diff
@@ -68,21 +81,39 @@ function filterRegExp(filterExpression, regexpConstructorArgs) {
...
+//   ancestor is within the maxDepth, then the descendant is included as well.
+// maxDepth: The maximum depth of a descendant event to include. If undefined, all descendants are included.
+// Returns: A set of events that matched the filterFn.
+function markSubtrees(events, filterFn, maxDepth) {
+  const matchingEvents = new Set();
+
+  // Collect all 'call' events that match explicitly.
+  events
+    .filter((e) => e.isCall())
+    .forEach((e) => (filterFn(e) ? matchingEvents.add(e) : undefined));
```
dustinbyrne, line 96: Nit: Calling `filterFn` in the filter would clean this up a bit.

```ts
events
    .filter((e) => e.isCall() && filterFn(e))
    .forEach((e) => matchingEvents.add(e))
```

packages/models/loadPetWithVisit_depth_1.appmap.json
```diff
@@ -0,0 +1,243 @@
+{
```
dustinbyrne, line 1: It looks like this file was accidentally added

packages/models/listMicropostsCompactAppMap_sql_context.appmap.json
```diff
@@ -0,0 +1,177 @@
+{
```
dustinbyrne, line 1: This also looks accidentally added as it also exists in `tests/unit/fixtures`

packages/models/types/index.d.ts
```diff
@@ -365,6 +365,12 @@ declare module '@appland/models' {
       constructor(on: boolean, defaultValue: boolean, names: string[]);
     }
 
+    export class DeclutterContextNamesProperty extends DeclutterNamesProperty {
+      public depth: number;
+
+      constructor(on: boolean, defaultValue: boolean, names: string[], depth: number);
```
dustinbyrne, line 371: These arguments aren't required because they have default values set in the implementation. Marking them as optionally undefined might be a better fit.

dustinbyrne (approved):

packages/models/types/index.d.ts
kgilpin: OK, I was following along the other example types in this file. I've made the depth optional.

PR #1472, opened 2023-11-27
Correctly set initial collapse depth in sequence diagram

### LINKED ISSUE #1471: Collapse depth is always 0 in VS Code
When using Storybook the collapse depth is displayed properly, but when opening a map in the VS Code extension, the collapse depth is always initially set to `0`.

Image

ahtrotta:
Fixes #1471 

In the VS Code extension, the collapse depth was initially being set to `0` whenever a map was opened. This behavior only happens in VS Code and it does not happen in Storybook.

## Before

Image

## After

image

### DIFF
--- a/packages/components/src/components/DiagramSequence.vue
+++ b/packages/components/src/components/DiagramSequence.vue
@@ -269,6 +269,7 @@ export default {
       rootActionSpecs.forEach((h) => visit(h.action));
     },
     getMaxActionDepth() {
+      if (this.actions.length === 0) return undefined;
       return this.actions.reduce((maxDepth, action) => {
         const depth = action.ancestorIndexes.length;
         if (depth > maxDepth) return depth;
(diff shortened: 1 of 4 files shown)


kgilpin:
What’s the root cause? Different initialization order? 

Will the max depth always be 3 now?

ahtrotta:
> What’s the root cause? Different initialization order?

VS Code initially loads an empty appmap and then subsequently loads an appmap with all the correct data. So when it loads the empty map, the max depth gets set to `0` and stays there.

> Will the max depth always be 3 now?

No, the max depth will be correctly computed.

ahtrotta:
I looked into how we load AppMaps in VS Code and I don't think we should change it. To change how the map loads, we would need to change the entire way that the Vue component works.

brikelly (approved):

PR #1474, opened 2023-11-27
Add AppMap conversational frontend

dustinbyrne:
This PR adds a conversational front-end component and client API.


packages/components/src/pages/Chat.vue
```diff
@@ -0,0 +1,182 @@
...
+
+export default {
+  name: 'v-chat',
+  components: {
+    VUserMessage,
+    VChatInput,
+    VSuggestionGrid,
+  },
+  props: {
+    apiKey: {
```
kgilpin, line 44: Can it get an interface to the chat provider instead? The API key as a component parameter surprises me.

packages/components/src/pages/Chat.vue
```diff
@@ -0,0 +1,182 @@
...
+  },
+  data() {
+    return {
+      messages: [] as Message[],
+      threadId: undefined as string | undefined,
+    };
+  },
+  computed: {
+    inputPlaceholder() {
+      return 'Where do you want to go?';
```
kgilpin, line 65: When I see this prompt in the demos, it's not quite landing for me; a little too abstract.

PR #1480, opened 2023-12-04
Better navigation for selected code objects

ahtrotta:
[Try out this feature here](https://5fe18baf66f2280021e634d0-fksdwkxkrf.chromatic.com/iframe.html?args=&id=pages-vs-code--extension-with-default-sequence-view&viewMode=story)

This PR introduces a new navigation menu for selected objects. This was inspired by Kevin's work in which he wanted to open an AppMap with multiple code objects selected. This update makes it feasible to navigate multiple selected code objects using the arrows and the dropdown menu (see below). A follow-on PR will make it possible to pass a list of selected objects when opening an AppMap.

This PR makes the following changes:
* A new navigation menu for selected code objects consisting of left and right arrows and a drop-down menu.
* The code object parents (if any) are now below the code object header.
* The `Help us improve AppMap` request is only shown when no code objects are selected.

## Before

image

## After

image


dustinbyrne (approved):
Looks good! I'm leaving a few minor comments which you can choose to address however you like.

1. There's still an additional trailing chevron separator in the 'parents' section 
image

2. What ends up in the history dropdown feels a little inconsistent. If I click something in the main viewport, it's added to the history. However, arrow keys don't seem to affect it. I think it'd be best if history was maintained only when navigating the left side panel and reset (cleared) when selecting something from the main panel.

ahtrotta:
> Looks good! I'm leaving a few minor comments which you can choose to address however you like.
> 
> 1. There's still an additional trailing chevron separator in the 'parents' section
>    image
> 2. What ends up in the history dropdown feels a little inconsistent. If I click something in the main viewport, it's added to the history. However, arrow keys don't seem to affect it. I think it'd be best if history was maintained only when navigating the left side panel and reset (cleared) when selecting something from the main panel.

I think I'll merge this as-is and make a new issue for each of your comments as follow-ons.

PR #1490, opened 2023-12-11
Ensure that styles are scoped

ahtrotta:
@dustinbyrne mentioned that we have Vue components with non-`scoped` style tags, meaning the styles are not scoped to the Vue element where they are defined. This PR makes it so that nearly every Vue component has `scoped` styles. Unfortunately, the largest element (`VsCodeExtension.vue`) is still un-`scoped` because `scope`-ing it breaks a bunch of stuff. Fixing that will be a larger effort for a different time IMO.

### DIFF
--- a/packages/components/src/components/Accordion.vue
+++ b/packages/components/src/components/Accordion.vue
@@ -34,7 +34,7 @@ export default {
 };
 </script>
 
-<style lang="scss">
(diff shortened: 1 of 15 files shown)


dustinbyrne:
image
I think it'll be best if we scope components as we see them.

PR #1498, opened 2023-12-14
Configuration tool changes for new Node agent

brikelly:
This will change the existing instructions based on `appmap-agent-js` to use `appmap-node` instead.


kgilpin:
@dustinbyrne important that we have a reliable way to keep the doc agent updated!

kgilpin:
Closed due to - implement by other work.

PR #1499, opened 2023-12-15
Sidebar above map

ahtrotta:
This is WIP and it's proving even more difficult than expected, so it may not be worth the effort.


kgilpin:
It seems we should go with a pop out to show and pop out to hide effect, since both sliding out animation and “above map” are proving difficult. Adam you’re right to question if this is worth the trouble.

PR #1500, opened 2023-12-20
feat: Chat + Search with server-side AI

kgilpin:
* Adds an [`explain` RPC function](https://github.com/getappmap/appmap-js/pull/1500/files#diff-61690c952470b87ba7e54edd5330753aaa43d6438a2b294900b8b2056d60007f)
* Adds an [RPC client](https://github.com/getappmap/appmap-js/pull/1500/files#diff-1a22c505fc95f3c9440dc67a8eddcc2ff979c25b35c9b0425402f2a498e1630e)
* RPC client is [used to ask the question](https://github.com/getappmap/appmap-js/pull/1500/files#diff-bd71fe5e839c60f8c8f8ea43113f261760b1bd9297f16dac1550511b7b7b0b96R225).


packages/client/src/aiClient.ts
```diff
@@ -77,9 +82,9 @@ export default class AIClient {
     });
   }
 
-  panic(error: Error): never {
+  panic(error: Error): void {
     this.disconnect();
-    throw error;
+    this.callbacks.onError?.(error);
```
kgilpin, line 87: @dustinbyrne throwing this error was causing the RPC server to exit, so I changed it to call the `onError` callback. Is that OK?

PR #1503, opened 2023-12-20
feat: Provide RPC methods for 'explain'

kgilpin:
Add the RPC `explain` and `explain.status` functions.

`explain.status` is used because JSON-RPC does not support events or callbacks.

@dustinbyrne there are a few changes to the Client that you may want to look at - https://github.com/getappmap/appmap-js/pull/1503/files#diff-55749a5e7fbf0ad45cf0b61341650b74abe6dcb491c4fc8605c6f60061675fce

These were needed to get the error handling working.


packages/cli/src/cmds/index/rpcServer.ts
```diff
@@ -45,10 +51,14 @@ export default class RPCServer {
     log(`Available JSON-RPC methods: ${Object.keys(rpcMethods).sort().join(', ')}`);
     warn(`Consult @appland/rpc for request and response data types.`);
 
-    this.server = new jayson.Server(rpcMethods).http();
-    this.server.listen(this.bindPort).on('listening', () => {
-      assert(this.server);
-      const address = this.server.address();
+    const server = new jayson.Server(rpcMethods);
```
dustinbyrne, line 54: It's worth mentioning that the server won't be instantiated until all AppMaps are polled to determine whether or not they need to be indexed. On a local project of mine, it took about 80 seconds before the server was started.

dustinbyrne (approved):

packages/cli/src/cmds/index/rpcServer.ts
kgilpin: How many AppMaps was that? It should just be checking the timestamp of the AppMap JSON file against the timestamp in the index.

dustinbyrne: Looks like I have about 7000 :smile:

kgilpin: It should only have to check the timestamp of each AppMap and then read one small file, I wouldn't expect that to take 80 seconds for 7,000 files.

PR #1505, opened 2023-12-20
Fix and re-enable "View in Sequence" test

ahtrotta:
This test was failing because calling `scrollIntoView` on the selected element was doing nothing. This PR fixes this problem by re-calling `scrollIntoView` until scrolling starts.

### DIFF
--- a/packages/components/src/components/DiagramSequence.vue
+++ b/packages/components/src/components/DiagramSequence.vue
@@ -196,15 +196,42 @@ export default {
         element.scrollIntoView(SCROLL_OPTIONS);
       });
     },
+    scrollElementIntoView(element) {
+      const boundingRect = element.getBoundingClientRect();
+
+      let intervalId;
+      let retries = 0;
+      let lastScrollTop = boundingRect.top;
+      let lastScrollLeft = boundingRect.left;
+
+      // Once scrolling has started or we've retried more than 5 times, stop retrying
+      const tryToScrollIntoView = () => {
+        element.scrollIntoView(SCROLL_OPTIONS);
+        const newBoundingRect = element.getBoundingClientRect();
+
+        if (
+          lastScrollTop !== newBoundingRect.top ||
+          lastScrollLeft !== newBoundingRect.left ||
+          retries > 5
+        ) {
+          clearInterval(intervalId);
+        } else {
+          lastScrollLeft = newBoundingRect.left;
+          lastScrollTop = newBoundingRect.top;
+          retries += 1;
+        }
+      };
+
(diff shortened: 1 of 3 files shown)


kgilpin:
I thought that the assertions would retry for a while? So explicit waiting was not necessary?

ahtrotta:
> I thought that the assertions would retry for a while? So explicit waiting was not necessary?

The Cypress assertions have a built-in waiting system, but I believe that the problem was that the `scrollIntoView` function was being called before the DOM was fully rendered, so it didn't scroll the selected element into the center of the viewport correctly. If the element isn't visible then Cypress will keep re-trying and failing to find it in the viewport.

kgilpin:
Ok, is there anyway to determine when the DOM is ready? Fixes based on milliseconds are so fragile…

kgilpin:
also a fixed wait might pass the test but it might not work in real world scenarios eg with large AppMaps.

ahtrotta:
> Ok, is there anyway to determine when the DOM is ready? Fixes based on milliseconds are so fragile…

The `this.$nextTick` function is supposed to do that, but it wasn't working unless I also added a timeout, but I'll try to make it more deterministic.

> also a fixed wait might pass the test but it might not work in real world scenarios eg with large AppMaps.

This is a weird situation where I've never run into this problem in real life. It's only appeared in this test for me.

dustinbyrne:
How often was it failing? It looks like this is either a workaround for a separate outstanding issue or a case where the assertion was timing out before the scroll finished. Though, if it were the latter, this likely wouldn't fix it.

ahtrotta:
> How often was it failing? It looks like this is either a workaround for a separate outstanding issue or a case where the assertion was timing out before the scroll finished. Though, if it were the latter, this likely wouldn't fix it.

This test was failing in CI every time. It wasn't a problem with the assertion timing out, the scroll just wasn't happening. I verified that the `scrollIntoView` function was being called correctly, it just wasn't scrolling the element into view in this test for some reason. As I said above, I've never seen this problem occur outside of Cypress.

ahtrotta:
I've run into similar problems when writing tests for the [search bar](https://github.com/getappmap/appmap-js/pull/1501). I did some research and it seems like [this is a problem with Cypress](https://github.com/cypress-io/cypress/issues/805#issuecomment-608107451). It doesn't work correctly when using `scrollIntoView` with the `{ scrollBehavior: 'smooth' }` option.

ahtrotta:
Since this is a problem with Cypress, I think we should modify how our Cypress tests run rather than modify our Vue code. I'll make a fix in a different issue/PR.

kgilpin:
Nice find!

PR #1511, opened 2023-12-29
fix: Ensure that the loading indicator is hidden on error

kgilpin:


packages/components/tests/unit/chat/ChatSearch.spec.js
```diff
@@ -0,0 +1,143 @@
...
+
+        return wrapper;
+      }
+
+      describe('401 code', () => {
+        it('activates the login instructions', async () => {
+          const wrapper = await simulateError(undefined, { code: 401, message: 'Unauthorized' });
+          expect(wrapper.find('.status-unauthorized').exists()).toBe(true);
+          expect(wrapper.find('.messages .message').exists()).toBe(false);
+          expect(wrapper.find('.messages [data-error="true"]').exists()).toBe(false);
```
dustinbyrne, line 97: Try to use `data-*` attributes as much as possible. We use the `data-cy` attribute as specific selector for test cases. From the [Cypress docs](https://docs.cypress.io/guides/references/best-practices#Selecting-Elements):

> Targeting the element above by `tag`, `class` or `id` is very volatile and highly subject to change. You may swap out the element, you may refactor CSS and update ID's, or you may add or remove classes that affect the style of the element.
>
> Instead, adding the `data-cy` attribute to the element gives us a targeted selector that's only used for testing.
>
> The `data-cy` attribute will not change from CSS style or JS behavioral changes, meaning it's not coupled to the behavior or styling of an element.
>
> Additionally, it makes it clear to everyone that this element is used directly by test code.

kgilpin: Ok. In this case it seems like .messages was unlikely to change.

dustinbyrne (approved):

PR #1512, opened 2023-12-29
Parameterize prompts

dustinbyrne:
This change allows instances of the Chat page to specify their own prompt recommendations.

### DIFF
--- a/packages/components/src/pages/Chat.vue
+++ b/packages/components/src/pages/Chat.vue
@@ -1,6 +1,6 @@
 <template>
   <div class="chat-container">
-    <v-chat ref="vchat" class="chat" :send-message="sendMessage" />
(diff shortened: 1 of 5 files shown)


packages/components/src/pages/Chat.vue
```diff
@@ -23,6 +23,10 @@ export default {
       type: String,
       default: DefaultApiURL,
     },
+    suggestions: {
```
kgilpin, line 26: Will this have a default which is the current suggestions?

dustinbyrne: It didn't. I added defaults in https://github.com/getappmap/appmap-js/pull/1512/commits/6e7bdf437399308ada6a8139b89db2b01ebdda92.

ahtrotta (approved):

PR #1514, opened 2023-12-29
fix: "New chat" resets thread

kgilpin:


packages/components/src/components/chat/Chat.vue
```diff
@@ -115,32 +149,30 @@ export default {
...
-        message: content || '',
-        sentiment: 0,
-      };
-      this.messages.push(message);
-
+    // Creates a new message, from the user or from the AI.
+    addUserMessage(content: string) {
+      this.messages.push(new UserMessage(content));
+      this.scrollToBottom();
+    },
```
dustinbyrne, line 172: Is this comment accurate? It looks like this is only for user messages.

kgilpin: Yeah, it looks like it's only for User messages.

packages/components/tests/unit/chat/ChatHelp.spec.js
```diff
@@ -0,0 +1,100 @@
...
+      expect(wrapper.find(`.messages`).text()).toContain('Begin by installing the AppMap agent');
+
+      await waitFor(
+        `.messages should contain " then configure your project per the instructions"`,
+        () =>
+          wrapper
+            .find(`.messages`)
+            .text()
+            .includes(' then configure your project per the instructions')
+      );
```
dustinbyrne, line 59: `waitFor` doesn't look to be necessary. These elements should all be in an expected state the first time this assertion is made.

packages/components/tests/unit/chat/ChatHelp.spec.js
```diff
@@ -0,0 +1,100 @@
...
+        ]);
+
+        expect(wrapper.find('.status-unauthorized').exists()).toBe(false);
+        expect(wrapper.find('.messages .message').exists()).toBe(true);
+        expect(wrapper.find(`.messages`).text()).toContain('Begin by installing the AppMap agent');
+
+        // NOTE: Ideally this would have different formatting than a regular server message.
+        await waitFor(`.messages should contain "Server error"`, () =>
+          wrapper.find(`.messages`).text().includes('Server error')
+        );
```
dustinbyrne, line 94: Same thing with this `waitFor`

dustinbyrne (approved):

packages/components/tests/unit/chat/ChatHelp.spec.js
kgilpin: They aren't though.

PR #1517, opened 2024-01-04
feat: Update appland dependency versions

kgilpin:
@dustinbyrne @ahtrotta I am a bit confused - it is necessary to do this?


dustinbyrne (approved):
For the time being, yes.

PR #1518, opened 2024-01-04
feat: Add a page to notify of invalid license

dustinbyrne:
This change adds the following two notices for use in Confluence. There's an additional change which emits link clicks as events so they can be passed through an API to (purposefully) circumvent CSP.

image
image

### DIFF
--- a/packages/components/src/pages/VsCodeExtension.vue
+++ b/packages/components/src/pages/VsCodeExtension.vue
@@ -330,32 +330,7 @@
       </div>
     </div>
 
-    <div class="no-data-notice" v-if="isEmptyAppMap && !isLoading && !hasStats">
-      <div class="notice">
-        <p class="no-data-notice__title">Sorry, but there's no data to display :(</p>
-        <ul class="why-me">
-          <strong>Top 3 reasons why this appmap is empty:</strong>
-          <li>appmap.yml did not list packages/modules/folders of your application logic</li>
-          <li>
-            If this AppMap was recorded from a test, the test did not provide sufficient coverage
-            for good data
-          </li>
-          <li>
-            If other manual method was used to record this AppMap, the instrumented code objects
-            were not executed during the recording.
-          </li>
-        </ul>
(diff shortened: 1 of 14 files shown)


kgilpin:
@dustinbyrne the Confluence plugin shows the Loading spinner until it's configured with an AppMap. Would it make sense to also in this PR replace that loading spinner with instructions to configure the plugin component?

dustinbyrne:
> @dustinbyrne the Confluence plugin shows the Loading spinner until it's configured with an AppMap. Would it make sense to also in this PR replace that loading spinner with instructions to configure the plugin component?

Thanks for the feedback, I've added this notice. See the screenshot above.

packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -173,14 +173,19 @@
             </button>
           </v-popper>
           <v-popper
-            v-if="showDownload && !isGiantAppMap"
+            v-if="showDownload"
             class="hover-text-popper"
             text="Export in SVG format"
```
kgilpin, line 178: Ability to download the AppMap from the AppMap itself would be pretty useful for things like the Confluence plugin - so the user who's viewing the page can download the data directly.

As I've thought about this, it would also be useful to include the filter settings in the AppMap itself, say under a `filter` key at the top level. This would be a way that we could make the filter settings portable with the AppMap. Suppose I am looking at an AppMap in my code editor and I've configured the filter settings. Then I click the "Export AppMap data", it opens the AppMap file in my system Finder / Explorer with the filter settings embedded. I drop this into Confluence and my filter settings will be preserved.

kgilpin (approved):

evlawler:
@dustinbyrne  Change "To continue enjoying AppMap's enhanced features for Atlassian an active license is required. We appreciate your purchase of this integration."

dustinbyrne:
Thanks for the feedback. Here's an updated screenshot.

image

evlawler:
@dustinbyrne looks good.

PR #1519, opened 2024-01-04
fix: Allow user scroll

kgilpin:
* When the user scrolls up, stop auto-scrolling.
* When the user scrolls back down to the bottom again, resume auto-scrolling.

It's important to base this on the `scrollTop` property, rather than on the `scrollHeight`, because the `scrollHeight` is frequently changing due to new content being added. It's also not a good idea to try and track state variables like `autoScroll` and activate/cancel them based on scroll events, because there's no good way to distinguish between scroll events that are triggered by the user and those that are triggered by DOM changes.

https://www.loom.com/share/a32dd639a44a4290a620fd062a61ca6c

### DIFF
--- a/packages/components/src/components/chat/Chat.vue
+++ b/packages/components/src/components/chat/Chat.vue
@@ -3,7 +3,7 @@
     <div class="button-panel" v-if="isChatting">
       <v-button class="clear" size="small" kind="ghost" @click.native="clear">New chat</v-button>
     </div>
-    <div class="messages" data-cy="messages" ref="messages">
+    <div class="messages" data-cy="messages" ref="messages" @scroll="manageScroll">
       <v-user-message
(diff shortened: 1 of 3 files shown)


packages/components/src/components/chat/Chat.vue
```diff
@@ -235,6 +250,26 @@ export default {
 
       message.sentiment = sentiment;
     },
+    isAutoScrolled() {
+      const { messages } = this.$refs;
+      return messages.scrollTop === this.autoScrolTop;
```
dustinbyrne, line 255: This will always evaluate to false due to `autoScrolTop` (sp) being undefined. In fact, it looks like this method is unused.

dustinbyrne (approved):

PR #1524, opened 2024-01-05
Map controls are responsive to column width

### LINKED ISSUE #1522: Appmap toolbar gets cut off in chat
This is a general problem, but it has recently revealed itself when we've been working on the Navie chat interface. The AppMap toolbar gets cut off:

Image

This is because there is responsive styling that is [reactive to the width of the entire page](https://github.com/getappmap/appmap-js/blob/main/packages/components/src/components/Tabs.vue#L134). Instead, we should make the styling responsive to the width of the Vue component (`#app`).

ahtrotta:
Fixes #1522 

This PR fixes the issue where the AppMap control buttons were being clipped in the Chat Search interface. Instead of using the width of the viewport for responsive styling, this PR uses the width of the AppMap view component.

## Before

image

## After

image

### DIFF
--- a/packages/components/src/components/Tabs.vue
+++ b/packages/components/src/components/Tabs.vue
@@ -104,6 +104,10 @@ export default {
     padding-top: 0.4rem;
     z-index: 100;
     position: absolute;
+
+    @media (max-width: 900px) {
+      overflow: auto;
+    }
   }
 
   &__group {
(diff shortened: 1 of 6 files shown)


packages/components/src/components/Tabs.vue
```diff
@@ -131,36 +135,35 @@ export default {
...
-    }
-    &__content {
-      padding-top: 70px;
-    }
-    &__controls {
-      gap: 1rem;
-      position: fixed;
-      top: 0.5rem;
-    }
+.narrow {
```
dustinbyrne, line 138: Can this be a container query instead? Using `.tabs` as the container, this can all flow reactively in pure CSS.

e.g. (shortened for brevity)
```scss
.tabs {
  container-type: inline-size;
}

@container (max-width: 600px) {
  // reactive styles here
}
```

dustinbyrne: Is this too bleeding edge?
https://caniuse.com/css-container-queries

ahtrotta: I originally tried to fix this when I was working on the search bar and I used the `@container` query, but I ran into problems. I don't remember what they were, but I'll give it another try.

ahtrotta: It's causing a weird rendering problem, so I think we shouldn't use the container queries:

image

dividedmind (commented):
I don't quite like how javascript-heavy this solution is. Surely if we really cannot use container queries, we can do just the bare minimum in javascript, that is, adding or removing a class on the element that detects that, and then do everything in CSS off of that.

ahtrotta:
> I don't quite like how javascript-heavy this solution is. Surely if we really cannot use container queries, we can do just the bare minimum in javascript, that is, adding or removing a class on the top level element that detects it, and then do everything in CSS off of that. We don't need to pass `isNarrow` down the hierarchy or calculate styles manually.

I refactored this so that it doesn't use as much JavaScript. Let me know what you think.

packages/components/src/pages/VsCodeExtension.vue
```diff
@@ -197,6 +202,7 @@
             v-if="!isPrecomputedSequenceDiagram && !isGiantAppMap"
             :isHighlight="filtersChanged"
             data-cy="filters-button"
+            :class="[isNarrow ? 'narrow' : '']"
```
dividedmind, line 205: Can the CSS look at the container class instead, to remove the duplication?

dividedmind (approved):
Much better, thanks!

PR #1528, opened 2024-01-10
feat: Apply saved filters

kgilpin:
## Description 

The primary change in this PR is to enable the ChatSearch page to accept and handle `savedFilters`, so that filter can be synchronized with the container environment (e.g. code editor). 

Basic "Navie" styling is also added.

The "Hide unlabeled" filter is removed from the UI, because it consumes vertical space and hasn't proven to be useful. 

AppMap uploading / sharing is removed, since this feature is no longer supported by the server or by the IDE plugins.

## Dependency of

https://github.com/getappmap/vscode-appland/pull/859


packages/components/src/pages/ChatSearch.vue
```diff
@@ -221,6 +221,15 @@ export default {
     },
   },
   methods: {
+    getAppMapState() {
+      return this.$refs.vappmap?.getState();
```
dustinbyrne, line 225: FYI, I've noticed something's up with the optional chaining operator. It should transpile correctly into the distribution bundle, but it throws an error in Storybook.

dustinbyrne (approved):

packages/components/src/pages/ChatSearch.vue
dustinbyrne: This is used in a method, so it should be fine.

kgilpin: Ok - do I need to change it or not? :-)

PR #1530, opened 2024-01-11
Removed Scoped Style from the Run Config Icon

Daniel-Warner-X:
We are starting to use scoped styles in our Vue components, but in the case of SVGs it doesn't work flawlessly. Consequently this line is causing the 'huge logo' problem...

image

As a fix I am removing the 'scoped' attribute, because it's not really needed. The run config icon should always be icon width. 

Before:

image

After:

image

### DIFF
--- a/packages/components/src/components/install-guide/ProjectPickerRow.vue
+++ b/packages/components/src/components/install-guide/ProjectPickerRow.vue
@@ -456,6 +456,11 @@ export default {
 <style lang="scss" scoped>
 $brightblue: rgba(255, 255, 255, 0.1);
 
+.run-config-icon {
+  width: 22px;
+  transform: translateY(25%);
+}
+
 .page-control-wrap {
   display: flex;
   flex-direction: row;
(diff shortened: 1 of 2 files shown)


dustinbyrne:
I added the `scoped` attribute back in, and copied the style to the appropriate location. If we leave it unscoped, future changes or removal of that page will cause this bug to regress.

kgilpin:
> I added the `scoped` attribute back in, and copied the style to the appropriate location. If we leave it unscoped, future changes or removal of that page will cause this bug to regress.
> 
> Most of our components are scoped for this reason and only a dwindling handful remain unscoped.

I assume this icon could also be converted into a component with its own embedded styling?

dustinbyrne:
It could be, if we wanted to pick some predefined icon styles/sizings. That'd certainly be good practice. In this particular instance though, I think this is fine. With both classes being scoped, we shouldn't run into any more issues.

dustinbyrne (approved):

PR #1532, opened 2024-01-11
feat: Export AppMap data as JSON

### LINKED ISSUE #1526: Download AppMap data from the AppMap diagram Export menu
Ability to download the AppMap from the AppMap itself would be pretty useful for things like the Confluence plugin - so the user who's viewing the page can download the data directly.

As I've thought about this, it would also be useful to include the filter settings in the AppMap itself, say under a filter key at the top level. This would be a way that we could make the filter settings portable with the AppMap. Suppose I am looking at an AppMap in my code editor and I've configured the filter settings. Then I click the "Export AppMap data", it opens the AppMap file in my system Finder / Explorer with the filter settings embedded. I drop this into Confluence and my filter settings will be preserved.

kgilpin:
Add an Export option that's downloads the AppMap data with the view configuration (selected tab and filters) embedded. It also restores the view configuration when the AppMap data is loaded.

Note: The view filters are applied to the data before download, so if the AppMap is pruned, the pruned-out data will not be included in the download. This is intended to enable users to create much smaller data uploads than they would otherwise be able to, if they always had to upload the entire AppMap JSON.

Screen Shot 2024-01-11 at 2 49 07 PM

https://www.loom.com/share/64b87bad09004ca0978a5f4cee8e070a

Fixes #1526 
Related https://github.com/getappmap/vscode-appland/pull/862

### DIFF
--- a/packages/components/src/components/ExportJSON.vue
+++ b/packages/components/src/components/ExportJSON.vue
@@ -2,8 +2,6 @@
   <div><slot /></div>
 </template>
 <script>
-import { serializeFilter } from '@appland/models';
-
 export default {
   name: 'v-export-json',
   props: {
--- a/packages/components/src/components/ExportJSON.vue
+++ b/packages/components/src/components/ExportJSON.vue
@@ -1,0 +1,34 @@
+<template>
+  <div><slot /></div>
+</template>
+<script>
+import { serializeFilter } from '@appland/models';
+
+export default {
+  name: 'v-export-json',
+  props: {
+    appMap: {
+      type: Object,
+      required: true,
+    },
+    viewState: {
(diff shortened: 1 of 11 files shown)


dustinbyrne (approved):

ahtrotta:
When clicking from the `Sequence Diagram`, I styled the menu to look like this:

image

When clicking from any other view, it doesn't show a menu and just exports as JSON.

kgilpin:
> When clicking from the `Sequence Diagram`, I styled the menu to look like this:
> 
> image
> 
> When clicking from any other view, it doesn't show a menu and just exports as JSON.

OK cool - good luck updating the tests :-)

PR #1534, opened 2024-01-12
ci: Add CI workflow for GitHub Actions

kgilpin:


.github/workflows/main.yml
```diff
@@ -0,0 +1,211 @@
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          # The storybook runner is going to do some Git operations that rely on access to the main branch.
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Setup Node
+        uses: getappmap/appmap-js@actions/setup-node/v1
```
dividedmind, line 76: You can also [put the action in `.github/actions` directory](https://docs.github.com/en/actions/creating-actions/about-custom-actions#choosing-a-location-for-your-action).

.github/workflows/main.yml
```diff
@@ -0,0 +1,211 @@
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Setup Node
+        uses: getappmap/appmap-js@actions/setup-node/v1
+
+      - name: Build
+        run: yarn run build
```
dividedmind, line 129: Does it really need all the packages? `scanner` shouldn't be necessary, for one...

dividedmind (approved):

.github/workflows/main.yml
kgilpin: Updated. I don't know how you knew thish, I saw no reference to this anywhere...

.github/workflows/main.yml
kgilpin: You'll find lines like this one in the CLI:

```
import { scan } from '@appland/scanner';
```

.github/workflows/main.yml
kgilpin: That must literally be the only place they mention that you can do that.

dividedmind: > Updated. I don't know how you knew this, I saw no reference to this anywhere in GitHub docs or elsewhere...

Good google-fu xD

PR #1547, opened 2024-01-17
feat: Add AppMaps context to Navie

### LINKED ISSUE #1548: Instructions for using Navie are presented to the user
I'm not sure what kinds of questions I should ask or what data is available to the AI

kgilpin:
Builds on https://github.com/getappmap/appmap-js/pull/1551 , adding Components changes (only).

Shows stats about the available AppMaps, plus "how it works" and some tips.

Screen Shot 2024-01-19 at 10 01 12 AM

If no maps are found, the user sees something like this:

Screen Shot 2024-01-18 at 2 18 15 PM

If maps are found, a bit of information is shown about that:

Screen Shot 2024-01-19 at 10 01 47 AM

## TODO

- [x] Skip the Explain step if no AppMaps match the user's question. This should be done automatically by the backend in api-services.
- [x] What should happen if the Workspace contains no AppMaps? For current behavior, see http://localhost:6006/?path=/story/pages-chatsearch-instructions--instructions-no-app-maps
- [x] If no AppMaps match the question, reset the thread id to begin a new session. This will ensure that the user's next question will re-search the AppMaps.

Fixes #1548

### DIFF
--- a/packages/components/src/components/chat-search/Instructions.vue
+++ b/packages/components/src/components/chat-search/Instructions.vue
@@ -1,0 +1,115 @@
+<template>
+  <div class="instructions">
+    <h2>Navie, the AppMap AI</h2>
+    <div>
+      <h3>How Navie Works</h3>
+      <p>
+        Each chat with Navie begins by analyzing your AppMaps. AppMaps are detailed recordings of
+        your code behavior that you create by running your app with the AppMap agent or language
+        library enabled. Naive uses the data in AppMaps (which includes dynamic information like
+        database queries and web requests) as well as snippets of your code, to provide tailored
+        responses and suggestions.
+      </p>
+      <p>
+        Navie reports how many AppMaps were reviewed and the number of relevant source files used
+        for each response. Accuracy improves as you create more diverse and up-to-date AppMaps. You
+        can accomplish this by regularly interacting with your application and recording traces.
+      </p>
+      <p>
+        <strong>Remember!</strong> The better your AppMaps match your question, the better Navie's
+        answers will be.
+      </p>
+    </div>
+    <div v-if="appmapStats">
+      <h3>Your AppMaps</h3>
+      <p v-if="appmapStats.numAppMaps">
+        You have <strong>{{ appmapStats.numAppMaps }}</strong> AppMaps in your workspace. These
+        AppMaps contain <strong>{{ appmapStats.packages.length }}</strong> packages and
+        <strong>{{ appmapStats.classes.length }}</strong> classes.
+        <span v-if="appmapStats.routes.length"
+          >Your AppMaps contain <strong>{{ appmapStats.routes.length }}</strong> HTTP server routes
+        </span>
+        <span v-if="appmapStats.tables.length">
+          <span v-if="appmapStats.routes.length">and </span>
+          <span v-else>Your AppMaps contain </span
+          ><strong>{{ appmapStats.tables.length }}</strong> SQL tables</span
+        >.
+      </p>
+      <div data-cy="no-appmaps" v-else>
+        <p>
+          ⚠️ You don't have any AppMaps in your workspace. Before you can use Navie, you'll need to
+          create some.
+        </p>
+        <p>
+          Navie uses AppMap diagrams, data, and snippets of your code code to understand your
+          software and make suggestions. Navie’s accuracy improves as you make more AppMaps.
+        </p>
+        <p>
+          Each recording method targets different scenarios and provides various levels of detail
+          and control. For detailed instructions, visit
+          <a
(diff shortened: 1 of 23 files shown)


packages/cli/appmap.yml
```diff
@@ -1,15 +1,7 @@
 name: '@appland/appmap'
```
kgilpin, line 1: Update this to the the format of appmap-node.

packages/components/src/components/chat-search/Instructions.vue
```diff
@@ -0,0 +1,108 @@
...
+    <h2>Navie, the AppMap AI</h2>
+    <div>
+      <p>
+        Navie is an AI software architect. You ask Navie questions about your code, and she provides
+        comprehensive answers about code architecture and design. You can also work with Navie to
+        design new features, and to improve the performance and security of your code.
+      </p>
+    </div>
+    <div>
+      <h3>How Navie Works</h3>
```
Daniel-Warner-X, line 12: There is so much text in the 'How Navie Works' section, the noise it generates is likely to outweigh any clarity it intends to provide. People won't read it, and they may get confused about how to proceed. Until we have a reliable way to tell them they have enough/not enough AppMap data, the simple message of 'The better your AppMaps match your question, the better Navie's answers.' Should be enough to suffice.

packages/components/src/components/chat-search/Instructions.vue
```diff
@@ -0,0 +1,108 @@
...
+        >.
+      </p>
+      <p data-cy="no-appmaps" v-else>
+        ⚠️ You don't have any AppMaps in your workspace. Before you can use Navie, you'll need to
+        create some.
+      </p>
+    </div>
+    <div class="tips">
+      <h3>Tips</h3>
+      <ul>
```
Daniel-Warner-X, line 47: The 'Tips' section conflicts with the suggestions in the left hand column. If we are unsure if people have enough AppMap data then we shouldn't be suggesting they look for Security issues, etc. These messages feel really discordant.

packages/components/src/components/chat/SuggestionGrid.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div class="suggestion-grid" v-if="suggestions.length">
-    <p>Not sure? Here are some useful starting points:</p>
+    <p>Here are some useful starting points:</p>
```
Daniel-Warner-X, line 3: I am on the verge of advocating for the removal of the prompts entirely, but while they still exist they should match the Navie metaphor a little more closely. Instead of 'starting points' it's more like 'ask Navie'. And we can drop the map pin icons as well. they distract from the primary interaction -- the text input.

Daniel-Warner-X (changes requested):
Overall the initial Navie view re-introduces a lot of complexity, mostly in the form of text, that a chat bot interface is supposed to solve. We need to pull some of the extraneous elements out.

packages/components/src/components/chat-search/Instructions.vue
kgilpin: I'm not totally getting the hate for this; I wanted to put on the page the information that users need to know, and we can iterate on the presentation.

packages/components/src/components/chat-search/Instructions.vue
kgilpin: Again, I am not trying reconcile the entire design, just responding to the issues that people raised when they were testing this UI. There are things to know about Navie that make it different from ChatGPT or Copilot. Somehow we need to get those across.

packages/components/src/components/chat/SuggestionGrid.vue
kgilpin: What's nice about the Prompts is that they give people a way to try out the AI without having to even think of a question. The Tips are designed for maybe later on, when they are asking their own questions and want to learn how to use Navie better. And, Navie is a thing that needs to be learned, like other AI tools - there are ways that it works well and ways that it doesn't work well. It's not a magic box.

packages/components/src/components/chat-search/Instructions.vue
Daniel-Warner-X: Not really 'hate' just trying to clarify the presentation.  It's your call.

kgilpin: Your suggestion is to replace all the content with `The better your AppMaps match your question, the better Navie's answers`?

kgilpin:
@dustinbyrne it looks like this indication is being shown on follow-up questions, which is incorrect?

Screen Shot 2024-01-23 at 8 39 15 PM

dustinbyrne (approved):

PR #1551, opened 2024-01-19
feat: appmap.stats RPC

kgilpin:
Provide RPC command `appmap.stats`, which includes summary information about the AppMaps available in the workspace.

This PR is the basis for https://github.com/getappmap/appmap-js/pull/1547 , and includes the RPC and CLI changes (only).

### DIFF
--- a/packages/cli/tests/integration/rpc.appmap.spec.ts
+++ b/packages/cli/tests/integration/rpc.appmap.spec.ts
@@ -12,6 +12,31 @@ describe('RPC', () => {
   afterEach(async () => await rpcTest.teardownEach());
   afterAll(async () => await rpcTest.teardownAll());
 
+  describe('appmap.stats', () => {
+    it('can be retrieved', async () => {
+      const options: AppMapRpc.StatsOptions = {};
+      const response = await rpcTest.client.request(AppMapRpc.StatsFunctionName, options);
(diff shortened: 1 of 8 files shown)


packages/cli/tests/integration/rpc.appmap.spec.ts
```diff
@@ -12,6 +12,31 @@ describe('RPC', () => {
   afterEach(async () => await rpcTest.teardownEach());
   afterAll(async () => await rpcTest.teardownAll());
 
+  describe('appmap.stats', () => {
+    it('can be retrieved', async () => {
```
kgilpin, line 16: @dividedmind one weird thing - this test case should be showing up in the AppMap CI report as "New AppMap", but it's not.

kgilpin: The main branch is also using a different revision of the appmap.yml though so, I don't know what the impact of that is. Still, I'd expect to see changes in the PR report.

dustinbyrne (approved):

packages/cli/tests/integration/rpc.appmap.spec.ts
dividedmind: Hmm. Did you check if an AppMap actually gets created?

PR #1571, opened 2024-01-26
feat: Display code selection as a separate field in Navie UI

kgilpin:
When the user initiates a search using the "light bulb", preserve that snippet in the Navie UI, and use it as context for the user's question.

## Rough example

Screen Shot 2024-01-26 at 4 40 04 PM


kgilpin:
@dustinbyrne this can be closed now, right? Because it was merged in a different PR?

PR #1587, opened 2024-02-07
fix: Propagate codeSelections

kgilpin:
Ensure that `codeSelections` are sent to the backend.

### DIFF
--- a/packages/components/src/components/chat/Chat.vue
+++ b/packages/components/src/components/chat/Chat.vue
@@ -153,7 +153,7 @@ export default {
       type: String,
     },
     sendMessage: {
-      type: Function, // UserMessageHandler
+      type: Function, // (message: string, codeSelections?: string[]) => void
     },
     suggestions: {
       type: Array,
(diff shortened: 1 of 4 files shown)


packages/components/src/components/chat/Chat.vue
```diff
@@ -153,7 +153,7 @@ export default {
       type: String,
     },
     sendMessage: {
-      type: Function, // UserMessageHandler
+      type: Function, // (message: string, codeSelections?: string[]) => void
```
dustinbyrne, line 156: For typing, you could use `Function as (message: string, codeSelections?: string[]) => void`. Though, this file is flagged with `@ts-nocheck` anyway.

I did find that we can use proper TypeScript via `Vue.extend`, e.g.
https://github.com/getappmap/appmap-js/blob/main/packages/components/src/components/chat/CodeSelection.vue#L41

dustinbyrne (approved):

packages/components/src/components/chat/Chat.vue
kgilpin: Oh interesting - thanks

PR #1588, opened 2024-02-07
Fixes for bugs found with unfulfilled-promises scanner

dividedmind:


packages/cli/src/fingerprint/fingerprintQueue.ts
```diff
@@ -23,32 +24,28 @@ export default class FingerprintQueue {
       }
       this.pending.delete(appmapFileName);
     }, this.size);
-    this.queue.pause();
+    this.queue.drain(() => (this.handler.checkVersion = false));
```
kgilpin, line 27: Changing the way that the indexer works is a bit scary to me tbh.

dividedmind: Yeah, I did quite a bit of convincing myself the change is correct. Also the test coverage is pretty good; I think it should be fine.

kgilpin (approved):

PR #1594, opened 2024-02-14
Provide quota limit feedback to users

dustinbyrne:
This is a work in progress.

**TODO:**
- [ ] Check the current quota before a chat has begun
- [ ] Add automated tests

image

It's possible that the backend does not permit any free daily messages. If so:
image

### DIFF
--- a/packages/components/src/components/chat/QuotaExceededMessage.vue
+++ b/packages/components/src/components/chat/QuotaExceededMessage.vue
@@ -1,0 +1,163 @@
+<template>
+  <div class="message">
+    <span>{{ message }}</span>
+    <div class="options">
+      <div class="option" v-if="limit > 0">
+        <div class="option__grant center">Continue on the free plan</div>
+        <p>The free plan allows you to send up to {{ limit }} messages to Navie AI per day.</p>
+        <p>
+          Please wait <b>{{ timeLeft }} </b>to send another message.
+        </p>
+      </div>
+      <div class="option">
+        <div class="option__grant center">Upgrade to AppMap Pro</div>
+        <a href="https://appmap.io/pricing" class="center" target="_blank">See details</a>
+        <p>AppMap Pro has no daily restriction on the number of daily messages sent to Navie AI.</p>
+        <p>Upgrading will instantly restore access.</p>
+        <a
+          href="https://billing.stripe.com/p/login/14kcOoc0gbrIdqMeUU"
+          class="center"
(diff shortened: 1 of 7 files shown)


packages/components/src/components/chat/QuotaExceededMessage.vue
```diff
@@ -0,0 +1,163 @@
+<template>
+  <div class="message">
+    <span>{{ message }}</span>
+    <div class="options">
+      <div class="option" v-if="limit > 0">
+        <div class="option__grant center">Continue on the free plan</div>
+        <p>The free plan allows you to send up to {{ limit }} messages to Navie AI per day.</p>
```
kgilpin, line 7: Maybe “make up to X requests” or “have up to X conversations”

packages/components/src/components/chat/QuotaExceededMessage.vue
```diff
@@ -0,0 +1,163 @@
...
+          Please wait <b>{{ timeLeft }} </b>to send another message.
+        </p>
+      </div>
+      <div class="option">
+        <div class="option__grant center">Upgrade to AppMap Pro</div>
+        <a href="https://appmap.io/pricing" class="center" target="_blank">See details</a>
+        <p>AppMap Pro has no daily restriction on the number of daily messages sent to Navie AI.</p>
+        <p>Upgrading will instantly restore access.</p>
+        <a
+          href="https://billing.stripe.com/p/login/14kcOoc0gbrIdqMeUU"
```
kgilpin, line 18: We should have a page on getappmap.com for this because they may want to choose a personal subscription or a team subscription - or even Enterprise. So maybe “Compare plans”

packages/components/src/components/chat/QuotaExceededMessage.vue
```diff
@@ -0,0 +1,163 @@
...
+          class="center"
+          target="_blank"
+        >
+          <v-button label="Subscribe" size="medium" class="subscribe" />
```
kgilpin, line 22: “Compare plans”?

kgilpin:
@dustinbyrne this is a useful design, and I think we have the data we need for this now in the ConversationThread response.

We would need to handle the case where the thread enrollment response is an error, due to e.g. proxy settings or the Navie services being down. We would not want that condition to block the use of Navie.

kgilpin:
Implemented by other means

PR #1596, opened 2024-02-20
Update Navie interface

ahtrotta:
- Update text in Navie interface before chat begins.
- Update text in instructions flow when Navie is the next step.
- Add link `Looking for code visualizations?` that opens the AppMap tree in the sidebar of the IDE.

image

image


packages/components/src/components/chat-search/Instructions.vue
```diff
@@ -67,33 +70,34 @@
...
+        Navie knows how your application works at runtime.
       </p>
+      <h3>How to use Navie</h3>
       <p>
-        Navie reports how many AppMaps were reviewed and the number of relevant source files used
-        for each response.
+        <a href="#" @click.prevent="openRecordInstructions">
+          Run your application with the AppMap recording agent
+        </a>
+        so that Nave can observe your application at runtime.
```
dustinbyrne, line 84:
```suggestion
        so that Navie can observe your application at runtime.
```

PR #1604, opened 2024-02-27
fix: Empty or invalid Navie responses

### LINKED ISSUE #1601: Server-side messages reported via tokens are not visible to the client through RPC
The socket is closed, and no message is reported to the client.

They see "failed to fetch".

### LINKED ISSUE #1602: No response is returned when search returns zero results
```
Obtained 0 results
```

No tokens are output.

dustinbyrne:
This PR fixes two issues:
1. Responses were not being streamed in if the user had no AppMap matches
2. If the user had not agreed to the latest EULA, the socket would close without notifying the user

Resolves #1601 
Resolves #1602

### DIFF
--- a/packages/cli/src/rpc/explain/explain.ts
+++ b/packages/cli/src/rpc/explain/explain.ts
@@ -150,6 +150,20 @@ async function explain(
       searchStatusByUserMessageId.set(userMessageId, status);
       first() && resolve({ userMessageId, threadId });
     });
+
+    // If the request completes here, consider it an error. This would mean that the
+    // remote server terminated our connection early.
+    explain.on(
+      'complete',
+      () =>
+        first() &&
+        reject(
+          RpcError.fromException(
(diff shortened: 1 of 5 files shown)


packages/cli/src/rpc/explain/explain.ts
```diff
@@ -150,6 +150,20 @@ async function explain(
...
+    explain.on(
+      'complete',
+      () =>
+        first() &&
+        reject(
+          RpcError.fromException(
+            new Error(status.explanation?.join('') || 'The response completed unexpectedly')
+          )
+        )
+    );
```
dustinbyrne, line 165: I could see a case made for this being it's own error message or something similar, though we're not doing anything more with these messages where there's an explicit need to differentiate.

From the servers perspective:
1. A user connects
2. The server determines the user cannot proceed
3. A response is written to the client as raw tokens
4. An `end` message is sent and the connection is terminated

kgilpin (approved):

PR #1613, opened 2024-03-01
feat: AppMap stats in the Navie view are reactive

### LINKED ISSUE #1607: Indicator of you do / don't have AppMaps is responsive to changes in the number of AppMaps
Screen Shot 2024-02-29 at 11 13 21 AM

If I make AppMaps, this information does not change. It should disappear, or be replaced, when I make AppMaps, then re-appear if I delete all my AppMaps.

dustinbyrne:
Changes to 'most recent AppMaps' will trigger a reload of AppMap stats. The RPC call is debounced to one request per second.

[Peek 2024-03-01 16-22.webm](https://github.com/getappmap/appmap-js/assets/8737782/d09ebcac-f6f2-4f63-823e-67838fa8290e)

Resolves #1607

### DIFF
--- a/packages/components/src/pages/ChatSearch.vue
+++ b/packages/components/src/pages/ChatSearch.vue
@@ -121,6 +121,8 @@ import AppMapRPC from '@/lib/AppMapRPC';
 import authenticatedClient from '@/components/mixins/authenticatedClient';
 import type { ITool, CodeSelection } from '@/components/chat/Chat.vue';
 
+import debounce from '@/lib/debounce';
+
 export default {
   name: 'v-chat-search',
   components: {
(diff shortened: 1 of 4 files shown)


ahtrotta (approved):

packages/components/src/pages/ChatSearch.vue
```diff
@@ -195,9 +197,20 @@ export default {
...
       },
       ask: undefined,
+      loadAppMapStats: debounce(async () => {
+        const rpc = this.rpcClient();
+        try {
+          this.appmapStats = await rpc.appmapStats();
+        } catch (e) {
+          console.error('Error loading appmap stats', e);
+        }
+      }, 1000),
```
kgilpin, line 207: The AppMap stats can be expensive to compute. I'd suggest a longer delay, maybe 5 seconds. Or, it should be adaptive to however long the stats took to compute the last time. Requesting stats every 1 second could be punishing for large projects.

kgilpin (changes requested):

ahtrotta:
I think it might be good if we get this merged before we ship the [new status bar in the chat](https://github.com/getappmap/appmap-js/pull/1627). Otherwise users might be stuck with a big red status bar even if they have AppMaps.

kgilpin:
> I think it might be good if we get this merged before we ship the [new status bar in the chat](https://github.com/getappmap/appmap-js/pull/1627). Otherwise users might be stuck with a big red status bar even if they have AppMaps.

The behavior of this PR is intended to be (must be) backwards compatible. Because we have no way of forcing users to upgrade.

PR #1614, opened 2024-03-03
fix: Propagate appmaps selection

kgilpin:


kgilpin:
I don't think this is needed after all. The Request to remote Navie is not responsible for propagating the appmaps selection.

PR #1616, opened 2024-03-03
feat: "navie" CLI command with local and remote options

kgilpin:
# Navie runs in "Local" or "Remote" mode

## Navie Local

* Uses local file system for storing and retrieving history (initHistory, restoreMessages).
* Manipulates data locally without external dependencies.

## Navie Remote

* Deals with external services (e.g., AI or API calls) to process requests.
* Utilizes network operations to emit tokens, acknowledgments, etc., to manage requests and responses remotely.

## Comparative Analysis

**Storage & Data Processing:** Local implements local file operations for managing history and messages, ensuring all data is processed and stored on the user's machine. In contrast, Remote relies on external services or APIs, processing data or commands over the network.

**Performance Considerations:** Local operations can be faster for tasks that solely rely on the local environment (disk I/O) without requiring network latency. Remote operations might introduce latency based on network speed and the external service's response time but allow for distributed processing and potentially more powerful computational resources.

**Use Cases:** Local might be favored for offline functionalities, data privacy concerns, or when operations must be contained within the user's environment. Remote is beneficial for leveraging cloud-based resources, integrating with online services, or performing tasks that require external data or computation not available locally.

**Scalability and Flexibility:** Remote offers a more scalable solution as it can tap into cloud-based resources dynamically, adapting to varying load by leveraging external computational power. Local is constrained by the limitations of the local environment (e.g., computational power, storage capacity).

# Context is "right sized" to the token limit

Navie calls explain:searchContext, initiating the collection of relevant context based on the input `src/rpc/explain/collectContext.ts:9`. This involves querying full-text search (`fulltext:search`) to find matching code snippets, sequence diagrams, and code objects relevant to the question.

The `collectContext` function assembles a response containing search results and a detailed context consisting of sequence diagrams, code snippets, and code objects `src/rpc/explain/collectContext.ts:9`. The size of this context and the specificity of the results are refined based on a `charLimit` parameter, ensuring that the provided context doesn't exceed the limit expected by the client request.


packages/client/src/aiClient.ts
```diff
@@ -22,7 +22,7 @@ export type Callbacks = {
   onError?: (error: Error) => void;
 };
 
-type UserInput = {
```
kgilpin, line 25: Note: Changes from here on down are present in feat/navie-cli-2.

dustinbyrne (approved):

dustinbyrne (approved):

PR #1620, opened 2024-03-05
Users can activate with an email

ahtrotta:
This PR allows users to activate the AppMap extension with just their email address. 

Initially the user will see this webview:

image

Once they enter a valid email and click `Activate with Email`, then they will see this webview:

image

Once they enter a valid verification code and click `Complete Activation`, then the IDE is activated.

### IDE Changes

The IDE needs to handle the new `activate` event that gets sent once the verification code has been submitted and verified. The event will have the `api_key` as data, which can be used to activate the IDE.


packages/components/src/components/SignIn.vue
```diff
@@ -1,62 +1,192 @@
...
+          class="btn btn-primary"
+          data-cy="gitlab-activation-button"
+          href="/"
+          @click.prevent="signIn"
+        >
+          Activate with <span><GitLabLogo class="small-logo" /> GitLab</span>
+        </a>
+        <div class="subscriber-notice">
+          <div class="pink-asterisk">*</div>
+          <p>You must activate with GitHub or GitLab if you are using an organizational license.</p>
```
dustinbyrne, line 28: Why is this the case?

packages/components/src/components/SignIn.vue
```diff
@@ -1,62 +1,192 @@
...
 <script>
 import AppMapLogo from '@/assets/appmap-full-logo.svg';
-import shieldIcon from '@/assets/shield-icon.svg';
-import VExternalLinkIcon from '@/assets/external-link.svg';
+import ShieldIcon from '@/assets/shield-icon.svg';
+import GitHubLogo from '@/assets/github-logo.svg';
+import GitLabLogo from '@/assets/gitlab-logo.svg';
+import ExternalLinkIcon from '@/assets/external-link.svg';
+
+const APPMAP_SERVER_URL = 'https://getappmap.com';
```
dustinbyrne, line 116: This should ideally be a prop, considering it's a setting in VS Code

packages/components/src/components/SignIn.vue
ahtrotta: I'm not sure, but Elizabeth mentioned it [here](https://docs.google.com/document/d/1k9RUstEYu1aDz1gFJgSe2Lav-Ls50BVOKu2DxpF1Sd4/edit?disco=AAABIeZDiUQ).

packages/components/src/components/SignIn.vue
ahtrotta: Good call, I changed it :+1:

packages/components/src/components/SignIn.vue
ahtrotta: This was removed

packages/components/src/components/SignIn.vue
```diff
@@ -42,21 +120,108 @@
...
+        try {
+          const data = await response.json();
+          this.$root.$emit('activate', data.api_key);
+        } catch (error) {
+          this.submitted = false;
+          this.error = GENERIC_ERROR_MSG;
+        }
+      } else {
+        this.verificationCode = '';
+        this.submitted = false;
```
dustinbyrne, line 212: If the user inputs the verification code incorrectly then by changing this flag we're forcing them to request another as there's no way to get back to the second screen. Can we present the error on the second screen? It looks like there's already a 'try again' button if they need to go back.

ahtrotta: Yeah, good point, I made it so that the error is on the second screen and it doesn't go back to the first screen :+1:

dustinbyrne (approved):

PR #1622, opened 2024-03-07
Use activation terminology in Navie

ahtrotta:
If a user accesses Navie and they haven't activated their IDE (which is probably quite uncommon), then we want to show them messaging that is consistent with the new `Activate` terminology.

### DIFF
--- a/packages/components/src/components/chat/Chat.vue
+++ b/packages/components/src/components/chat/Chat.vue
@@ -45,15 +45,15 @@
     </div>
     <div v-if="!authorized" class="status-unauthorized status-container">
       <div class="status-label">
-        <p>You must be logged in to AppMap to use this feature.</p>
+        <p>You must have activated AppMap to use this feature.</p>
         <p>
           <b>VSCode</b>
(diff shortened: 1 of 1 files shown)


packages/components/src/components/chat/Chat.vue
```diff
@@ -45,15 +45,15 @@
     </div>
     <div v-if="!authorized" class="status-unauthorized status-container">
       <div class="status-label">
-        <p>You must be logged in to AppMap to use this feature.</p>
+        <p>You must have activated AppMap to use this feature.</p>
         <p>
           <b>VSCode</b>
-          To login with VSCode, run the command <tt>AppMap: Login</tt>, or click the Sign In link
-          from the AppMap panel.
+          To activate AppMap for VSCode, run the command <tt>AppMap: Login</tt>, or activate using
```
kgilpin, line 51: Hmm, what does AppMap: Login do? Maybe we can just point them to the sidebar, or is there a command to show/reveal the sidebar?

PR #1623, opened 2024-03-07
fix: Add missing dynamic import

dustinbyrne:
Resolves the following error in the binary distribution
```
Error: Cannot find module '/snapshot/appmap-js/node_modules/openai/_shims/auto/runtime-node.js'
```

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -149,7 +149,9 @@
       "node18-macos-arm64"
     ],
     "scripts": [
-      "built/cmds/record/state/record_*.js"
+      "built/cmds/record/state/record_*.js",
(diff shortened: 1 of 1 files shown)


packages/cli/package.json
```diff
@@ -149,7 +149,9 @@
       "node18-macos-arm64"
     ],
     "scripts": [
-      "built/cmds/record/state/record_*.js"
+      "built/cmds/record/state/record_*.js",
+      "node_modules/openai/_shims/auto/runtime-node.js",
+      "../../node_modules/openai/_shims/auto/runtime-node.js"
```
kgilpin, line 154: Wow this is crazy.

PR #1624, opened 2024-03-07
Test the native binary for @appland/appmap

dustinbyrne:
This isn't as extensive as it used to be, but it's enough to catch and prevent any missing module issues in the future, and lays the groundwork to expand tests via Jest.


packages/cli/package.json
```diff
@@ -18,7 +18,7 @@
     "lint:fix": "eslint 'src/**/*.js' 'tests/**/*.js' --fix",
     "pre-commit": "lint-staged",
     "test": "jest --filter=./tests/testFilter.js",
-    "jest": "jest --filter=./tests/testFilter.js",
```
dustinbyrne, line 21: Providing a `jest` script prevents us from running `yarn jest` without the args specified here.

PR #1625, opened 2024-03-10
feat: RPC service requires configuration with directory list

kgilpin:
A global setting for the process controls which data is going to be examined and processed by the RPC functions, including Search and Navie.

## Design doc

https://docs.google.com/document/d/1bRlYO8BVe2zQni12fQBNJnUih66yu7t166fh787om18/edit

### DIFF
--- a/packages/cli/src/cmds/index/index.ts
+++ b/packages/cli/src/cmds/index/index.ts
@@ -4,7 +4,7 @@ import chalk from 'chalk';
 
 import FingerprintDirectoryCommand from '../../fingerprint/fingerprintDirectoryCommand';
 import FingerprintWatchCommand from '../../fingerprint/fingerprintWatchCommand';
-import { handleWorkingDirectory } from '../../lib/handleWorkingDirectory';
+import { configureRpcDirectories, handleWorkingDirectory } from '../../lib/handleWorkingDirectory';
 import { locateAppMapDir } from '../../lib/locateAppMapDir';
 import { verbose } from '../../utils';
 import { log, warn } from 'console';
@@ -138,6 +138,8 @@ export const handler = async (argv) => {
       ) => new RemoteNavie(threadId, contextProvider, projectInfoProvider);
 
       const navieProvider = useLocalNavie() ? buildLocalNavie : buildRemoteNavie;
+
+      configureRpcDirectories([process.cwd()]);
 
       const rpcMethods: RpcHandler<any, any>[] = [
         numProcessed(cmd),
--- a/packages/cli/src/cmds/index/index.ts
+++ b/packages/cli/src/cmds/index/index.ts
@@ -18,7 +18,7 @@ import { explainHandler, explainStatusHandler } from '../../rpc/explain/explain'
 import RPCServer from './rpcServer';
 import appmapData from '../../rpc/appmap/data';
 import { loadConfiguration } from '@appland/client';
-import appmapStats from '../../rpc/appmap/stats';
+import { appmapStatsV1 } from '../../rpc/appmap/stats';
 import LocalNavie from '../../rpc/explain/navie/navie-local';
 import RemoteNavie from '../../rpc/explain/navie/navie-remote';
 import { Context, ProjectInfo } from '@appland/navie';
@@ -142,7 +142,7 @@ export const handler = async (argv) => {
       const rpcMethods: RpcHandler<any, any>[] = [
         numProcessed(cmd),
         search(),
-        appmapStats(),
+        appmapStatsV1(),
         appmapFilter(),
         appmapData(),
         metadata(),
--- a/packages/cli/src/cmds/index/index.ts
+++ b/packages/cli/src/cmds/index/index.ts
@@ -141,13 +141,13 @@ export const handler = async (argv) => {
 
       const rpcMethods: RpcHandler<any, any>[] = [
         numProcessed(cmd),
-        search(appmapDir),
-        appmapStats(appmapDir),
+        search(),
+        appmapStats(),
         appmapFilter(),
         appmapData(),
         metadata(),
         sequenceDiagram(),
-        explainHandler(navieProvider, appmapDir),
+        explainHandler(navieProvider),
         explainStatusHandler(),
       ];
       const rpcServer = new RPCServer(port, rpcMethods);
--- a/packages/cli/src/cmds/index/rpc.ts
+++ b/packages/cli/src/cmds/index/rpc.ts
@@ -18,6 +18,7 @@ import { Context, ProjectInfo } from '@appland/navie';
 import { InteractionEvent } from '@appland/navie/dist/interaction-history';
 import { configureRpcDirectories } from '../../lib/handleWorkingDirectory';
 import { loadConfiguration } from '@appland/client';
+import { getConfiguration, setConfiguration } from '../../rpc/configuration';
 
 const AI_KEY_ENV_VARS = ['OPENAI_API_KEY'];
 
@@ -134,6 +135,8 @@ export const handler = async (argv) => {
     sequenceDiagram(),
     explainHandler(navieProvider),
     explainStatusHandler(),
+    setConfiguration(),
+    getConfiguration(),
   ];
   const rpcServer = new RPCServer(port, rpcMethods);
   rpcServer.start();
--- a/packages/cli/src/cmds/index/rpc.ts
+++ b/packages/cli/src/cmds/index/rpc.ts
@@ -11,14 +11,14 @@ import sequenceDiagram from '../../rpc/appmap/sequenceDiagram';
 import { explainHandler, explainStatusHandler } from '../../rpc/explain/explain';
 import RPCServer from './rpcServer';
 import appmapData from '../../rpc/appmap/data';
-import appmapStats from '../../rpc/appmap/stats';
+import { appmapStatsV1, appmapStatsV2 } from '../../rpc/appmap/stats';
 import LocalNavie from '../../rpc/explain/navie/navie-local';
 import RemoteNavie from '../../rpc/explain/navie/navie-remote';
 import { Context, ProjectInfo } from '@appland/navie';
 import { InteractionEvent } from '@appland/navie/dist/interaction-history';
 import { configureRpcDirectories } from '../../lib/handleWorkingDirectory';
 import { loadConfiguration } from '@appland/client';
-import { getConfiguration, setConfiguration } from '../../rpc/configuration';
+import { getConfigurationV1, setConfigurationV1 } from '../../rpc/configuration';
 
 const AI_KEY_ENV_VARS = ['OPENAI_API_KEY'];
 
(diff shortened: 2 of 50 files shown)


kgilpin:
@dustinbyrne thanks for working on this. 

One thing I was thinking , we may need to include the directory on each search result so that we can resolve the source file. I don’t know if you’ve run into that? I can work on it tomorrow if not.

dustinbyrne:
Hmm, it's possible. The manner in which I was running this locally I don't think I would have noticed.

dustinbyrne:
It looks like I messed up a rebase. 

f12090ded5f35d24f278dc7f7c813bf49ccb847a contains changes that should have overwritten 1f91cef4abf55bcc57185a8fc9059cb202aab00e.

I'll fix that up before this branch is merged, but I'm not planning on force pushing while this branch is active.

packages/rpc/src/configuration.ts
```diff
@@ -0,0 +1,8 @@
+export namespace ConfigurationRpc {
+  export const SetFunctionName = 'configuration.set';
+  export const GetFunctionName = 'configuration.get';
+
+  export type Configuration = {
```
kgilpin, line 5: One thing we have to watch out for @dustinbyrne is that the backend is going to **require** the Configuration to be set before any of the methods will have a sane behavior. Otherwise everything is going to return empty results.

kgilpin: So the RPC process should be launched as early as possible as the AppMap extension is initializing, and the RPC configuration must be kept up to date. For example, updated when appmap.yml add/remove events are handled, and we may want to back that up with a polling process that also updates the config every 10 seconds or so. Just to make sure we don't miss anything. Events are nice when they work, but they aren't 100% reliable so it's important to have a polling backup.

packages/cli/src/cmds/index/index.ts
```diff
@@ -141,13 +141,13 @@ export const handler = async (argv) => {
 
       const rpcMethods: RpcHandler<any, any>[] = [
```
kgilpin, line 142: @dustinbyrne I think what we need to do here is set the Configuration from the CLI parameters. That way, older versions of the extension that are still launching `index -p -w` (an indexer + RPC per-process) will continue to work. We can do that until everyone in the wild has upgraded beyond extension versions that still do that.

packages/cli/src/cmds/index/rpc.ts
```diff
@@ -0,0 +1,140 @@
...
+import LocalNavie from '../../rpc/explain/navie/navie-local';
+import RemoteNavie from '../../rpc/explain/navie/navie-remote';
+import { Context, ProjectInfo } from '@appland/navie';
+import { InteractionEvent } from '@appland/navie/dist/interaction-history';
+import { configureRpcDirectories } from '../../lib/handleWorkingDirectory';
+import { loadConfiguration } from '@appland/client';
+
+const AI_KEY_ENV_VARS = ['OPENAI_API_KEY'];
+
+export const command = 'rpc';
```
kgilpin, line 24: Only new up-to-date versions of the IDE extension will invoke this command.

packages/cli/src/cmds/index/rpc.ts
```diff
@@ -0,0 +1,140 @@
...
+    contextProvider: Context.ContextProvider,
+    projectInfoProvider: ProjectInfo.ProjectInfoProvider
+  ) => {
+    loadConfiguration(false);
+    return new RemoteNavie(threadId, contextProvider, projectInfoProvider);
+  };
+
+  const navieProvider = useLocalNavie() ? buildLocalNavie : buildRemoteNavie;
+
+  configureRpcDirectories(directories);
```
kgilpin, line 126: We should also do this in the `index` command, to keep it compatible.

packages/cli/src/cmds/navie/projectInfo.ts
```diff
@@ -0,0 +1,31 @@
+import { ProjectInfo } from '@appland/navie';
+import { appmapStatsHandler } from '../../rpc/appmap/stats';
+import configuration from '../../rpc/configuration';
+
+export default async function collectProjectInfos(): Promise<ProjectInfo.ProjectInfo[]> {
+  const projectInfoByPath: { [key: string]: ProjectInfo.ProjectInfo } = {};
+  // KEG: I'm not sure the point of this, since the appmapStatsHandler will also check the configuration.
```
kgilpin, line 7: This comment can be removed, I get it now.

packages/cli/src/fulltext/AppMapIndex.ts
```diff
@@ -72,8 +76,12 @@ async function buildDocument(metadataFile: string): Promise<any> {
     });
   }
 
+  let appmapId = indexDir;
+  if (appmapId.startsWith(directory)) appmapId = appmapId.substring(directory.length + 1);
+
+  const id = packRef(directory, appmapId);
```
kgilpin, line 82: Store both the directory and the full appmap path in the `id`. This seems to be the best/only way to have non-indexed metadata fields.

packages/cli/src/lib/handleWorkingDirectory.ts
```diff
@@ -1,3 +1,12 @@
+import { join } from 'path';
+import { setConfiguration } from '../rpc/configuration';
+
 export function handleWorkingDirectory(directory?: string) {
   if (directory) process.chdir(directory);
 }
+
+export function configureRpcDirectories(directory: string | string[]) {
```
kgilpin, line 8: I know I have said this a lot; this should be called from the `index` command also.

packages/cli/src/rpc/appmap/stats.ts
```diff
@@ -25,10 +32,11 @@ export async function appmapStatsHandler(appmapDir: string): Promise<AppMapRpc.S
     }
   );
   await processNamedFiles(appmapDir, 'canonical.sqlTables.json', collectStrings(tables));
-  let numAppMaps = 0;
   await processNamedFiles(appmapDir, 'metadata.json', async () => numAppMaps++);
 
   return {
+    name: config.name,
+    path: config.path,
```
kgilpin, line 39: Is `path` being removed / not reported? Is `path` actually `directory`? Consider renaming it.

packages/cli/src/rpc/configuration.ts
```diff
@@ -0,0 +1,62 @@
+import { ConfigurationRpc } from '@appland/rpc';
+import { RpcHandler } from './rpc';
+import { dirname, join } from 'path';
+import loadAppMapConfig, { AppMapConfig } from '../lib/loadAppMapConfig';
+
+// KEG: What is the extra 'path' argument for? Isn't that provided by Configuration.directories?
```
kgilpin, line 6: Can be removed.

packages/cli/src/rpc/explain/buildContext.ts
```diff
@@ -60,7 +60,7 @@ export default async function buildContext(searchResults: SearchRpc.SearchResult
 
       if (codeSnippets.has(event.location)) continue;
 
-      const snippets = await lookupSourceCode(event.location);
+      const snippets = await lookupSourceCode(result.directory, event.location);
```
kgilpin, line 63: The directory argument here is crucial because it's how the source file path (which is a relative path) can be resolved. You may want to dig into `lookupSourceCode` and satisfy yourself that it's doing the right thing.

packages/cli/src/rpc/explain/lookupSourceCode.ts
```diff
@@ -23,15 +23,15 @@ let FILE_NAMES = new Set<string>();
...
+export default async function lookupSourceCode(directory: string, location: string): Promise<string[] | undefined> {
   if (verbose()) warn(chalk.gray(`Looking up source code for ${location}`));
   const [requestedFileName, lineNoStr] = location.split(':');
 
   const extension = requestedFileName.slice(requestedFileName.lastIndexOf('.'));
 
   if (!scannedExtensions.has(extension)) {
     scannedExtensions.add(extension);
-    const fileNames = await promisify(glob)(`**/*${extension}`);
+    const fileNames = await promisify(glob)(`${directory}/**/*${extension}`);
```
kgilpin, line 34: Here's that `directory`.

packages/cli/src/rpc/search/search.ts
```diff
@@ -15,13 +18,29 @@ export type HandlerOptions = {
...
+  const config = configuration();
+  const { directories } = config;
   let appmapSearchResponse: SearchResponse;
   if (appmaps) {
+    const results = appmaps
+      .map((appmap) => {
+        let directory: string | undefined;
+        if (directories.length === 1) directory = directories[0];
+        else directory = directories.find((dir) => appmap.startsWith(dir));
+        if (!directory) return undefined;
```
kgilpin, line 34: This is a bit awkward. We have to choose a directory to report for the appmap. Make some guesses / defaults.

packages/cli/src/rpc/search/search.ts
```diff
@@ -50,8 +69,12 @@ export async function handler(
     };
     const eventsSearchResponse = await searchSingleAppMap(result.appmap, query, searchOptions);
     results.push({
-      appmap: result.appmap,
-      events: eventsSearchResponse.results,
+      appmap: isAbsolute(result.appmap) ? result.appmap : join(result.directory, result.appmap),
```
kgilpin, line 72: Always report `appmap` as absolute path. This will make it easier in other systems to resolve it to a file.

kgilpin: Is this going to break in like a dev container scenario? I guess we can refit later if needed.

packages/cli/src/cmds/index/index.ts
kgilpin: This is for testing and backwards compatibility. 

The IDE extensions should launch without the directory argument and set configuration as soon as they have the port number.

packages/cli/src/cmds/navie/navie.ts
```diff
@@ -1,15 +1,15 @@
...
-import { handleWorkingDirectory } from '../../lib/handleWorkingDirectory';
-import loadAppMapConfig from '../../lib/loadAppMapConfig';
+import { configureRpcDirectories, handleWorkingDirectory } from '../../lib/handleWorkingDirectory';
 import collectContext from '../../rpc/explain/collectContext';
-import { appmapStatsHandler } from '../../rpc/appmap/stats';
+import configuration from '../../rpc/configuration';
+import collectProjectInfos from './projectInfo';
 
 export const command = 'navie [question]';
 export const describe = 'Generate a test case using Navie AI';
```
kgilpin, line 15: I've noticed that this description is out of date.

packages/cli/src/cmds/navie/navie.ts
```diff
@@ -134,9 +125,7 @@ function buildExplainRequest(
 export const handler = async (argv) => {
   verbose(argv.verbose);
   handleWorkingDirectory(argv.directory);
```
kgilpin, line 127: Changing the working directory shouldn't be needed any more. We may want to update this navie one-shot CLI command to accept a directory list and propagate that via `configureRpcDirectories`.

packages/cli/src/cmds/search/search.ts
```diff
@@ -169,8 +168,7 @@ export const handler = async (argv: any) => {
     const options: SearchOptions = {
       maxResults,
     };
-    const appmapDir = await locateAppMapDir();
-    const response = await AppMapIndex.search(appmapDir, query, options);
+    const response = await AppMapIndex.search([process.cwd()], query, options);
```
kgilpin, line 171: It seems wrong that `configuration.directories` isn't respected here? Is this RPC function used by the frontend?

dustinbyrne: The `search` RPC method isn't called directly from the frontend as far as I'm aware, only via `explain`. In that context, it does look like it's using the correct directories.

It also looks correct when calling the `search` RPC method
https://github.com/getappmap/appmap-js/blob/eb4b88c049a040e395455ee0bb0d24730b9ec452/packages/cli/src/rpc/search/search.ts#L60

dustinbyrne (approved):

PR #1627, opened 2024-03-11
Move Navie context status into chat

### LINKED ISSUE #1626: Status indicators are moved from right to left
The Navie 'explainer' ("Introducing Navie...") text can be removed and replaced with the status indicators present on the RHS.

Along with this:
- The status indicator boxes should collapse by default to only show the top line message
- Colored indicators should be replaced with icons

ahtrotta:
Fixes #1626 

Moves the Navie status indicator from the right-hand panel to the left-hand chat panel.

### Screenshots

If the user has AppMaps, then they see this before they begin chatting:

image

The status bar can be expanded:

image

Once they start chatting then the status bar goes away:

image

If the user does not have AppMaps, then they see this before they start chatting:

image

The status bar can be expanded:

image

Once they start chatting then the status bar stays:

image


packages/components/src/components/chat-search/ContextStatus.vue
```diff
@@ -0,0 +1,209 @@
+<template>
+  <v-status-bar v-if="appmapStats && appmapStats.numAppMaps" level="info">
```
dustinbyrne, line 2: `appmapStats` will be updated soon to be [`Stats | Stats[]`](https://github.com/getappmap/appmap-js/blob/350c84aa8fa5b5282224e68a6ab7291b73b070d0/packages/rpc/src/appmap.ts#L10-L19). It's probably worth adding support for this.

packages/components/src/components/chat-search/ContextStatus.vue
```diff
@@ -0,0 +1,209 @@
...
+          </a>
+          for Navie to know about your project.
+        </p>
+      </div>
+    </template>
+  </v-status-bar>
+</template>
+
+<script lang="ts">
+// @ts-nocheck
```
dustinbyrne, line 71: I've been trying not to do this because it completely hides compiler errors.

dustinbyrne:
The code looks good. I'm going to hold off on an approval until after this has been demo'd tomorrow.

packages/components/src/components/chat-search/ContextStatus.vue
```diff
@@ -0,0 +1,209 @@
+<template>
+  <v-status-bar v-if="appmapStats && appmapStats.numAppMaps" level="info">
+    <template v-slot:header>
+      <div class="context__header">Navie is aware of your project</div>
```
kgilpin, line 4: I think we should iterate a bit on the "is aware" / "is not aware" message. It doesn't quite seem evident to me.

packages/components/src/components/chat-search/ContextStatus.vue
kgilpin: @dustinbyrne is this the time to introduce the "front end version" flag and have the backend be responsive to this flag? This way, the frontend can declare what features it can handle. In this case, `Stats` would be the old behavior and `Stats[]` (even if it's an array with 1 element) would be the new behavior.

packages/components/src/components/chat-search/ContextStatus.vue
ahtrotta: Cool, I deleted this line :+1:

packages/components/src/components/chat-search/ContextStatus.vue
ahtrotta: I changed it based on our conversation in Slack and updated the screenshots.

packages/components/src/components/chat-search/ContextStatus.vue
```diff
@@ -0,0 +1,208 @@
...
+      <div class="context__sub-header">
+        <div class="context__table">
+          <div class="context__table__header">Navie's context:</div>
+          <div class="context__table__row">
+            <v-fail-icon class="fail-icon" data-cy="fail-icon" />
+            <div>Runtime traces (AppMap recordings)</div>
+          </div>
+          <div class="context__table__row">
+            <v-fail-icon class="fail-icon" data-cy="fail-icon" />
+            <div>Source code snippets</div>
```
kgilpin, line 35: The other type of information that we refer to in the context is "Data requests", which is a grab bag for I/O information such as SQL, HTTP, etc. I think it would be worth including that in this status as well.

packages/components/src/components/chat-search/ContextStatus.vue
```diff
@@ -0,0 +1,208 @@
...
+          </div>
+        </div>
+        <p class="mt-0">
+          You can still ask Navie questions, but Navie does not have access to your project's data.
+        </p>
+        <template v-if="!appmapYmlPresent">
+          <p class="mb-0 mt-0" data-cy="status-no-config">
```
kgilpin, line 42: Absence of appmap.yml doesn't mean that the user hasn't followed the instructions. For example, with Node.js, all they have to do is `npx appmap-node <cmd>`. Sending them back to the instructions won't fix the problem of missing appmap.yml. What they need to do is run their app using a recording method. We've made sure that all our recording methods will generate and use a default appmap.yml. So, I don't think "missing appmap.yml" is really a state that we can give specific treatment to.

packages/components/src/components/chat-search/ContextStatus.vue
```diff
@@ -0,0 +1,208 @@
...
+        <p
+          v-else-if="appmapStats && appmapStats.numAppMaps === 0"
+          class="mt-0 mb-0"
+          data-cy="status-no-data"
+        >
+          You need to
+          <a href="#" @click.prevent="openRecordInstructions" data-cy="record-guide">
+            record runtime data
+          </a>
+          for Navie to know about your project.
```
kgilpin, line 63: Two options seem useful here:

1) A button or link to return to the recording instructions.
2) A sample prompt to ask Navie how to make AppMaps (let her take it from there).

packages/components/src/components/chat-search/ContextStatus.vue
kgilpin: In other words, I think we can drop this section.

packages/components/src/components/chat-search/StatusBar.vue
```diff
@@ -0,0 +1,139 @@
+<template>
+  <div :class="['status-bar', `status-bar--${level}`]" data-cy="status-bar">
+    <div class="status-bar__header" data-cy="status-bar-header" @click="toggleShowBody">
+      <span class="status-bar__header__title">
+        <v-warning-icon class="warning-icon" v-if="level === 'warning' || level === 'error'" />
+        <v-success-icon class="success-icon" v-if="level === 'success'" />
+        <slot name="header" />
+      </span>
+      <v-chevron-down-icon :class="['toggle-icon', showBody ? 'down' : 'up']" />
```
dustinbyrne, line 9: I feel like the collapsed chevron is reversed. I think it should point down or right.

packages/components/src/components/chat-search/StatusBar.vue
```diff
@@ -0,0 +1,139 @@
...
+<style lang="scss" scoped>
+$fg: #ececec;
+
+.status-bar {
+  align-items: center;
+  color: $fg;
+  padding: 0.75rem;
+  transition: background-color 0.2s;
+  border: 1px solid rgba(255, 255, 255, 0.1);
+  box-shadow: 0 0 8px rgba(255, 255, 255, 0.15);
```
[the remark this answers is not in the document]
ahtrotta: I toned it down a bunch, so it's pretty subtle now.

packages/components/src/components/chat-search/StatusBar.vue
ahtrotta: Good point, I fixed this

packages/components/src/components/chat-search/ContextStatus.vue
dustinbyrne: I'm going to add this to #1625

packages/components/src/components/chat-search/ContextStatus.vue
```diff
@@ -0,0 +1,197 @@
+<template>
+  <v-status-bar v-if="appmapStats && appmapStats.numAppMaps" level="success">
+    <template v-slot:header>
+      <div class="context__header">{{ appmapStats.numAppMaps }} AppMaps available</div>
```
dustinbyrne, line 4: Nitpick - handle the non-plural form
```suggestion
      <div class="context__header">
        {{ appmapStats.numAppMaps }} AppMap{{ appmapStats.numAppMaps === 1 ? '' : 's' }} available
      </div>
```

dustinbyrne (approved):

PR #1628, opened 2024-03-13
Better whitespace handling in sign-in forms

ahtrotta:
Ensure that leading and trailing whitespace is removed from strings during email verification flow.

### DIFF
--- a/packages/components/src/components/SignIn.vue
+++ b/packages/components/src/components/SignIn.vue
@@ -175,9 +175,12 @@ export default {
       this.verificationCode = '';
       this.submitted = false;
     },
+    trimWhitespace(stringToTrim) {
+      return String(stringToTrim).trim();
+    },
     async activateWithEmail() {
       const url = new URL('/api/activations', this.appmapServerUrl);
-      this.addParamsToUrl(url, { email: this.email });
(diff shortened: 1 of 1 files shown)


kgilpin:
It’s fine to do this here but essential to handle it in appmap-server as well. Then no front end update (which takes a long time to propagate) would be needed.

packages/components/src/components/SignIn.vue
```diff
@@ -175,9 +175,20 @@ export default {
...
     },
+    trimWhitespace(stringToTrim) {
+      let result = stringToTrim;
+      if (typeof stringToTrim !== 'string') return '';
+      try {
+        result = stringToTrim.trim();
+      } catch (error) {
+        // fall back to original string if trimming fails
+        console.error('Error trimming string:', error);
+      }
```
dustinbyrne, line 186: According to MDN there's no exception case here:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim

dustinbyrne (approved):

packages/components/src/components/SignIn.vue
ahtrotta: Yeah, I was just being extra cautious, but it does seem unnecessary in retrospect.

PR #1632, opened 2024-03-19
feat: Migrate ChatSearch to v2 stats API

dustinbyrne:
This PR modifies the ChatSearch frontend to request stats from `v2.appmap.stats` instead of the v1 method. It aggregates the response to fit within the current frontend.

### DIFF
--- a/packages/components/src/lib/AppMapRPC.ts
+++ b/packages/components/src/lib/AppMapRPC.ts
@@ -1,3 +1,4 @@
+import { AppMapRpc } from '@appland/rpc';
 import { browserClient, reportError } from './RPC';
 
 import { EventEmitter } from 'events';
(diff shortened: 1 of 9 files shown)


packages/components/src/lib/AppMapRPC.ts
```diff
@@ -117,11 +118,15 @@ export default class AppMapRPC {
 
   appmapStats(): Promise<any> {
     return new Promise((resolve, reject) => {
-      this.client.request('appmap.stats', {}, function (err: any, error: any, result: any) {
-        if (err || error) return reportError(reject, err, error);
+      this.client.request(
+        AppMapRpc.Stats.V2.Method,
```
dustinbyrne, line 122: Should we consider this a breaking change due to reliance on the v2 API? The index process doesn't bind the v2 stats handler (though maybe it should).

kgilpin: But the index process also expects "one process per appmap.yml" so v2 stats isn't really meaningful in that scenario.

PR #1633, opened 2024-03-19
Update filter menu and increase contrast

ahtrotta:
* Make scrollbars a lighter shade of grey so it has better contrast.
* Make the zoom sliders have better contrast.
* Always show the "X" close icon for the filter menu.
* When hiding a code object with a long name, don't extend the width of the filter menu.

## Before

image

## After

image

### DIFF
--- a/packages/components/src/components/FilterMenu.vue
+++ b/packages/components/src/components/FilterMenu.vue
@@ -537,7 +537,7 @@ export default {
     border-bottom: 1px solid $gray2;
     display: flex;
     justify-content: flex-start;
-    align-items: center;
+    align-items: baseline;
 
     svg {
       width: 1em;
@@ -588,12 +588,6 @@ export default {
       margin-left: 2rem;
       svg {
         fill: white;
-      }
-    }
-
-    @media (min-width: 900px) {
-      &-close {
-        display: none;
       }
     }
   }
--- a/packages/components/src/components/FilterMenu.vue
+++ b/packages/components/src/components/FilterMenu.vue
@@ -85,14 +85,16 @@
               :suggestions="hideNamesSuggestions"
               suggestions-placement="top"
             />
-            <div class="filters__hide" v-if="hiddenNames.length">
-              <div class="filters__hide-item" v-for="(name, index) in hiddenNames" :key="name">
-                {{ name }}
-                <CloseThinIcon
-                  class="filters__hide-item-icon"
-                  @click.stop="removeHiddenName(index)"
-                />
-              </div>
+          </div>
+        </div>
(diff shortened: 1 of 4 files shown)


packages/components/src/components/FilterMenu.vue
```diff
@@ -85,14 +85,16 @@
...
-                  class="filters__hide-item-icon"
-                  @click.stop="removeHiddenName(index)"
-                />
-              </div>
+          </div>
+        </div>
+        <div class="filters__block-row">
+          <div class="filters__hide" v-if="hiddenNames.length">
+            <div class="filters__hide-item" v-for="(name, index) in hiddenNames" :key="name">
+              {{ name }}
```
kgilpin, line 93: Should these clip with like an ellipsis instead of wrapping? Maybe a tool tip for the full content? 

Please try with very long code object names a la Java. 

One other trick in Java is to shorten the package names to just the first letter of each package.

dustinbyrne:
These changes aside, I'm not certain an 'X' to close the filters will fix the issue we're seeing in JetBrains. It's almost like all pointer events are disabled.

The 'X' is still welcome, I'm just not certain it's a silver bullet.

ahtrotta:
> These changes aside, I'm not certain an 'X' to close the filters will fix the issue we're seeing in JetBrains. It's almost like all pointer events are disabled.
> 
> The 'X' is still welcome, I'm just not certain it's a silver bullet.

Have we been able to reproduce it?

kgilpin:
> These changes aside, I'm not certain an 'X' to close the filters will fix the issue we're seeing in JetBrains. It's almost like all pointer events are disabled.
> 
> The 'X' is still welcome, I'm just not certain it's a silver bullet.

So what doesn’t work? Because clicking on diagram nodes does work.

packages/components/src/components/FilterMenu.vue
kgilpin: Ok. The problem is that the “Save Filter” content can be pushed out of the viewport and then it can’t be accessed.

ahtrotta: That's strange, I can't reproduce that problem. I can always access the bottom of the menu using the scrollbar, even if I try extremely small viewports. Do you have steps to reproduce?

kgilpin:
Mobile phone :-)

packages/components/src/components/FilterMenu.vue
kgilpin: I think I’ve only seen it on mobile so, not a priority.

dustinbyrne (approved):

PR #1636, opened 2024-03-20
Open Navie from an appmap

ahtrotta:
Fixes getappmap/board#316

This PR and it's [VS Code counterpart](https://github.com/getappmap/vscode-appland/pull/920) allow a user to open Navie from within an AppMap and then chat with Navie using that AppMap as the context.

## Changes to AppMaps

In an AppMap, there are three clickable elements that will open Navie with the AppMap as context:

1. There is a small compass icon in the control bar.
2. When the details panel is collapsed, there is a compass icon.
3. When the details panel is open, there is an `Ask Navie` button.

image

image

## Changes to Navie

When Navie is opened with a single AppMap as context:

* The AppMap is rendered from the start in the right panel.
* There is text describing that the user is asking Navie about a single AppMap.
* There is a button that allows the user to switch to using the full workspace as context.
* The placeholder is `What do you want to know about this AppMap?`.
* The status bar is not shown.

image

### DIFF
--- a/packages/components/src/assets/compass-simpler.svg
+++ b/packages/components/src/assets/compass-simpler.svg
@@ -1,0 +1,11 @@
+<svg viewBox="0 0 86 85" fill="none" xmlns="http://www.w3.org/2000/svg">
+  <circle cx="43.5" cy="42.5" r="30" stroke="white" stroke-width="3"/>
+  <path d="M43 0L49.0622 42.75H36.9378L43 0Z" fill="white"/>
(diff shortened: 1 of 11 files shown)


kgilpin (approved):
Looks great! The only change I would like is to dim the placeholder a bit.

ahtrotta:
> Looks great! The only change I would like is to dim the placeholder a bit.

Cool! I dimmed the placeholder:

image

PR #1637, opened 2024-03-20
fix: Navie fixes

kgilpin:
See commits for details.


packages/cli/src/rpc/explain/buildContext.ts
```diff
@@ -48,7 +49,13 @@ export default async function buildContext(searchResults: SearchRpc.SearchResult
 
   const examinedLocations = new Set<string>();
   for (const result of searchResults) {
-    await buildSequenceDiagram(result);
+    try {
+      await buildSequenceDiagram(result);
+    } catch (e) {
+      warn(`Failed to build sequence diagram for ${result.appmap}`)
+      warn(e);
+      continue;
```
ahtrotta, line 57: The `continue` is unnecessary since the execution would continue normally after the `catch` block anyway.

ahtrotta (approved):
Looks good!

packages/cli/src/rpc/explain/buildContext.ts
kgilpin: What it's doing is skipping the rest of the loop body, but maybe I don't need to be doing that.

ahtrotta: Ah, makes sense! I misunderstood the intention

PR #1642, opened 2024-03-24
Allow configuring LLM with environment variables

dividedmind:
This change allows using `APPMAP_NAVIE_MODEL` and `APPMAP_NAVIE_TOKEN_LIMIT` environment variables along with `OPENAI_{API_KEY,BASE_URL}` and `AZURE_OPENAI_API_{KEY,VERSION,INSTANCE_NAME,DEPLOYMENT_NAME` used by `@langchain/openai` to configure target LLM for Navie.

### DIFF
--- a/packages/navie/src/explain.ts
+++ b/packages/navie/src/explain.ts
@@ -112,8 +112,8 @@ export interface ClientRequest {
 }
 
 export class ExplainOptions {
-  modelName = 'gpt-4-0125-preview';
-  tokenLimit = 8000;
+  modelName = process.env.APPMAP_NAVIE_MODEL ?? 'gpt-4-0125-preview';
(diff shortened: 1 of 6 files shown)


packages/navie/src/explain.ts
```diff
@@ -112,8 +112,8 @@ export interface ClientRequest {
 }
 
 export class ExplainOptions {
-  modelName = 'gpt-4-0125-preview';
-  tokenLimit = 8000;
+  modelName = process.env.NAVIE_MODEL ?? 'gpt-4-0125-preview';
```
kgilpin, line 115: If we expose configuration via env var, I feel that we should prefix all our variables with `APPMAP_`.

kgilpin (approved):

PR #1643, opened 2024-03-25
feat: Navie updates (SWE bench)

kgilpin:
As a result of running SWE bench, I encountered a few problems or areas that needed improvement.

They are collected here! Review commit by commit.

## TODO

- [ ] Consider the impact of these changes, if any, on appmap-services
- [ ] Update appmap-services


packages/cli/src/rpc/explain/navie/navie-local.ts
```diff
@@ -40,8 +40,21 @@ class LocalHistory {
   }
 }
 
+const OPTION_SETTERS: Record<string, (explainOptions: Explain.ExplainOptions, value: string | number) => void> = {
+  tokenLimit: (explainOptions, value) => {
+    explainOptions.tokenLimit = value as number;
+  },
+  temperature: (explainOptions, value) => {
+    explainOptions.temperature = value as number;
+  },
```
dustinbyrne, line 56: You'd need to use something like `parseInt` to effectively coerce `value` into a string. This will only quiet the type checker, the underlying value will still be a string.

packages/cli/src/cmds/index/rpc.ts
```diff
@@ -49,6 +49,12 @@ export const builder = (args: yargs.Argv) => {
     boolean: true,
     default: false,
   });
+  args.option('ai-option', {
```
dustinbyrne, line 52: Might be worth adding an `alias` for this. `-o`?

dustinbyrne (approved):
I'm not aware of any reasons this would not work in `appmap-services`. The type casting of a string to number should probably be fixed before merging.

PR #1645, opened 2024-03-25
Show loading spinner when loading data

### LINKED ISSUE #1641: The UI has some indication that it's working when performing a long-running task

ahtrotta:
Fixes #1641


kgilpin:
@ahtrotta if we can hit a few of the most important cases in this PR, then we can always follow-on with other cases later. We don't have to get all of them right away.

ahtrotta:
> @ahtrotta if we can hit a few of the most important cases in this PR, then we can always follow-on with other cases later. We don't have to get all of them right away.

I'm having a bit of troubling getting things to behave the way I want to.

PR #1698, opened 2024-04-02
feat: Index project code

kgilpin:
Provides a web service endpoint `/context/<provider>` that returns project context according to a provider parameter.

Supported providers are:

- **bm25** Full-text search via bm25. Only files present in `git ls-files` are considered.
- **appmap** AppMap-based context as used by Navie.

Parameters:

- **charLimit** Desired number of characters to obtain.
- **keywords** List of keywords to search. Note that this is _not_ processed by the context service. If the user question should be processed by AI into keywords / vector terms, that needs to be done upstream.

HTTP method: Must be `POST`, to avoid any issues with the request size.
HTTP required headers: `Content-Type`

Run server:

```
appmap context-provider -d <dir>
```

Fetch context:

```
curl -H "Content-Type: application/json" \
  -X POST http://localhost:<port>/context/appmap \
  --data '{"charLimit": 10000,"keywords":["pet"]}' | jq | less
```

You can also use `context/bm25` to get full-text search context rather than the appmap (navie) context.


packages/cli/src/rpc/explain/buildContext.ts
```diff
@@ -61,10 +61,17 @@ export default async function buildContext(searchResults: SearchRpc.SearchResult
         continue;
       }
 
-      if (examinedLocations.has(event.location)) continue;
+      warn(`Location: ${event.location}`);
+      if (examinedLocations.has(event.location)) {
```
kgilpin, line 65: Remove. Committed by accident.

packages/cli/src/rpc/explain/buildContext.ts
```diff
@@ -61,10 +61,17 @@ export default async function buildContext(searchResults: SearchRpc.SearchResult
...
+        warn(`Duplicate location: ${event.location}`);
+        continue;
+      }
       examinedLocations.add(event.location);
 
-      if (codeSnippets.has(event.location)) continue;
+      if (codeSnippets.has(event.location)) {
```
kgilpin, line 71: Remove. Committed by accident.

PR #1704, opened 2024-04-04
fix: Allow unexpected returns 

### LINKED ISSUE #1703: Returns without calls are dropped
If a recording is toggled on in the middle of an async task, it can be possible for a return event to be generated with no corresponding call. If we see this upon normalization, we'll raise an exception.

However, this can be difficult to debug. The preferred behavior would be to drop the orphaned return.

dustinbyrne:
Resolves #1703

I've manually pulled this change into IntelliJ to verify it'll load AppMaps which were previously malformed.


dividedmind (approved):
NB, the last commit should probably come first (with modified message) so that all the commits build.

packages/models/src/appMapBuilder/eventStack.js
```diff
@@ -46,7 +40,10 @@ export default class EventStack {
           call.parent = parent;
         }
       } else {
-        throw new Error(`return #${event.id} is missing call #${event.parent_id}`);
+        // This can happen in async contexts where the recording began in the middle
+        // of another async task. It's possible that the parent call was executed
+        // before the recording began. If this is the case, discard the return event.
+        return;
```
dividedmind, line 46: It would be good idea to raise a warning, but I'm not sure where it should go...

dustinbyrne: I thought the same thing. The issue is that this is called in so many different contexts it felt wrong throwing a `console.log` in there, and we don't have any other logging abstraction.

kgilpin: Yeah, good discussion but the main thing is to just not crash :-)

PR #1713, opened 2024-04-10
feat: Emit navie-local telemetry events

### LINKED ISSUE #1712: Telemetry for BYOK/M

dividedmind:
Fixes #1712

On first explain request, emits an event like
```yaml
name: "@appland/appmap/navie-local"
properties:
  modelName: some-model
  azureVersion?: version
```

This allows tracking bring your own key/model. If one is just using bring you own key, the model will be the default (currently `gpt-4-0125-preview`). If Azure OpenAI is used, azureVersion field will be present.

### DIFF
--- a/packages/cli/src/rpc/explain/navie/navie-local.ts
+++ b/packages/cli/src/rpc/explain/navie/navie-local.ts
@@ -8,6 +8,7 @@ import { Context, Explain, Help, Message, ProjectInfo, explain } from '@appland/
 
 import INavie from './inavie';
 import { AgentMode } from '@appland/navie/dist/agent';
+import Telemetry from '../../../telemetry';
 
 class LocalHistory {
   constructor(public readonly threadId: string) {}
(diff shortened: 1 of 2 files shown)


packages/cli/src/rpc/explain/navie/navie-local.ts
```diff
@@ -128,4 +130,18 @@ export default class LocalNavie extends EventEmitter implements INavie {
     this.history.saveMessage({ content: response.join(''), role: 'assistant' });
     this.emit('complete');
   }
+
+  #skipTelemetry: boolean = !Telemetry.enabled;
```
kgilpin, line 134: Why does it start with a # ?

dividedmind: It's [private](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties).

PR #1718, opened 2024-04-12
feat: Refactor context data

kgilpin:
Change Navie context from an object to an array. This allows different kinds of context to be propagated to Navie using the same type. It also allows additional information to be propagated, such as file path and score for any context item.

## Done

- [x] Change ContextItem data from an object with type keys to an Array.
- [x] Test RemoteNavie adapter between V1 and V2 context types.
- [x] Move `applyContext` logic from lib-navie to the context provider. This will allow the context provider to properly implement options like the item type weights. See https://github.com/getappmap/appmap-js/pull/1721

## Notes

`Callbacks.onRequestContext` may now return `unknown[]`, in addition to `Record<string, unknown>`. Expect this in the hosted Navie provider.


packages/cli/src/rpc/explain/explain.ts
```diff
@@ -72,28 +72,28 @@ export class Explain extends EventEmitter {
...
     // pruned by the client AI anyway.
     // The meaning of tokenCount is "try and get at least this many tokens"
     const charLimit = tokenCount * 3;
-    const searchResult = await collectContext(
-      this.directories,
-      this.appmaps,
-      vectorTerms,
-      charLimit
-    );
+    const searchResult = await collectContext(this.directories, this.appmaps, keywords, charLimit);
```
kgilpin, line 96: This can be refactored to return an array instead of an object, because we are converting it into an array anyway.

packages/cli/src/rpc/explain/explain.ts
```diff
@@ -72,28 +72,28 @@ export class Explain extends EventEmitter {
...
 
     this.status.searchResponse = searchResult.searchResponse;
```
kgilpin, line 98: Is the frontend using this, or can it be dropped?

packages/cli/src/rpc/explain/navie/navie-remote.ts
```diff
@@ -35,7 +37,47 @@ export default class RemoteNavie extends EventEmitter implements INavie {
         async onRequestContext(data) {
           try {
             if (data.type === 'search') {
-              return await self.contextProvider(data as unknown as Context.ContextRequest);
+              const { version } = data;
+              let contextRequestV2: ContextV2.ContextRequest;
+              // Adapt v1 calls to v2
+              const isVersion1 = !version || version === 1;
+              if (isVersion1) {
+                const contextRequestV1 = data as ContextV1.ContextRequest;
```
kgilpin, line 45: Here a v1 request is converted to v2 format. This may occur if lib-navie is emitting v1, for example if navie is running on getappmap.com and it hasn't been updated yet.

packages/cli/src/rpc/explain/navie/navie-remote.ts
```diff
@@ -35,7 +37,47 @@ export default class RemoteNavie extends EventEmitter implements INavie {
...
+                throw new Error(`Unexpected context request version: ${version}`);
+              }
+
+              const responseV2: ContextV2.ContextResponse = await self.contextProvider({
+                ...contextRequestV2,
+                type: 'search',
+                version: 2,
+              });
+              if (isVersion1) {
+                // Adapt from V2 response back to V1. Some data may be lost in this process.
```
kgilpin, line 62: In case the client is expecting a v1 format, convert it back.

packages/navie/src/context.ts
```diff
@@ -1,19 +1,84 @@
...
+    tokenCount: number;
+  };
+
+  export type ContextItem = {
+    name: string;
+    score?: number;
+    content: string;
+  };
+
+  export type ContextResponse = {
```
kgilpin, line 14: In V1 the response is an object and the content types are the keys.

packages/navie/src/context.ts
```diff
@@ -1,19 +1,84 @@
...
+    // When specified, only return context items of these types.
+    itemTypes?: ContextItemType[];
+    // Weight the importance of the context items. The sum of the weights should be 1.
+    // Item types not specified will be omitted from the response, along with item types whose weight is 0 or less.
+    // If the user's question is directed most specifically to a certain type of context item, the weights should be
+    // set to emphasize that type of context item. If the user's question is more general, the weights can be omitted.
+    weights?: Record<ContextItemType, number>;
+  };
+
+  export type ContextResponse = ContextItem[];
```
kgilpin, line 79: In V2 the response is an array, with each array item carrying more information than in V1 (e.g. score).

packages/navie/src/interaction-history.ts
```diff
@@ -161,35 +165,38 @@ export class HelpLookupEvent extends InteractionEvent {
 }
 
 export class ContextItemEvent extends InteractionEvent {
-  constructor(public contextItem: ContextItem, public file?: string) {
+  constructor(public promptType: PromptType, public content: string, public location?: string) {
```
kgilpin, line 168: Context items will be emitted as prompt entries, so require each one here to match a known PromptType.

packages/cli/tests/integration/rpc.explain.spec.ts
```diff
@@ -3,7 +3,7 @@ import { ExplainRpc } from '@appland/rpc';
 import { Help, explain } from '@appland/navie';
 import { AI } from '@appland/client';
 import { AIClient, AICallbacks, AIInputPromptOptions, AIUserInput } from '@appland/client';
-import { ContextProvider, InteractionHistory, ProjectInfo } from '@appland/navie';
+import { ContextV2, InteractionHistory, ProjectInfo } from '@appland/navie';
```
kgilpin, line 6: TODO: This would be a good place to ensure that a V1 request is properly handled and a V1 response is returned.

packages/navie/src/context.ts
```diff
@@ -1,19 +1,84 @@
...
+    type: ContextItemType;
+    // The content of the context item. The content is specific to the type of context item.
+    content: string;
+    // A score that indicates the relevance of the context item to the request. Score is present
+    // only when the context item was obtained by searching for keywords (terms).
+    score?: number;
+  };
+
+  // Request a set of context items from the context provider.
+  export type ContextRequest = {
```
kgilpin, line 54: TODO: Some of these fields, e.g. `recent` are for future use, so they should be removed for now since the context provider does not implement them yet.

packages/cli/src/rpc/explain/buildContext.ts
```diff
@@ -44,34 +46,53 @@ export default async function buildContext(searchResults: SearchRpc.SearchResult
       formatOptions: { disableMarkup: true },
     });
     assert(typeof plantUML === 'string');
-    sequenceDiagrams.push(plantUML);
+    sequenceDiagrams.push({
+      type: ContextV2.ContextItemType.SequenceDiagram,
+      content: plantUML,
+      score: result.score,
```
dustinbyrne, line 52: Possible to include a location here? This would be useful to link to an AppMap from the frontend.

packages/client/src/aiClient.ts
```diff
@@ -17,7 +17,11 @@ export type Callbacks = {
   onComplete(): void;
   onRequestContext?: (
     data: Record<string, unknown>
-  ) => Record<string, unknown> | Promise<Record<string, unknown>>;
+  ) =>
+    | undefined
+    | Record<string, unknown>
+    | unknown[]
+    | Promise<Record<string, unknown> | unknown[] | undefined>;
```
dustinbyrne, line 24: The backend (maybe incorrectly) already considers cases where the context is an `unknown[]`, but `undefined` is new. Is `undefined` necessary?

packages/cli/src/rpc/explain/buildContext.ts
kgilpin: Yes good idea.

packages/client/src/aiClient.ts
kgilpin: You’re right we can return an empty object or array.

kgilpin:
Will merge as https://github.com/getappmap/appmap-js/pull/1723

PR #1721, opened 2024-04-12
feat: Apply fixes and updates to the apply-context function.

kgilpin:
Apply the context items to the Navie context within the context request. 

This allows the context provider to prioritize among all the context that's available, rather than just sending it all back to Navie to be applied. By being smarter in this way, the context provider can make better decisions about which different types of context to include, when to go deeper on a particular type of context, when to drop very large context items that won't fit into the requested context window (or will take up most of the room and leave little room for other context).

## Examples

Here are some examples to try and make the benefits clear.

### Diagram that dominates the context

 Suppose Navie asks the context provider for 12,000 characters of context (a typical amount). The context provider finds an 8,000 character sequence diagram and many more characters of code snippets, SQL, etc. In the existing implementation, the sequence diagram and 4,000 of other context will be sent to Navie, whereupon it will be sent to the AI. This is not a good solution, because the context is dominated by one large sequence diagram. In this new branch, individual context items are not allowed to dominate the context like that.

### Not enough variety (too many diagrams)

In this scenario, the context provider finds 10 sequence diagrams that are all very similar (maybe they are request recordings of different invocations of the same route). Then it looks for code snippets from those diagrams, and finds, say 4,000 characters of code snippets. So the context provider will now return 8,000 characters of very similar sequence diagrams over and over, plus 4,000 of code snippets. This is not a great context. In the new implementation, the context provider will not "fill out" the context with a bunch of sequence diagrams once it runs out of code snippets and data requests. Instead, it will dig deeper to find more code snippets and data requests that are maybe not as high of a match score, but still add more useful context and variety than a bunch of very similar sequence diagrams.

### Not enough variety (future)

In a coming PR, the context provider will be able to locate code snippets through full-text search; not just via AppMap events. So if AppMaps don't yield a lot of great code snippets, the context provider can broaden the search to go find snippets via full-text search. For example, it will start to pull in frontend files, styles, snippets of test cases and READMEs, whatever matches the user's query. These kind of snippets are not good if they are all that's available to the AI, but they are good for rounding out the context. 

## Recap

The context provider gets a lot more authority over which context is actually sent to the AI, out of all the context that is found via searching (both searching AppMaps and via full-text search of the code repo).

### DIFF
--- a/packages/navie/src/lib/apply-context.ts
+++ b/packages/navie/src/lib/apply-context.ts
@@ -1,0 +1,149 @@
+import { log } from 'console';
+
+import { ContextV2 } from '../context';
+
+const VERBOSE = process.env.DEBUG === 'true';
+
+enum ContextItemStatus {
+  OK = 'ok',
+  SIMILAR_CONTENT = 'similar_content',
+  ITEM_UNDEFINED = 'item_undefined',
+  ITEM_TOO_BIG = 'item_too_big',
+}
+
+export const DEFAULT_SELECTION_FREQUENCY: Map<ContextV2.ContextItemType, number> = Object.entries({
+  [ContextV2.ContextItemType.SequenceDiagram]: 1,
+  [ContextV2.ContextItemType.CodeSnippet]: 5,
+  [ContextV2.ContextItemType.DataRequest]: 2,
+  [ContextV2.ContextItemType.HelpDoc]: 1,
+}).reduce((map, [key, value]) => {
+  map.set(key as ContextV2.ContextItemType, value);
+  return map;
+}, new Map<ContextV2.ContextItemType, number>());
+
+export default function applyContext(
+  context: ContextV2.ContextItem[],
+  characterLimit: number,
+  maxContentLength: number = characterLimit / 5,
+  selectionFrequency: Map<ContextV2.ContextItemType, number> = DEFAULT_SELECTION_FREQUENCY,
+  verbose = VERBOSE
+): ContextV2.ContextItem[] {
+  const itemTypeByIndex = Array<ContextV2.ContextItemType>();
+  for (const [itemType, frequency] of selectionFrequency) {
+    for (let i = 0; i < frequency; i += 1) {
+      itemTypeByIndex.push(itemType);
+    }
+  }
+
+  const availableContext: Array<ContextV2.ContextItem | undefined> = [...context];
+  const roundSize = itemTypeByIndex.length;
+
+  // Select items in a round-robin fashion, to ensure a mix of content types. Heuristically, we
+  // want to see one sequence diagram, N code snippets, and M data requests. If we run out of
+  // one type of content type, we'll continue adding the other types.
+  let index = 0;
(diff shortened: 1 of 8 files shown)


packages/navie/src/lib/apply-context.ts
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    if (verbose) log(`Context item ${itemDescription(contextItem)} status: ${status}`);
+
+    if (hasSequenceDiagramBeenAdded && onlySequenceDiagramsRemaining(i)) {
+      log(
+        `Only sequence diagrams remain in the context, and at least one sequence diagram has already been added. No further context will be added.`
+      );
+      break;
+    }
```
dustinbyrne, line 129: Should this occur before `addContextItem` is called, or should it be `onlySequenceDiagramsRemaining(i + 1)`?

Otherwise this will consider the current context item, which could then be followed by only sequence diagrams. Maybe this is okay but the log message makes it look like this isn't intended.

packages/navie/src/lib/apply-context.ts
```diff
@@ -0,0 +1,144 @@
...
+  let charsRemaining = characterLimit;
+  log(`Remaining characters before context: ${charsRemaining}`);
+  const appliedContext = new Array<ContextV2.ContextItem>();
+
+  const itemDescription = (contextItem: ContextV2.ContextItem): string =>
+    [contextItem.type, contextItem.location].filter(Boolean).join(' ');
+
+  const addContextItem = (contextItem: ContextV2.ContextItem): ContextItemStatus => {
+    // Don't consume too much of the character limit on a single item.
+    if (contextItem.content.length > maxContentLength) {
```
dustinbyrne, line 73: `charsRemaining` isn't considered, so it appears that this context item could potentially overflow the token limit?

```suggestion
    if (contextItem.content.length > Math.min(maxContentLength, charsRemaining)) {
```

packages/navie/src/lib/apply-context.ts
```diff
@@ -0,0 +1,144 @@
...
+    if (verbose) log(`Context item ${itemDescription(contextItem)} status: ${status}`);
+
+    if (hasSequenceDiagramBeenAdded && onlySequenceDiagramsRemaining(i)) {
+      log(
+        `Only sequence diagrams remain in the context, and at least one sequence diagram has already been added. No further context will be added.`
+      );
+      break;
+    }
+
+    if (charsRemaining < 0) {
```
dustinbyrne, line 131: Seems like we should stop at 0 as well
```suggestion
    if (charsRemaining <= 0) {
```

dustinbyrne (approved):
Overall it looks good, though I left some considerations below. Ultimately this is merging into another feature branch, so I'm going to approve ahead of time.

packages/navie/src/lib/apply-context.ts
kgilpin: It's OK to overflow the token limit somewhat. It's a target to aim for but not expected to be a hard limit.

PR #1724, opened 2024-04-15
De-emphasize "No AppMaps" status in Navie

ahtrotta:
* Changes the "No AppMaps" status bar from red to blue to de-emphasize it. 
* Changes the warning/error exclamation point icon to an info icon.
* Makes it so that the "No AppMaps" status bar does not stay on the page while the user is chatting.
* Changed `No AppMaps available` to `No AppMap data available`.
* Changed `33 AppMaps available` to `33 AppMap recordings available` (where 33 is an example number).

## Before

image

## After

image

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -10,7 +10,7 @@
     "build": "tsc",
     "watch": "tsc --watch",
     "lint": "eslint",
-    "test": "appmap-agent-js"
+    "test": "appmap-node jest"
   },
   "publishConfig": {
     "access": "public"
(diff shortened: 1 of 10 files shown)


kgilpin:
@ahtrotta @Daniel-Warner-X this is a good step.

Next we need to use the RHS panel of Navie to communicate back to the user information like:

* What context is available (AppMaps, source files, number of projects included in the Navie workspace, etc)
* When a question is asked, what context has been selected. AppMaps / sequence diagrams (if applicable), source files, SQL, data requests, exceptions (future), logs (future), etc
* Each context source file is a link to open that file in the IDE
* Each AppMap is a link to open that AppMap in the IDE (I think we should move away from showing the AppMaps in-place. No-one is using this even when it would help address their question).

dustinbyrne (approved):

PR #1733, opened 2024-04-20
Lint TypeScript in packages/cli

dividedmind:
I noticed that we weren't linting .ts files in the CLI package.

I added the right configuration to enable this; also, removed the (quite obsolete) airbnb preset and replaced it with currently recommended ones.

I've fixed a couple low hanging lint problems and turned all the ones that are widespread in the package into warnings so they don't break the builds. We should strive to whittle down the list slowly.

### DIFF
--- a/packages/cli/.eslintrc.cjs
+++ b/packages/cli/.eslintrc.cjs
@@ -51,7 +51,7 @@ module.exports = {
   },
   overrides: [
     {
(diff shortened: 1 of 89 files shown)


kgilpin:
OK; let's get Navie shipped otherwise this looks like merge conflict hell.

PR #1745, opened 2024-04-26
Warn users in Navie when they don't have AppMap data

ahtrotta:
* Adds a message when there is no AppMap data to help the user create AppMap data.
* The `Open Instructions` button opens the installation instructions.
* Re-orders the context data so that AppMap data is above code snippets.
* Clips long AppMap names/paths with an ellipsis in the context view.

image

### DIFF
--- a/packages/components/src/components/chat-search/Context.vue
+++ b/packages/components/src/components/chat-search/Context.vue
@@ -16,6 +16,25 @@
       <div>
         <h2>Context</h2>
         <div class="context__body">
+          <div
+            v-if="numAppMaps === 0"
+            class="context__body__no-appmaps"
+            data-cy="create-appmap-data"
+          >
+            <p>
+              You don't have any AppMap data. Adding AppMap data to Navie's context improves Navie's
+              code suggestions and answers.
+            </p>
+            <v-button
+              data-cy="create-appmap-data-btn"
+              class="create-appmap-data"
+              size="small"
+              kind="ghost"
+              @click.native="openInstallInstructions"
+            >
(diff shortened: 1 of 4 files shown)


packages/components/src/components/chat-search/Context.vue
```diff
@@ -16,6 +16,25 @@
...
+              code suggestions and answers.
+            </p>
+            <v-button
+              data-cy="create-appmap-data-btn"
+              class="create-appmap-data"
+              size="small"
+              kind="ghost"
+              @click.native="openInstallInstructions"
+            >
+              Create AppMap Data
```
kgilpin, line 35: @evlawler since this is going to open the instructions, I might update the label to say "Open instructions" or "Open AppMap data instructions". Just so it's clear to the user what happens when the push the button. Pushing a button is always a pretty big decision - people want to know ahead of time what will happen. (e.g. it's not going to open a terminal, run code, launch a website, send an AI message, edit files, or do anything else risky. It's just opening instructions).

kgilpin: Should we also have a button that will start a Navie conversation in `@help` mode about how to make AppMap data?

kgilpin: So it could be like `Open instructions | Ask Navie for @help`

ahtrotta: I agree that it's good to be clear to the user, so I changed it to `Open Instructions`

dustinbyrne (approved):

PR #1758, opened 2024-04-30
ci: Move to a single appmap.yml in the project root

kgilpin:


kgilpin:
Accomplished elsewhere.

PR #1768, opened 2024-05-01
feat: Buffer and interpolate chunked completions

### LINKED ISSUE #1762: Completions returned via `explain` RPC are buffered and interpolated
Completion streaming via the `explain` RPC is really chunky because it's sampled every so often, instead of emitting tokens as they're generated. 

To fix this, we can buffer the completions and re-emit them in a smooth, fine grained stream.

dustinbyrne:
This gives the illusion of token streaming even though completions are streaming in multi-token chunks.

Before

[interp-before.webm](https://github.com/getappmap/appmap-js/assets/8737782/5ce788c8-094f-4cb8-8a5b-b2ea3a1e8e52)

After

[interp-after.webm](https://github.com/getappmap/appmap-js/assets/8737782/543bcf70-ad73-4d34-978c-dee02582cff7)

Resolves #1762

### DIFF
--- a/packages/components/src/components/chat/UserMessage.vue
+++ b/packages/components/src/components/chat/UserMessage.vue
@@ -167,9 +167,17 @@ export default {
 
   data() {
     return {
+      timestamp: Date.now(),
       sentimentTimeout: undefined,
       copiedMessageTimeout: undefined,
     };
+  },
+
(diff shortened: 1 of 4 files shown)


ahtrotta (approved):
Looks great!

PR #1780, opened 2024-05-08
feat: Add conversation parameters

kgilpin:
See https://github.com/getappmap/appmap-services/pull/70

These changes apply to navie-local only:

* Allocate thread and message ids by making API calls to api-services.
* Emit events when the agent name and classifiers are chosen.
* Update message metadata as the conversation progresses.

## TODO

- [x] Local Navie should not fail if the remote service is not available. For example, if the outbound calls are blocked by proxy.
- [x] Local Navie should allocate its own thread id if thread pre-allocation fails.
- [x] Local Navie should submit message metadata. Obtain the data from the Navie interaction history.
- [x] If thread pre-allocation fails, Local Navie should not attempt to post the conversation messages metadata (because the server will not have the thread id).

### DIFF
--- a/packages/navie/src/explain.ts
+++ b/packages/navie/src/explain.ts
@@ -1,0 +1,223 @@
+import { warn } from 'console';
+import EventEmitter from 'events';
+
+import Message from './message';
+import MemoryService from './services/memory-service';
+import VectorTermsService from './services/vector-terms-service';
+import CompletionService, { OpenAICompletionService } from './services/completion-service';
+import InteractionHistory, {
+  AgentSelectionEvent,
+  ClassificationEvent,
+  InteractionEvent,
+  InteractionHistoryEvents,
+} from './interaction-history';
+import { ContextV2 } from './context';
+import ProjectInfoService from './services/project-info-service';
+import { ProjectInfo, ProjectInfoProvider } from './project-info';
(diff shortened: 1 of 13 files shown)


packages/navie/src/explain.ts
```diff
@@ -188,12 +197,28 @@ export default function explain(
...
+        if (event instanceof ClassificationEvent) {
+          this.emit('classification', event.classification);
+        }
+      });
+    }
+
+    // eslint-disable-next-line class-methods-use-this
+    async *execute(): AsyncIterable<string> {
+      yield* codeExplainerService.execute(clientRequest, options, chatHistory);
+    }
```
dustinbyrne, line 220: Similar to returning a `Promise`, I don't believe the `async`/`yield` is necessary. We're not consuming the iterator, so it can be returned directly.

```suggestion
    execute(): AsyncIterable<string> {
      return codeExplainerService.execute(clientRequest, options, chatHistory);
    }
```

OTOH, this is what it looked like previously so maybe there's something I'm missing.

dustinbyrne (approved):

packages/navie/src/explain.ts
kgilpin: Yeah. Weirdly, it won't compile without the async.

PR #1782, opened 2024-05-09
fix: Directory resolution

kgilpin:
`listProjectFiles` is yielding a full path if there is a match in the project directory. Fix this function so that all matches are relative to the project directory.

Related: When the `rpc` command gets multiple `-d` options, handle these directories using the configuration v2 style.

### DIFF
--- a/packages/cli/src/fulltext/listProjectFiles.ts
+++ b/packages/cli/src/fulltext/listProjectFiles.ts
@@ -20,17 +20,17 @@ export default async function listProjectFiles(
   // Perform a breadth-first traversal of a directory, collecting all non-binary files and
   // applying the directory ignore list.
   const processDir = async (dir: string) => {
-    const queue = [dir];
+    const queue = ['.'];
     while (queue.length > 0 && files.length < fileLimit) {
       const currentDir = queue.shift();
       assert(currentDir, 'queue should not be empty');
 
-      const entries = await readdir(currentDir, { withFileTypes: true });
+      const entries = await readdir(join(dir, currentDir), { withFileTypes: true });
       for (const entry of entries) {
-        const fullPath = join(currentDir, entry.name);
+        const path = currentDir === '.' ? entry.name : join(currentDir, entry.name);
         if (entry.isDirectory()) {
-          if (!ignoreDirectory(entry.name)) queue.push(fullPath);
-        } else files.push(fullPath);
+          if (!ignoreDirectory(entry.name)) queue.push(path);
+        } else files.push(path);
       }
     }
   };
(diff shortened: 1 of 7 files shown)


packages/cli/src/fulltext/SourceIndex.ts
```diff
@@ -101,7 +101,7 @@ export class SourceIndex {
       if (verbose()) console.log(`No language found for file: ${fileName}`);
       splitter = new RecursiveCharacterTextSplitter();
     }
-    const filePath = join(directory, fileName);
+    const filePath = isAbsolute(fileName) ? fileName : join(directory, fileName);
```
kgilpin, line 104: @dustinbyrne I saw this fileName occurring as an absolute path so that when it was joined with directory it was a non-existent path. I'm not sure how this could have happened though, because the file paths come from `git ls-files` and `git status` and those should be relative paths, right?

kgilpin: It looks like it may be coming from the old listProjectFiles.

packages/cli/src/fulltext/listProjectFiles.ts
```diff
@@ -20,17 +20,17 @@ export default async function listProjectFiles(
...
       for (const entry of entries) {
-        const fullPath = join(currentDir, entry.name);
+        const path = currentDir === '.' ? entry.name : join(currentDir, entry.name);
         if (entry.isDirectory()) {
-          if (!ignoreDirectory(entry.name)) queue.push(fullPath);
-        } else files.push(fullPath);
+          if (!ignoreDirectory(entry.name)) queue.push(path);
+        } else files.push(path);
       }
     }
```
dustinbyrne, line 35: nit - Using `relative` might better capture the intention of the path transformations here
```ts
  const processDir = async (dir: string) => {
    const queue = [dir];
    while (queue.length > 0 && files.length < fileLimit) {
      const currentDir = queue.shift();
      assert(currentDir, 'queue should not be empty');

      const entries = await readdir(currentDir, { withFileTypes: true });
      for (const entry of entries) {
        const path = join(currentDir, entry.name);
        if (entry.isDirectory()) {
          if (!ignoreDirectory(entry.name)) queue.push(path);
        } else {
          files.push(relative(directory, path));
        }
      }
    }
  };
  ```

dustinbyrne (approved):

packages/cli/src/fulltext/listProjectFiles.ts
kgilpin: OK, I'll change in a different branch.

PR #1783, opened 2024-05-09
fix: Don't prompt with appmapStats for non-arch questions

kgilpin:
Prompt is too verbose with routes and packages for large projects.

Omit this info unless the user has asked a general architecture question.

### DIFF
--- a/packages/navie/src/services/project-info-service.ts
+++ b/packages/navie/src/services/project-info-service.ts
@@ -1,9 +1,22 @@
 import { dump } from 'js-yaml';
-import { ProjectInfo, ProjectInfoProvider } from '../project-info';
+import {
+  AppMapConfig,
+  AppMapStats,
+  CodeEditorInfo,
+  ProjectInfo,
+  ProjectInfoProvider,
+} from '../project-info';
 import InteractionHistory, { PromptInteractionEvent } from '../interaction-history';
 import { PromptType, buildPromptDescriptor, buildPromptValue } from '../prompt';
+import assert from 'assert';
 
 type Test = () => boolean;
+
+type MinimumStats = Omit<AppMapStats, 'routes' | 'tables' | 'packages'> & {
+  routes?: string[];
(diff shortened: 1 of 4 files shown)


packages/navie/src/services/project-info-service.ts
```diff
@@ -20,6 +20,25 @@ export default class ProjectInfoService {
...
+      appmapStats.packages?.length > 20 ||
+      appmapStats.routes?.length > 20 ||
+      appmapStats.tables?.length > 20;
+
+    if (!isArchitecture) {
+      for (const info of emitProjectInfo) {
+        if (info.appmapStats && isLargeProject(info.appmapStats))
+          info.appmapStats = {
+            numAppMaps: info.appmapStats.numAppMaps,
+          } as AppMapStats;
```
dustinbyrne, line 39: This violates the type definition for `AppMapStats`. Downstream consumers may rightfully expect `packages` / `routes` / `tables` to be valid arrays.

packages/navie/src/services/project-info-service.ts
```diff
@@ -20,6 +20,25 @@ export default class ProjectInfoService {
...
+    return projectInfo;
+  }
+
+  promptProjectInfo(isArchitecture: boolean, projectInfo: ProjectInfo[]) {
+    const emitProjectInfo: ProjectInfo[] = projectInfo.map((info) => ({ ...info }));
+
+    const isLargeProject = (appmapStats: AppMapStats) =>
+      appmapStats.packages?.length > 20 ||
+      appmapStats.routes?.length > 20 ||
+      appmapStats.tables?.length > 20;
```
dustinbyrne, line 32: These shouldn't be nullish
https://github.com/getappmap/appmap-js/blob/c0b609098e5c52fd4cc193366446faeaa665e305/packages/navie/src/project-info.ts#L10-L15

dustinbyrne (approved):

PR #1785, opened 2024-05-10
fix: update pip install command

### LINKED ISSUE #1784: Python installation instructions for pip include `--require-virtualenv`
The install instructions for pip should be `pip install --require-virtualenv appmap`, to reduce the chance that the user will install `appmap` into their system python environment.

apotterri:
Update the pip install instructions to include `--require-virtualenv`.

Fixes #1784 .

### DIFF
--- a/packages/components/src/components/install-guide/install-instructions/Python.vue
+++ b/packages/components/src/components/install-guide/install-instructions/Python.vue
@@ -15,7 +15,14 @@
     <section>
       <h3>Add the <code class="inline">appmap</code> package to your project</h3>
       <div>
-        <v-code-snippet clipboard-text="pip install appmap" :kind="ctaButtonType" />
+        <v-code-snippet
+          clipboard-text="pip install --require-virtualenv appmap"
+          :kind="ctaButtonType"
+        />
+        <p>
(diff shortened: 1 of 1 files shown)


packages/components/src/components/install-guide/install-instructions/Python.vue
```diff
@@ -15,7 +15,14 @@
...
       <h3>Add the <code class="inline">appmap</code> package to your project</h3>
       <div>
-        <v-code-snippet clipboard-text="pip install appmap" :kind="ctaButtonType" />
+        <v-code-snippet
+          clipboard-text="pip install --require-virtualenv appmap"
+          :kind="ctaButtonType"
+        />
+        <p>
+          Note that this requires that you create and activate a
+          <a href="https://docs.python.org/3/library/venv.html">virtual environment</a>.
```
kgilpin, line 24: My only question was whether it works to put links in these pages. From looking at other pages, it looks like we do this elsewhere so it's fine. @dustinbyrne  ?

dustinbyrne: Links like this work as expected from our webviews.

PR #1788, opened 2024-05-10
Navie oneshot command

dividedmind:
This adds `appmap navie` command to run one-shot interactions with Navie that do the same context gathering as the RPC commands.

```console
$ LC_ALL=C ./built/cli.js navie --help
cli.js navie [question..]

Explain a question using Navie

Positionals:
  question  Question text; appended to any other input     [array] [default: []]

Options:
      --version         Show version number                            [boolean]
  -v, --verbose         Run with verbose logging                       [boolean]
      --help            Show help                                      [boolean]
  -d, --directory       program working directory
             [array] [default: ["/home/divide/projects/appmap-js/packages/cli"]]
      --navie-provider  navie provider to use
                                           [string] [choices: "local", "remote"]
      --log-navie       Log Navie events to stderr    [boolean] [default: false]
      --ai-option       Specify an extended option to the AI provider, in the
                        form of a key=value pair. May be repeated.       [array]
      --agent-mode      Agent mode which to run the Navie AI. The agent can also
                        be controlled by starting the question with '@<agent> '.
                                        [choices: "explain", "generate", "help"]
      --code-editor     Active code editor. This information is used to tune the
                        @help responses. If unspecified, the code editor may be
                        picked up from environment variables APPMAP_CODE_EDITOR,
                        TERM_PROGRAM and TERMINAL_EMULATOR.             [string]
  -o, --output          Output path                                     [string]
  -i, --input           Input path                                      [string]

$ ./built/cli.js navie -d /tmp/user/1000/swe-navie-marshmallow-3.0-d14cebrt/ "what does this project do?" 2>/dev/null
The project appears to revolve around Marshmallow, a Python library used for object serialization. Marshmallow assists in converting complex data types, like ORM/ODM entities or custom class objects, to and from native Python data types like dictionaries, which can then be rendered into standard formats such as JSON.
[...]
```

### DIFF
--- a/packages/cli/src/cmds/navie.ts
+++ b/packages/cli/src/cmds/navie.ts
@@ -1,0 +1,236 @@
+import { warn } from 'node:console';
+import { createWriteStream } from 'node:fs';
+import { readFile } from 'node:fs/promises';
+import type { Writable } from 'node:stream';
+import { text } from 'node:stream/consumers';
+
+import chalk from 'chalk';
+import yargs from 'yargs';
+
+import { loadConfiguration } from '@appland/client';
+import { Agents, ContextV2, Help, ProjectInfo } from '@appland/navie';
+import { InteractionEvent } from '@appland/navie/dist/interaction-history';
+
+import { configureRpcDirectories } from '../lib/handleWorkingDirectory';
+import { explainHandler } from '../rpc/explain/explain';
+import INavie, { INavieProvider } from '../rpc/explain/navie/inavie';
+import LocalNavie from '../rpc/explain/navie/navie-local';
+import RemoteNavie from '../rpc/explain/navie/navie-remote';
+import detectAIEnvVar, { AI_KEY_ENV_VARS } from './index/aiEnvVar';
+
+interface ExplainArgs {
+  aiOption?: string[];
+  agentMode?: Agents;
+  navieProvider?: string;
+  logNavie?: boolean;
+}
+
+interface NavieCommonCmdArgs extends ExplainArgs {
+  codeEditor?: string;
+  directory: string[];
+}
+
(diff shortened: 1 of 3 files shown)


packages/cli/src/fulltext/FileIndex.ts
```diff
@@ -39,7 +39,7 @@ export class FileIndex {
     const searchExpr = queryKeywords(keywords).join(' OR ');
     const rows = this.database.prepare(query).all(searchExpr, limit);
     rows.forEach((row: any) => {
-      if (verbose()) console.log(`Found row ${row.file_name}`);
+      if (verbose()) console.warn(`Found row ${row.file_name}`);
```
kgilpin, line 42: I love getting the answer on stdout too, but I recommend that for your scripting purposes you add a file output option. Because in the future someone is probably going to accidentally write to the `log` again.

packages/cli/src/cmds/navie.ts
```diff
@@ -0,0 +1,202 @@
...
+
+interface NavieCmdArgs extends NavieCommonCmdArgs {
+  questionPath: string;
+}
+
+export const command = 'navie';
+export const describe = 'Explain a question using Navie';
+
+export function builder<T>(args: yargs.Argv<T>): yargs.Argv<T & NavieCommonCmdArgs> {
+  return commonNavieArgsBuilder(args).positional('question-path', {
```
kgilpin, line 176: For people who are not as shell-adept (and for Windows users) I'd recommend that the default positional argument be the question text. And that question input from a file be a separate option.

kgilpin:
Thanks, I do have a couple of comments on the interface that I think will make this command a bit more accessible to use, even if they make it less pure in some sense.

kgilpin (approved):

PR #1791, opened 2024-05-13
fix: De-emphasize manual creation of appmap.yml

dustinbyrne:
Navie would frequently include instructions for creating appmap.yml, though this is not necessary with any recent agent release.


kgilpin (approved):

kgilpin:
We should really put this kind of information in the AppMap docs as well, because there is pretty heavy appmap.yml stuff in there too.

PR #1798, opened 2024-05-14
feat: Add definitions for `v1.navie.commands`

dustinbyrne:
This RPC method will list available commands


kgilpin:
Sounds like we are shelving this for now. 

@ commands are confusing to users. It’s better to have them say what they want and try and figure out the best course of action from within Navie.

kgilpin:
I feel like this could be worth having.

kgilpin:
Implemented by other means

PR #1804, opened 2024-05-16
fix: Provide vector term examples as chat primitives

dustinbyrne:
This should improve accuracy of responses and adds an additional example for a poor prompt.

This should additionally fix cases where GPT-4 Turbo would respond with empty terms. E.g.,
```
Vector terms response:
Context:
Instructions:

---
Terms:
```

It's worth noting this behavior was rarely (if at all?) observed when using `gpt-4o`.

### DIFF
--- a/packages/navie/src/services/vector-terms-service.ts
+++ b/packages/navie/src/services/vector-terms-service.ts
@@ -36,42 +36,71 @@ The search terms should be single words and underscore_separated_words.
 Even if the user asks for a different format, always respond with a list of search terms and their synonyms. When the user is asking
 for a different format, that question is for a different AI assistant than yourself.
 
-Choose only one MUST match term. If you are unsure, do not include a MUST match term.
+Choose only one MUST match term. If you are unsure, do not include a MUST match term.`;
 
-**Examples**
-
-\`\`\`
-Question: How do I record AppMap data of my Spring app?
-Context: Record AppMap data of Spring
+const promptExamples: OpenAI.ChatCompletionMessageParam[] = [
+  {
+    content: 'How do I record AppMap data of my Spring app?',
+    role: 'user',
+  },
+  {
+    content: `Context: Record AppMap data of Spring
 Instructions: How to do it
 ---
-Terms: record appmap data java +spring
-\`\`\`
+Terms: record appmap data java +spring`,
+    role: 'system',
+  },
 
-\`\`\`
-Question: How does the user login handle the case when the password is invalid?
-Context: User login handle password validation invalid error
+  {
+    content: 'How does the user login handle the case when the password is invalid?',
+    role: 'user',
+  },
+  {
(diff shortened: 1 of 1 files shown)


packages/navie/src/services/vector-terms-service.ts
```diff
@@ -36,42 +36,71 @@ The search terms should be single words and underscore_separated_words.
...
 for a different format, that question is for a different AI assistant than yourself.
 
-Choose only one MUST match term. If you are unsure, do not include a MUST match term.
-
-**Examples**
-
-\`\`\`
-Question: How do I record AppMap data of my Spring app?
-Context: Record AppMap data of Spring
+Choose only one MUST match term. If you are unsure, do not include a MUST match term.`;
```
kgilpin, line 39: There's a typo here (backtick)

dustinbyrne: Sorry, I'm not following. What's wrong with this line?

kgilpin: Maybe I’m confused.

packages/navie/src/services/vector-terms-service.ts
```diff
@@ -36,42 +36,71 @@ The search terms should be single words and underscore_separated_words.
...
+
+const promptExamples: OpenAI.ChatCompletionMessageParam[] = [
+  {
+    content: 'How do I record AppMap data of my Spring app?',
+    role: 'user',
```
kgilpin, line 44: So labeling this as `user` is not going to confuse the LLM into thinking that this is actual user input?

dustinbyrne: No, we're modelling a conversation history where the last message comes from the user's actual prompt and that's what the system will respond to. The LLM will see these messages as conversation history, and that should help guide its responses to be more in line with the examples we've provided it.

kgilpin (approved):
Seems to work OK, although it will still emit an empty list sometimes: 

```
node ./built/cli.js navie -d ~/source/land-of-apps/spring-petclinic -v "@vector-terms HI"
Debugger attached.
Setting RPC configuration: {"projectDirectories":["/Users/kgilpin/source/land-of-apps/spring-petclinic"],"appmapConfigFiles":["/Users/kgilpin/source/land-of-apps/spring-petclinic/appmap.yml"]}
No output specified, writing to stdout
Detected code editor: vscode
Using local Navie provider due to presence of environment variable OPENAI_API_KEY
[local-navie] Starting new thread e28b163c-2e48-46a7-ac87-ec1321148296
[local-navie] Processing question e433eb32-ca87-4046-bc8f-e8a51aa11cc7 in thread 21bb3381-00a5-49d4-87df-d5ba350e9ad2
Vector terms response:
Context: 
Instructions: 
---
Terms: 
Vector terms result: 
```

PR #1805, opened 2024-05-16
fix: Drop token interpolation

dustinbyrne:
It presents a few issues that would need time to solve:
1. There is evidence that tokens can emit at about 1 token/s. I'm not certain on how reproduce this at the moment.
2. The speed of token emission makes some existing issues more visible such as:
    a.) A single new line may not be enough buffer to catch the auto-scroll behavior. In other words, users sometimes need to 
scroll manually during completion.
    b.) The cursor jumps in between the boundary of span and text nodes, particularly in code blocks. When a new line is written in a text node, the new line's width will stretch to include the previous line, causing the cursor to drop to its own new line.


packages/components/src/pages/ChatSearch.vue
```diff
@@ -343,34 +344,28 @@ export default {
...
         };
 
         const onError = (error) => {
-          interp.dispose();
           onComplete();
           this.$refs.vchat.onError(error, systemMessage);
           reject();
         };
 
-        const interp = new CompletionInterpolation();
```
kgilpin, line 359: Note the warning about the unused import. So it probably makes senes to drop the file.

kgilpin (approved):
I'm not sure how to review this, but please do your best with testing it, I think it's important to get this out.

PR #1809, opened 2024-05-17
fix: install subcommand checks that target exists

### LINKED ISSUE #1808: `install` subcommand checks that the target direcetory exists
The `install` subcommand should check that the directory chose for installation exists and is a directory. Right now, it offers to open a support ticket:
```
ajp@Alans-MacBook-Pro empty % npx @appland/appmap install /tmp/foo

Help is available from the AppMap support team! 

? Details of the error will be provided to the support team.

Would you like to review the details here before submitting the support ticket? Yes

Error: ENOENT: no such file or directory, scandir '/tmp/foo'
```

apotterri:
Inform the user if the directory they've chosen for installation doesn't exist.

Fixes #1808.

### DIFF
--- a/packages/cli/tests/unit/agentInstall/installCommand.spec.ts
+++ b/packages/cli/tests/unit/agentInstall/installCommand.spec.ts
@@ -5,6 +5,7 @@ import sinon, { SinonStub } from 'sinon';
 import inquirer from 'inquirer';
 import Telemetry from '../../../src/telemetry';
 import yargs from 'yargs';
+import { existsSync } from 'fs';
 
 import { PoetryInstaller } from '../../../src/cmds/agentInstaller/pythonAgentInstaller';
 import { BundleInstaller } from '../../../src/cmds/agentInstaller/rubyAgentInstaller';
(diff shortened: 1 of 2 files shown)


packages/cli/tests/unit/agentInstall/installCommand.spec.ts
```diff
@@ -799,6 +800,28 @@ appmap_dir: tmp/appmap
       sinon.stub(commandRunner, 'run').resolves({ stdout: '', stderr: '' });
     });
 
+    it("tells the user if the target directory doesn't exist", async () => {
+      expect.assertions(5);
+
+      const badDirectory = '/tmp/foo';
```
dividedmind, line 806: Maybe use something more unique here. That's the goto path for me when creating throwaway directories ;)

packages/cli/tests/unit/agentInstall/installCommand.spec.ts
```diff
@@ -799,6 +800,28 @@ appmap_dir: tmp/appmap
...
+      expect(existsSync(badDirectory)).toBeFalsy(); // sanity check
+
+      const log = jest.spyOn(console, 'error');
+      const installProcedureStub = sinon
```
dividedmind, line 810: This mixing of sinon and jest is a bit curious.

dividedmind (approved):

packages/cli/tests/unit/agentInstall/installCommand.spec.ts
apotterri: lol, ok

packages/cli/tests/unit/agentInstall/installCommand.spec.ts
apotterri: I just grabbed bits that I needed from other places in this file....

I switched to using `sinon` exclusively.

PR #1813, opened 2024-05-17
Add some additional prompts to try and reduce the recommending of appmap.yml

petecheslock:


packages/navie/src/agents/explain-agent.ts
```diff
@@ -51,7 +54,10 @@ Observe the following guidelines when encouraging the user to make AppMap data:
 2) For Ruby, do not instruct the user to set \`APPMAP=true\`, because
   these languages already generate AppMap data automatically based on the Rails environment.
 3) Do not instruct the user to install the NPM package or CLI binary package \`@appland/appmap\` for the purpose of
-  making AppMap data, beacuse this package is not needed for that purpose.
+  making AppMap data, because this package is not needed for that purpose.
+4) Do not suggest that a user needs to create, update, or generate an appmap.yml configuration file unless the user specifically asks about that.
```
kgilpin, line 58: Do you think that this language might be too strong? Could we say "Do not show examples of appmap.yml or changes to appmap.yml unless the user is asking specifically about it, or is asking to change the behavior of their AppMap configuration in a way that requires changes to appmap.yml. (such as adding new packages, removing packages and/or code, making recording more shallow, etc).

petecheslock: yea was thinking about that as there might be a scenario someone wants to do advanced things and does ask for it. i'll adjust and make some more queries to see how it performs

PR #1817, opened 2024-05-20
fix: Fall back to user input on empty vector search

dustinbyrne:
If the LLM responds with no vector terms, the user input will be used verbatim.

### DIFF
--- a/packages/navie/src/services/vector-terms-service.ts
+++ b/packages/navie/src/services/vector-terms-service.ts
@@ -5,6 +5,7 @@ import { ChatOpenAI } from '@langchain/openai';
 
 import InteractionHistory, { VectorTermsInteractionEvent } from '../interaction-history';
 import trimFences from '../lib/trim-fences';
+import isEmpty from '../lib/is-empty';
 
 const SYSTEM_PROMPT = `You are assisting a developer to search a code base.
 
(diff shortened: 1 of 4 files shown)


packages/navie/src/services/vector-terms-service.ts
```diff
@@ -136,6 +137,10 @@ export default class VectorTermsService {
       responseText = contentAfter(responseText, 'Terms:');
       responseText = trimFences(responseText);
       searchTermsObject = parseJSON(responseText) || parseText(responseText);
+      if (isEmpty(searchTermsObject)) {
```
kgilpin, line 140: I think it's easier than this, because this case can only happen if the response is valid JSON (which we aren't actually asking for, but I suppose could happen). Otherwise, parseText is already just splitting the text on whitespace. Then there is a bunch of postprocessing that happens after this, e.g. to eliminate non-character words and very short words.

packages/navie/src/services/vector-terms-service.ts
```diff
@@ -5,6 +5,7 @@ import { ChatOpenAI } from '@langchain/openai';
 
 import InteractionHistory, { VectorTermsInteractionEvent } from '../interaction-history';
 import trimFences from '../lib/trim-fences';
+import isEmpty from '../lib/is-empty';
```
kgilpin, line 8: Overall I think it would be more effective to handle the case on the backend when the search terms are empty; and also to handle the empty search terms case within Navie and not not perform a context lookup if the vector terms are empty.

If the vector terms are empty, it's very unlikely that substituting the search with the user's raw search words is going to do any better. I haven't seen the LLM just fail to come up with any search terms as long as the user's question is at all reasonable (e.g. it isn't "HI").

PR #1818, opened 2024-05-20
fix: Skip searching for context on empty input

dustinbyrne:
This change prevents context search from continuing if the supplied list of vector terms is empty.

Fixes https://github.com/getappmap/appmap-js/issues/1767

### DIFF
--- a/packages/cli/src/rpc/explain/collectContext.ts
+++ b/packages/cli/src/rpc/explain/collectContext.ts
@@ -6,7 +6,7 @@ import FindEvents, {
 } from '../../fulltext/FindEvents';
 import { DEFAULT_MAX_DIAGRAMS, DEFAULT_MAX_FILES } from '../search/search';
 import buildContext from './buildContext';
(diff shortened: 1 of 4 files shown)


packages/cli/src/rpc/explain/collectContext.ts
```diff
@@ -118,6 +118,11 @@ export class ContextCollector {
     searchResponse: SearchRpc.SearchResponse;
     context: ContextV2.ContextResponse;
   }> {
+    if (this.vectorTerms.length === 0 || this.vectorTerms.every((term) => term.trim() === '')) {
```
kgilpin, line 121: Yup you can do the same thing for Help context and even handle it also on the Navie side if you wish.

dustinbyrne: I'll post a separate PR with a change to Navie

kgilpin (approved):

PR #1820, opened 2024-05-22
feat: Stop Navie answer

zermelo-wisen:
Adds functionality to stop the assistant's responses mid-answer.

### DIFF
--- a/packages/components/src/components/chat/Chat.vue
+++ b/packages/components/src/components/chat/Chat.vue
@@ -54,10 +54,12 @@
     </div>
     <v-chat-input
       @send="onSend"
+      @stop="onStop"
       :placeholder="inputPlaceholder"
       :class="inputClasses"
       :question="question"
       :code-selections="codeSelections"
+      :is-stop-active="isResponseStreaming"
       ref="input"
     />
   </div>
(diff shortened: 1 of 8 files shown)


packages/components/src/components/chat/Chat.vue
```diff
@@ -171,6 +173,7 @@ export default {
       enableScrollLog: false, // Auto-scroll can be tricky, so there is special logging to help debug it.
       codeSelections: [] as CodeSelection[],
       appmaps: [] as string[],
+      isAssistantMessageActive: false,
```
dustinbyrne, line 176: There's a `complete` property on each message, indicating whether or not the response has completed. You could use a computed property to determine whether or not there is an active system response instead of tracking this state manually and separately.

```ts
{

  // ...

  computed: {
    isResponseStreaming(): boolean {
      return this.getMessage({ complete: false, isUser: false }) !== undefined;
    }
  }

  // ...

}

kgilpin: It may seem nitpicky but UI changes are very hard to fix or revert once they are in the wild.

zermelo-wisen: I removed the extra state variable and applied @dustinbyrne's suggestion. Thank you.

dustinbyrne (approved):

PR #1822, opened 2024-05-23
feat: Sync docs from applandinc.appmap.io

kgilpin:
Sync docs from https://github.com/getappmap/applandinc.github.io


petecheslock:
Fixed by https://github.com/getappmap/appmap-js/pull/1848

PR #1834, opened 2024-06-05
feat: Output search context

dustinbyrne:
Cherry picked from https://github.com/getappmap/appmap-js/commit/478c3d6ac4703ad997fbfb93b5d7a59354c6b3bc

This information is useful for understanding how a plan was formed. This should be output along with the rest of the solve logs and data.


kgilpin:
There is already a @context command though

kgilpin:
I’m not sure I understand the use case for this

kgilpin:
Also there is already a file logged called context.yml

PR #1835, opened 2024-06-06
fix: Trim trailing whitespace in `change`

dustinbyrne:
The LLM tends to add additional spaces to the end of change sequences. This would likely cause a matches to fail.


kgilpin:
Fixed by https://github.com/getappmap/appmap-js/pull/1837

PR #1837, opened 2024-06-06
fix: Apply changes in a more robust way

dividedmind:
I'm not sure why it fails with the old code, but I didn't dig in; instead I rewrote it from ground up in a cleaner manner and it seems to work fine.

I added a test that fails with the old code (minimized real example) and succeeds with the new.

### DIFF
--- a/packages/navie/test/services/file-update-service/whitespace-mismatch/fields-expected.py
+++ b/packages/navie/test/services/file-update-service/whitespace-mismatch/fields-expected.py
@@ -1,0 +1,13 @@
+
+
+    def _bind_to_schema(self, field_name, schema):
+        super()._bind_to_schema(field_name, schema)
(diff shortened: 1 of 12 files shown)


packages/navie/test/services/file-update-service/whitespace-mismatch/fields-expected.py
```diff
@@ -1,5 +1,4 @@
-
-
+    # hello
```
kgilpin, line 1: Hello

packages/navie/test/services/file-update-service/whitespace-mismatch/fields.py
```diff
@@ -1,5 +1,4 @@
-
-
+    # hello
```
kgilpin, line 1: Hello again

PR #1844, opened 2024-06-11
fix: Tolerate mismatch in number of blank lines when applying update

### LINKED ISSUE #1841: Patch applier could tolerate a difference in the number of blank lines
LLM will sometimes generate a patch "original" that differs from the actual content only in a sequence of blank lines.

The patch applier could tolerate this, regard it as a match and apply the patch.

dividedmind:
Fixes #1841


kgilpin (approved):

kgilpin:
Merged into swe-bench

PR #1845, opened 2024-06-13
feat: --prompt and /noterms options

kgilpin:
Sideload system prompt text to Navie.

Pass `/noterms` to skip vector term expansion, and use the raw question history and code selection for context search.

### DIFF
--- a/.github/workflows/build-native-scanner.yml
+++ b/.github/workflows/build-native-scanner.yml
@@ -8,7 +8,7 @@ on:
       - '@appland/scanner-v*'
 
 jobs:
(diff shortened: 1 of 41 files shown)


dustinbyrne:
I've read through these changes and everything looks pretty straightforward to me. :+1:

kgilpin:
> I've read through these changes and everything looks pretty straightforward to me. 👍

Thanks, I will run through the benchmark & then merge if it all looks OK. Do we want to be merging to main? I was holding off on that because of the asset build problems.

PR #1852, opened 2024-06-24
feat: Extract with AI

kgilpin:
Extract code change suggestions using the LLM

## Note

Changing the prompting to emit XML-style suggestions and then using a regexp / XML parser would probably work better.


kgilpin:
Closing - no plan to use this at this time.

PR #1855, opened 2024-06-25
wip: Add an 'apply' button to relevant code snippets

dustinbyrne:
This change:
- Adds an apply button to code blocks which follow a comment directive containing a file path, e.g.,
  ```html
  <!-- file: app/models/user.rb -->
  ```
- Migrates code block frontend from inline strings to Vue components
- TODO: Highlight code in the code block instead of the user message?
- TODO: Improve the lifetime management of dynamic markdown Vue components (is this even necessary)?

image

### DIFF
--- a/packages/components/src/components/chat/MarkdownCodeSnippet.vue
+++ b/packages/components/src/components/chat/MarkdownCodeSnippet.vue
@@ -1,0 +1,109 @@
+<template>
+  <div class="code-snippet">
+    <div class="code-snippet__header">
+      <span class="code-snippet__language">{{ language }}</span>
+      <span class="code-snippet__button-group">
+        <span class="code-snippet__button" data-cy="apply" v-if="sourcePath">
+          <v-apply-icon />
+          Apply
+        </span>
+        <span class="code-snippet__button" data-cy="copy">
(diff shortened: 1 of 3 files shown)


dustinbyrne:
Four examples of referencing a code snippet seems to be working well. Out of over 30 generations I ran, all of them referenced files correctly.

I've also split the path up into the location and directory. I've noticed that when referencing files outside of code snippets, e.g., in a list of proposed changes, the LLM was tending to emit absolute paths. This seems to have improved that as well.

Another small fix to make sure Windows paths are able to be rendered properly (with tests).

dustinbyrne:
After plenty more testing I caved and ended up supporting the file directive both before the code block and as the first line within the code block.

E.g., both cases are now valid
````
<!-- file: /home/user/projects/blog/src/index.js -->
```js
const app = require('express')();
```
````

````
```js
<!-- file: /home/user/projects/blog/src/index.js -->
const app = require('express')();
```
````

Despite prompting for the former, the latter seems impossible to avoid. Given gpt-4o's unprompted tendency to include a file path as a comment on the first line of a code block, this may have been a fine-tuned behavior.

PR #1857, opened 2024-06-25
feat: Improve the readability of LLM configuration

dustinbyrne:
It should now be clear what's a button vs. what isn't, as well as what each item does. I've also added a close button for clarity.
image


kgilpin:
Looks nice. Is it able to indicate the current state?

dustinbyrne:
When using the Navie backend the state is reflected. The other two can always be selected.
image

PR #1871, opened 2024-07-05
feat: Update Java, Ruby and Python recording instructions

dustinbyrne:
This brings them in line with the (already released and live in VS Code) JavaScript instructions.

Python is also updated to include changes introduced in v2 of the agent, such as running `appmap-python`. This will need some fact checking.

### DIFF
--- a/packages/components/src/components/install-guide/record-instructions/Python.vue
+++ b/packages/components/src/components/install-guide/record-instructions/Python.vue
@@ -1,115 +1,202 @@
 <template>
   <section>
+    <h3>Getting started</h3>
     <p>
-      When you run your Python code with the <code class="inline"> appmap </code> package, AppMap
-      will be enabled for recording.
+      To record your Python application with AppMap, run your application using the
+      <code class="inline">appmap-python</code> command.
     </p>
-    <h2>Choose the best recording method for this project</h2>
-    <template v-if="testFramework">
-      <section class="recording-method">
-        <h3>
-          <!--TODO make recommended tag dynamic-->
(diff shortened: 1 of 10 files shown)


apotterri:
Would it be possible to attach some snapshots of the updated instructions for Java and Python? I'll be happy to review them, but it's really hard to sift through all the markup surrounding them in the commits.

dustinbyrne:
Are you able to access this page?
https://5fe18baf66f2280021e634d0-jvvulqxbcs.chromatic.com/?path=/story/pages-vs-code-install-guide-pages-record-appmaps--python-only

apotterri:
Yup, I'll review from there.

packages/components/src/components/install-guide/record-instructions/Python.vue
```diff
@@ -1,121 +1,231 @@
...
   },
+
+  mounted() {
+    if (!this.$refs.startCommand) return;
+    if (!(this.$refs.startCommand instanceof HTMLElement)) return;
+
+    this.typewrite(this.$refs.startCommand, [
+      '<start command>',
+      'script.py',
+      'flask --app main.app',
```
apotterri, line 215: This should be `flask --debug --app main.app`.

apotterri: Also, in the chromatic page at least, these scroll by really quickly.

apotterri (changes requested):
Needs a tweak for flask, but looks good otherwise.

PR #1877, opened 2024-07-10
feat: Improve classification of the latest question

dustinbyrne:
This change introduces a short conversational context window with the goal of improving the classification of the latest user question.

From my own observation, classifications tend to stick around given the inclusion of the entire user message history despite the intent of the most recent prompt. E.g., a user who asked for a diagram will repeatedly be classified as `diagram`, causing the diagram agent to respond to inquiries.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,7 +37,12 @@ pnpm-debug.log*
 
 release
 
+# Jupyter
+notebooks
+.env.notebook
+
 # AppMap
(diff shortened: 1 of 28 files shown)


kgilpin:
I have a similar change so we can compare notes.

kgilpin:
Incorporated via https://github.com/getappmap/appmap-js/pull/1880/commits

PR #1897, opened 2024-07-22
feat: Update command

kgilpin:
## RPC command `file.update`

```
export namespace FileRpc {
  export const UpdateFunctionName = 'file.update';

  export type UpdateOptions = {
    file: string;
    modified: string;
  };

  export type UpdateResponse = {
    exists: boolean;
    succeeded: boolean;
    original?: string;
    modified?: string;
  };
}
```

Creates-or-updates the named file, using the Navie `@update` command to build a "change" object if the file does not exist. From there, the update mechanism is the same as the one we used in SWE Bench, which was effective.

## Navie command `@update`

Given:

* Full content of an existing file, as Navie `codeSelection`
* Proposed content change, as Navie `question`

Returns:

JSON object containing:

* `original` snippet of content in the existing file that should be replaced with the proposed content change.
* `modified` code to replace the `existing` code with.

## CLI command `rpc-client`

Provides a way to invoke an RPC command without running a server. This would be useful e.g. in scripts that want to use `file.update` without having to run a server and connect.

This is needed because file updates can no longer be performed by invoking Navie, now that the @apply command is removed.

## Diagram

```mermaid
sequenceDiagram
    actor Client as Client (CLI)
    participant FileService as "RPC file.update"
    participant NavieProvider as Navie
    participant ComputeUpdateService as Compute Update Service
    participant LLM as LLM

    


    Client->>FileService: exists(file)
    activate FileService
    alt File does not exist
        FileService-->>Client: false
        Client->>FileService: writeFile(file, modified)
        FileService-->>Client: Write Success
        Client-->>Client: UpdateResponse(success)
    else File exists
        FileService-->>Client: true
        Client->>FileService: readFile(file)
        FileService-->>Client: original content
        Client->>NavieProvider: handler(file, modified)
        activate NavieProvider
        NavieProvider->>ComputeUpdateService: computeUpdate(original, modified)
        activate ComputeUpdateService
        ComputeUpdateService->>LLM: requestExplanation(original, modified)
        activate LLM
        LLM-->>ComputeUpdateService: explanation response
        ComputeUpdateService-->>NavieProvider: object
        deactivate LLM
        deactivate ComputeUpdateService
        alt Explain completed
            NavieProvider->>FileService: applyFileUpdate(file, original, modified)
            NavieProvider-->>Client: UpdateResponse(success)
        else Explain error
            NavieProvider-->>Client: UpdateResponse(failure)
        end
        deactivate NavieProvider
    end
    deactivate FileService
```

## How to try this

**Invoking the RPC `file.update` command**

```
node ./built/cli.js rpc-client v2.configuration.get "{}"
{
  "appmapConfigFiles": [
    "/Users/kgilpin/source/appland/appmap-js/packages/cli/appmap.yml"
  ],
  "projectDirectories": [
    "/Users/kgilpin/source/appland/appmap-js/packages/cli"
  ],
  "baseUrl": "https://api.openai.com"
}
```

**Invoking Navie `@update`**

```
node ./built/cli.js navie \
  -c /path/to/existing/file \
  '@update

<proposed file change - may be very fuzzy>
'

{
  "original": "    const completion = this.completionService.complete(messages, { temperature: 0 });\n    for await (const message of completion) {\n      tokens.push(message);\n    }",
  "modified": "    const completion = this.completionService.complete(messages, { temperature: 0 });\n    for await (const message of completion) {\n    }"
}
```

## TODO

- [x] Add the RPC method `file.update`
- [x] Also emit the `modified` code
- [x] Port logic in [FileUpdateService](https://github.com/getappmap/appmap-js/pull/1897/files#diff-83545f61e4534f3b0b929be0d0c6596c4361438e08cf8264f23414939990a73d) to RPC `file.update` method.
- [x] Test cases for Navie @update command

## Related

https://github.com/getappmap/appmap-js/issues/1895

### DIFF
--- a/packages/cli/src/rpc/file/update.ts
+++ b/packages/cli/src/rpc/file/update.ts
@@ -1,0 +1,141 @@
+import { ExplainRpc, FileRpc } from '@appland/rpc';
+import { RpcHandler } from '../rpc';
+import { exists } from '../../utils';
+import { mkdir, readFile, writeFile } from 'fs/promises';
+import { warn } from 'console';
+import { INavieProvider } from '../explain/navie/inavie';
+import { explain, explainStatus } from '../explain/explain';
+import applyFileUpdate from './applyFileUpdate';
+import { dirname } from 'path';
+
+function reportError(operation: string, e: Error | unknown, file: string) {
+  const tokens = [`[file-update] Failed to ${operation} ${file}`];
+  if (e instanceof Error) tokens.push(e.message);
+  warn(tokens.join(': '));
+}
+
+async function createFile(file: string, modified: string): Promise<FileRpc.UpdateResponse> {
+  try {
+    await mkdir(dirname(file), { recursive: true });
+    await writeFile(file, modified, 'utf-8');
+
+    return {
+      exists: true,
+      succeeded: true,
+      modified,
+    };
+  } catch (e) {
+    reportError('write file', e, file);
+
+    return {
+      exists: true,
+      succeeded: false,
+    };
+  }
+}
+
+async function updateFile(
+  navieProvider: INavieProvider,
+  file: string,
+  modified: string
+): Promise<FileRpc.UpdateResponse> {
+  // Read the original file
+  let original: string | undefined;
+  try {
+    original = await readFile(file, 'utf-8');
+  } catch (e) {
+    reportError('read file', e, file);
+
+    return {
+      exists: true,
+      succeeded: false,
+    };
+  }
+
+  const explainResponse = await explain(
+    navieProvider,
+    ['@update', modified].join('\n'),
+    original,
+    [],
+    undefined,
+    undefined,
+    undefined
+  );
+
+  let status: ExplainRpc.ExplainStatusResponse | undefined;
+
+  const awaitCompletion = async (interval: number, timeout: number) => {
+    const finalSteps = [ExplainRpc.Step.COMPLETE, ExplainRpc.Step.ERROR];
+    const complete = () => status && finalSteps.includes(status.step);
+
+    let remaining = timeout;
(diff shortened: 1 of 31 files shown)


packages/cli/src/rpc/file/update.ts
```diff
@@ -0,0 +1,139 @@
...
+    modified: result.modified,
+  };
+}
+
+export async function handler(
+  navieProvider: INavieProvider,
+  file: string,
+  modified: string
+): Promise<FileRpc.UpdateResponse> {
+  const fileExists = async () => exists(file);
```
dustinbyrne, line 123: Minor caveat here that this path must be absolute given the RPC service is not run with a particular working directory.

kgilpin: Yeah; I can’t really think of an alternative.

packages/cli/src/rpc/file/update.ts
```diff
@@ -0,0 +1,139 @@
...
+
+function reportError(operation: string, e: Error | unknown, file: string) {
+  const tokens = [`[file-update] Failed to ${operation} ${file}`];
+  if (e instanceof Error) tokens.push(e.message);
+  warn(tokens.join(': '));
+}
+
+async function createFile(file: string, modified: string): Promise<FileRpc.UpdateResponse> {
+  try {
+    await writeFile(file, modified, 'utf-8');
```
dustinbyrne, line 18: If we're creating new files, we should consider `mkdir(dirname(file), { recursive: true })` before writing the file. I've run in to cases where the LLM will add a new directory, and this writeFile will fail.

dustinbyrne (approved):

PR #1925, opened 2024-08-02
fix: "New chat" button doesn't destructively clear the chat anymore

### LINKED ISSUE #1922: "New Chat" button opens a new tab instead of clearing the chat
- [x] Support editor-provided function for new chat (https://github.com/getappmap/appmap-js/pull/1925)
- [x] Provide a function to open new chat in VSCode (https://github.com/getappmap/vscode-appland/pull/995)
- [x] Provide a function to open new chat in JetBrains (https://github.com/getappmap/appmap-intellij-plugin/pull/760)

dividedmind:
Instead of clearing the chat, "new chat" button now instead only shows when upstream (for example, a code editor) provides a method for that; if it does, it calls that. The intent is for the button to open a new tab instead.

Fixes #1922 (IDE patches to follow).

### DIFF
--- a/packages/components/src/components/chat/Chat.vue
+++ b/packages/components/src/components/chat/Chat.vue
@@ -6,11 +6,12 @@
         <div id="header-navie-logo">Navie</div>
       </div>
       <v-button
+        v-if="openNewChat"
         data-cy="new-chat-btn"
         class="clear"
         size="small"
         kind="ghost"
-        @click.native="clear"
+        @click.native="openNewChat"
       >
         New chat
       </v-button>
@@ -151,6 +152,9 @@ export default {
     VButton,
   },
   props: {
+    openNewChat: {
+      type: Function,
+    },
     // Initial question to ask
     question: {
       type: String,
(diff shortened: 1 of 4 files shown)


zermelo-wisen (approved):

packages/components/src/components/chat/Chat.vue
```diff
@@ -151,6 +152,9 @@ export default {
     VButton,
   },
   props: {
+    openNewChat: {
+      type: Function,
+    },
```
dustinbyrne, line 157: It looks like events were removed, but the pattern is still possible in Vue 3 with an external dependency
https://v3-migration.vuejs.org/breaking-changes/events-api#event-bus

The justification for dropping events seems warranted, but we don't really subscribe to the global (root) event bus, we only use it to publish events to an external source like VSCode. In many cases the events are so deeply nested that passing a function would be cumbersome. Inject/provide is an option, but TypeScript support is pretty poor, at least in Vue 2.

dustinbyrne (approved):

PR #1939, opened 2024-08-13
feat: Auto-completions and welcome message

dustinbyrne:
https://github.com/user-attachments/assets/d5806790-ec55-490c-9c7f-6b392f91cdbe

Sample:
https://5fe18baf66f2280021e634d0-whvhlpbixz.chromatic.com/?path=/story/pages-chatsearch--chat-search-mock

### DIFF
--- a/packages/cli/src/rpc/navie/metadata.ts
+++ b/packages/cli/src/rpc/navie/metadata.ts
@@ -1,0 +1,51 @@
+import { NavieRpc } from '@appland/rpc';
+import { RpcHandler } from '../rpc';
+
+export function navieMetadataV1(): RpcHandler<
+  NavieRpc.V1.Metadata.Params,
+  NavieRpc.V1.Metadata.Response
+> {
+  return {
+    name: NavieRpc.V1.Metadata.Method,
+    handler: () => ({
+      welcomeMessage:
+        "### Hi, I'm Navie!\n\nI can help you answer questions about your codebase, plan solutions, create diagrams, and generate code. Enter `@` to see a list of commands.",
+      inputPlaceholder: "Ask a question or enter '@' for commands",
+      commands: [
+        {
+          name: '@explain',
+          description:
+            'Ask questions about the codebase, and Navie will respond with explanations, diagrams, code snippets, and more. This is the default mode.',
+          referenceUrl: 'https://appmap.io/docs/reference/navie.html#explain',
+        },
+        {
+          name: '@diagram',
+          description: 'Navie can generate Mermaid diagrams based on your request.',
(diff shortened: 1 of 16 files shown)


kgilpin:
Thanks for getting this going to quickly!

any Navie feedback?

packages/cli/src/rpc/navie/metadata.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export function navieMetadataV1(): RpcHandler<
+  NavieRpc.V1.Metadata.Params,
+  NavieRpc.V1.Metadata.Response
+> {
+  return {
+    name: NavieRpc.V1.Metadata.Method,
+    handler: () => ({
+      welcomeMessage:
+        'Navie can help you answer questions about your codebase, plan solutions, and generate code. Enter `@` to see a list of commands.',
```
kgilpin, line 12: Do we want to change these to `/` commands at this point?

packages/components/src/components/chat/AutoComplete.vue
```diff
@@ -0,0 +1,205 @@
...
+
+      // activeText is the string from the start of the input to the cursor
+      const activeText = this.input.slice(0, endIndex);
+
+      // We need the word the cursor is on, so we split on whitespace and take the last token
+      const tokens = activeText.split(/\s+/);
+      return tokens[tokens.length - 1];
+    },
+    autoCompletions(): NavieRpc.V1.Metadata.Command[] {
+      if (!this.activeToken) return [];
```
kgilpin, line 99: Should we disable this feature completely if we don't have an activeToken or the RPC service doesn't support this command?

dustinbyrne:
> Thanks for getting this going to quickly!
> 
> any Navie feedback?

I didn't use Navie yet in this instance. The bulk of these changes are from two pre-existing branches that I've rebased and merged together. I plan on using Navie to put together the tests.

packages/cli/src/rpc/navie/metadata.ts
dustinbyrne: If possible, I don't want to tackle that as part of this change. All the pieces are in place so that we can update the welcome message and input placeholder returned by the `v1.navie.metadata` RPC method as a follow up.

packages/components/src/components/chat/AutoComplete.vue
dustinbyrne: Falling back to an empty array effectively does that here. Nothing is performed if `autoCompletions` is empty. Similarly, if the RPC method doesn't exist, `commands` will fall back to an empty array and effectively be a no-op, aside from computing `activeToken` above.

PR #1953, opened 2024-08-25
feat: Show diff when generating and applying changes

dividedmind:
## Add `original` parameter to RPC update method
If the original text is passed, the change is assumed to be a pregenerated search/replace change. In this case, the LLM is not asked to generate it; instead, it is applied directly.

## Generate changesets and display them in `@generate` command

If `/format-change` option is passed, a new prompt is used that causes the LLM to generate `<change>` tags instead of code snippets. This option is automatically added to `@generate` by components since it needs UI support to display and handle correctly. (The old method is used in its absence for backwards compatibility.)

The `<change>` tags are rendered streaming; once a change is complete, its verbatim text is used when the *apply* button is clicked.

## Allow apply command to take a preformatted changeset
The change file can be passed with a command line argument or on stdin, for example:
```console
$ jq .content -r ~/.appmap/navie/history/11390330-f0cb-47f6-839a-c1a8b4bcb22e/1716469088072.json  | appmap apply
No changes file specified, reading from stdin
Applied: /home/divide/projects/appmap-server/app/models/conversation_thread.rb
[file-update] Failed to find match for /home/divide/projects/appmap-server/app/models/conversation_thread.rb.
```

Note this is mostly useful in development for debugging problems with the patch application, however it could also be used in scripting.

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -7,7 +7,7 @@
     "serve": "vue-cli-service serve",
     "cli": "node src/cli.js",
     "build": "cross-env NODE_ENV=production rollup --config rollup.config.js --format es",
-    "watch": "cross-env NODE_ENV=production rollup --watch --config rollup.config.js --format es",
+    "watch": "cross-env rollup --watch --config rollup.config.js --format es",
     "test:unit": "vue-cli-service test:unit",
     "test:e2e": "cypress run",
     "test": "npm run test:unit && node .storybook/run.js",


packages/navie/src/agents/generate-agent.ts
```diff
@@ -55,7 +152,12 @@ export default class GenerateAgent implements Agent {
     const agentPrompt = [GENERATE_AGENT_PROMPT];
     // With the /noformat option, the user will explain the desired output format in their message.
     if (options.userOptions.isEnabled('format', true)) {
-      agentPrompt.push(GENERATE_AGENT_FORMAT);
+      agentPrompt.push(
+        // see if the new format is requested
+        options.userOptions.isEnabled('format-change', false)
```
kgilpin, line 157: I suggest `/format=xml`, this would align with some other agents that accept a format argument.

packages/cli/src/rpc/file/update.ts
```diff
@@ -83,51 +78,64 @@ async function updateFile(
...
+    original,
+    modified,
   };
 }
 
 export async function handler(
   navieProvider: INavieProvider,
   file: string,
-  modified: string
+  target: string,
```
kgilpin, line 119: I'm finding the name `target` a bit hard to understand. It looks like this is the same as `modified`?

kgilpin (approved):
It's hard for me to speak to the UI aspects of this, because I am not that facile with advanced changes there. LGTM otherwise, I did suggest a change in the option name to `/format`.

packages/cli/src/rpc/file/update.ts
dividedmind: It could be, or it could be a target snippet. I wanted to avoid the ambiguity of it being either an illustrative snippet with ellipses etc. or verbatim modified code.

dustinbyrne (approved):

packages/navie/src/agents/generate-agent.ts
dividedmind: Done!

dividedmind:
For the record, I changed the option to `/format=xml`, fixed the newline problem (sometimes apply would have added extra newlines) and made the copy button copy the modified code instead of the diff.

PR #1956, opened 2024-08-27
feat: Add tooltips to code snippet buttons

dustinbyrne:
This PR adds hover tooltips to code snippet / context items

https://github.com/user-attachments/assets/0acdd5d1-1f18-43b0-99dd-e26c8c522b1e

### DIFF
--- a/packages/components/src/components/chat/ContextContainer.vue
+++ b/packages/components/src/components/chat/ContextContainer.vue
@@ -15,24 +15,27 @@
     >
       <span class="context-container__title" data-cy="context-title">{{ title }}</span>
       <span class="context-container__button-group">
-        <span
-          :class="{
-            'context-container__button': 1,
-            'context-container__button--pending': !!pendingState,
-            'context-container__button--failure': pendingState === 'failure',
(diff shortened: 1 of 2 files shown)


packages/components/src/components/chat/ContextContainer.vue
```diff
@@ -15,24 +15,27 @@
...
+          >
+            <transition name="fade" mode="out-in">
+              <v-loader v-if="pendingState === 'pending'" />
+              <v-check-icon v-else-if="pendingState === 'success'" />
+              <v-close-icon v-else-if="pendingState === 'failure'" />
+              <v-apply-icon v-else />
+            </transition>
+          </span>
+        </v-popper>
+        <v-popper :time-to-display="250" text="Open" align="right" placement="top"></v-popper>
```
dividedmind, line 38: This seems duplicated.

dividedmind (approved):
Looks ok, thanks. Is there a good reason we're not using native tooltips, with `title`? This seems like it's an accessibility nightmare. I can't even begin to imagine how this sounds like on a screen reader.

PR #1968, opened 2024-09-02
Don't retry on 422 from OpenAI

### LINKED ISSUE #1954: Copilot completions tend to hang
Regardless of the model, generated tokens tend to hang after the first request. No indication of a malformed request is received. Usage of the `AsyncIterable` will cause the completion to hang as it awaits the first token.

This hang occurs in the `complete` method of `OpenAICompletionService`.

dividedmind:
This makes sure we don't try to retry on Copilot errors.

Along with https://github.com/getappmap/vscode-appland/pull/1010 this fixes #1954

### DIFF
--- a/packages/navie/src/services/openai-completion-service.ts
+++ b/packages/navie/src/services/openai-completion-service.ts
@@ -1,3 +1,5 @@
+import { isNativeError } from 'node:util/types';
+
 import { ChatOpenAI } from '@langchain/openai';
 import type { ChatCompletion, ChatCompletionChunk } from 'openai/resources';
 import { z } from 'zod';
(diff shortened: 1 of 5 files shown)


kgilpin:
Is this a Draft? Seems like it could use a test case.

dividedmind:
> Is this a Draft? Seems like it could use a test case.

Testing would probably need to involve creating a mock OpenAI server; I figured since testing would be difficult and the change is pretty much just modifying the default handler to also reject 422 it's ok to skip it. I can add it if you want.

kgilpin:
> > Is this a Draft? Seems like it could use a test case.
> 
> Testing would probably need to involve creating a mock OpenAI server; I figured since testing would be difficult and the change is pretty much just modifying the default handler to also reject 422 it's ok to skip it. I can add it if you want.

If you are comfortable with it as-is then it's fine with me.

packages/navie/src/services/openai-completion-service.ts
```diff
@@ -117,12 +119,50 @@ function estimateTokens(messages: readonly Message[]): number {
...
+      error.error &&
+      'code' in error.error &&
+      error.error.code === 'insufficient_quota'
+    ) {
+      const err = new Error(error.message);
+      err.name = 'InsufficientQuotaError';
+      throw err;
+    }
+  }
+}
```
dustinbyrne, line 157: To my eyes this looks like the same implementation with the additional 422 case, so, looking good. Is there a reason why any other 4xx might be retried?

dividedmind: 429 is rate limit, so retry with backoff could eventually get through: https://platform.openai.com/docs/guides/error-codes/api-errors

PR #1970, opened 2024-09-02
feat: Navie emits sent / received events for all LLM interactions

kgilpin:
Fixes https://github.com/getappmap/appmap-js/issues/1967
Fixes https://github.com/getappmap/appmap-js/issues/1969

### DIFF
--- a/packages/cli/src/cmds/navie.ts
+++ b/packages/cli/src/cmds/navie.ts
@@ -156,7 +156,6 @@ export function buildNavieProvider(argv: ExplainArgs) {
       const trajectory = new Trajectory(argv.trajectoryFile);
       navie.setTrajectoryHandler(trajectory);
 
-      process.on('SIGINT', () => trajectory.close());
       process.on('exit', () => trajectory.close());
     }
     applyAIOptions(navie);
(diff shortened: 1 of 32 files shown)


packages/cli/src/cmds/navie.ts
```diff
@@ -144,8 +150,15 @@ export function buildNavieProvider(argv: ExplainArgs) {
...
     loadConfiguration(false);
     const navie = new LocalNavie(contextProvider, projectInfoProvider, helpProvider);
+
     if (argv.threadId) navie.setThreadId(argv.threadId);
+    if (argv.trajectoryFile) {
+      const trajectory = new Trajectory(argv.trajectoryFile);
+      navie.setTrajectoryHandler(trajectory);
 
+      process.on('SIGINT', () => trajectory.close());
+      process.on('exit', () => trajectory.close());
```
dustinbyrne, line 160: Two things here:
1.) Binding a `SIGINT` listener will cause the default `SIGINT` handler to unbind itself. The process will no longer via this signal exit unless `process.exit` is called from within this listener.
2.) These listeners will be bound for every message sent. After 10 listeners, node will begin emitting warnings to the console about a potential memory leak. Consider binding these just once.

dustinbyrne (approved):
Looks good! Check out the comments below. I'm approving ahead of time because this behavior is hidden behind the `--trajactory-file` argument, and I don't see any issues that could affect end users.

packages/cli/src/cmds/navie.ts
kgilpin: OK, I think the exit handler is all we need. This will only be called once because it's a CLI entry function handler.

PR #1972, opened 2024-09-03
feat: Add `@observe` command for recording suggestions

dustinbyrne:
This pull request introduces a new `@observe` command to facilitate recording of test cases based on user inquiry. The user can specify some behavior they'd like to record and Navie will suggest a test case to run with AppMap enabled.

**Changes**:
1. **New Features**:
   - Introduced an `@observe` command to identify and suggest the most relevant test cases for recording AppMap trace data.
   
2. **Bug Fixes**:
   - Fixed a versioning issue in the `storybook` by correctly specifying the required version of `lru-cache`.
   
3. **Configuration Updates**:
   - Updated `package.json` and `yarn.lock` to reflect the corrected version constraints for dependencies.

**Testing**:
- Added unit tests to cover the new `@observe` command functionality.
- All existing tests have been run to ensure no breakage in current functionality.

Once released, the CLI will need to bump the included version of `@appland/navie` and add a command completion for `@observe` returned by the `v1.navie.metadata` RPC method.

### DIFF
--- a/packages/navie/src/commands/observe-command.ts
+++ b/packages/navie/src/commands/observe-command.ts
@@ -1,0 +1,253 @@
+import { z } from 'zod';
+
+import type Command from '../command';
+import type { CommandRequest } from '../command';
+import CompletionService from '../services/completion-service';
+import LookupContextService from '../services/lookup-context-service';
+import VectorTermsService from '../services/vector-terms-service';
+import { ContextV2 } from '../context';
+import { ExplainOptions } from './explain-command';
+import Message from '../message';
+import InteractionHistory, {
+  CompletionEvent,
+  PromptInteractionEvent,
+} from '../interaction-history';
+import { UserOptions } from '../lib/parse-options';
+import ProjectInfoService from '../services/project-info-service';
+
+const RelevantTest = z.object({
+  relevantTest: z
+    .object({
+      name: z.string().optional().describe('The name of the test case, if known'),
+      path: z.string().describe('The file path of the test file'),
+      language: z
+        .enum(['ruby', 'python', 'java', 'javascript', 'other'])
+        .describe('The programming language of the test file'),
+      framework: z.string().optional().describe('The test framework used'),
+    })
+    .optional()
+    .describe('A descriptor of the most relevant test to the requested behavior'),
+  installCommands: z
+    .array(
+      z.object({
+        command: z.string().describe('The command to execute'),
+        description: z.string().optional().describe('A description of the command'),
+      })
+    )
+    .optional()
+    .describe('An ordered list of terminal command(s) necessary to execute to install AppMap'),
+  testCommands: z
+    .array(
+      z.object({
+        command: z.string().describe('The command to execute'),
+        description: z.string().optional().describe('A description of the command'),
+      })
+    )
+    .optional()
+    .describe('The ordered list of terminal command(s) that can be executed to run the test'),
+});
+
+export default class ObserveCommand implements Command {
+  constructor(
+    private readonly options: ExplainOptions,
+    private readonly completionService: CompletionService,
+    private readonly lookupContextService: LookupContextService,
+    private readonly vectorTermsService: VectorTermsService,
+    private readonly interactionHistory: InteractionHistory,
+    private readonly projectInfoService: ProjectInfoService
+  ) {}
+
+  private async getLanguageDirective(userOptions: UserOptions): Promise<string> {
+    const language = userOptions.stringValue('language');
+    if (language) {
+      return `The language of the test must be ${language}.`;
+    }
+
+    const projectInfos = await this.projectInfoService.lookupProjectInfo();
+    if (projectInfos.length <= 1) {
+      const language = projectInfos?.[0]?.appmapConfig?.language;
+      return language
+        ? `The test case must be written in ${language}.`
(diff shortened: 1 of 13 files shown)


packages/navie/src/commands/record-command.ts
```diff
@@ -0,0 +1,141 @@
...
+          role: 'system',
+          content: prompt,
+        },
+        {
+          role: 'user',
+          content: userRequest,
+        },
+      ],
+      RelevantTest,
+      { model: ModelType.Full }
```
dustinbyrne, line 70: JSON completions don't work with the current `gpt-4o` pointer. Only the most recent model supports strict structured output in this manner. Maybe this can just use mini.

kgilpin: TBH we have been getting by just fine with extracting code from fences and interpreting the result.

packages/navie/src/commands/record-command.ts
```diff
@@ -0,0 +1,141 @@
...
+
+  async getMostRelevantTest(
+    userRequest: string,
+    fileContextItems: ContextV2.FileContextItem[]
+  ): Promise<z.infer<typeof RelevantTest>> {
+    const context = fileContextItems.map(
+      ({ content, location }) =>
+        `<code-snippet location="${location}">\n${content}\n</code-snippet>`
+    );
+    const prompt = `Given the following code snippets, identify the single most relevant test to the user request.
```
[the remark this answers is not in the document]
dustinbyrne: I'd originally tried using `InteractionHistory` but there was some reason I ended up not using it, though I don't remember the details. I can look into it if you prefer.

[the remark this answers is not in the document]
kgilpin: Yes please.

packages/navie/src/commands/observe-command.ts
```diff
@@ -0,0 +1,143 @@
...
+      `<document file="${doc.filePath}" from="${doc.from}" to="${doc.to}">${doc.content}</document>`
+  )
+  .join('\n')}
+</documentation>`,
+      },
+      {
+        role: 'user',
+        content: `Restate the information you've provided to me, in standalone format, as a step by step guide outlining the steps required to record the single test case that you've identified.
+If possible, include the terminal command needed to run the test. Avoid specifying file ranges in command line arguments. Only specify test patterns that are guaranteed to match based on previous context.
+Only provide instructions related to AppMap if they appear previously in the documentation. Ignore instructions that are phrased with tentative language such as "if you don't see this working". Assume everything works as expected.
```
kgilpin, line 133: If you require the AI to respond in a structured way - for example, give it an option to provide a terminal command, an option to specify an install command, or  code modification to activate the agent, then you may have less of a problem with free text instructions that include steps you don’t want it to include. 

And in a structured format, we would also have a better chance of being able to run the suggested command automatically.

packages/navie/src/commands/observe-command.ts
```diff
@@ -0,0 +1,143 @@
...
+        {
+          role: 'user',
+          content: userRequest,
+        },
+      ],
+      RelevantTest,
+      { model: ModelType.Mini }
+    );
+
+    return response?.relevantTest;
```
kgilpin, line 75: Is this expected to be a test file or a specific test?

There can be a lot of different syntax for how to run a specific test. So maybe we just go with the whole file?

packages/navie/src/commands/observe-command.ts
```diff
@@ -0,0 +1,143 @@
...
+    const vectorTerms = await this.vectorTermsService.suggestTerms(userRequest);
+    const tokenLimit = userOptions.numberValue('tokenlimit') || this.options.tokenLimit;
+    const testSnippets = await this.getTestSnippets(vectorTerms, tokenLimit);
+    const mostRelevantTest = await this.getMostRelevantTest(userRequest, testSnippets);
+    if (!mostRelevantTest) {
+      yield 'Sorry, I could not find any relevant tests to record.';
+      return;
+    }
+
+    if (mostRelevantTest.language === 'other') {
```
kgilpin, line 88: If the user has appmap.yml we would want to direct the LLM to ONLY consider tests in the configured language.

kgilpin: Specifying or overriding the language would also be a good thing to expose as an /option

packages/navie/src/commands/observe-command.ts
```diff
@@ -0,0 +1,143 @@
...
+  .map(
+    (doc) =>
+      `<document file="${doc.filePath}" from="${doc.from}" to="${doc.to}">${doc.content}</document>`
+  )
+  .join('\n')}
+</documentation>`,
+      },
+      {
+        role: 'user',
+        content: `Restate the information you've provided to me, in standalone format, as a step by step guide outlining the steps required to record the single test case that you've identified.
```
kgilpin, line 131: Based on the language and test framework, how much of this information is actually very structured and doesn’t need to be generated free form? Just something to think about.

packages/navie/src/commands/observe-command.ts
dustinbyrne: It returns a path to the file, so the command will typically run the entire file. There's an additional field for the name of the test, though it's optional, so sometimes the LLM will opt to suggest a command with a test filter to that name.

PR #1976, opened 2024-09-05
fix: Symlinks fall back to copying files on Windows

dustinbyrne:
This pull request addresses the issue of handling history on Windows by disabling the history functionality. Non-Administrator accounts cannot create symlinks by default on Windows, which is necessary for the current implementation of history.

#### Details:
- Introduced an `IHistory` interface to standardize history operations.
- Moved existing history-related functionality to the `History` class which implements `IHistory`.
- Created a `HistoryWindows` class that implements `IHistory` with no-op methods and throws a `ThreadAccessError` for unimplemented functionalities.
- Updated the initialization logic to conditionally use `HistoryWindows` on Windows platforms.
- Refactored references to `History` to use the new `IHistory` interface across the project.

### DIFF
--- a/packages/cli/src/rpc/explain/navie/historyWindows.ts
+++ b/packages/cli/src/rpc/explain/navie/historyWindows.ts
@@ -1,0 +1,21 @@
+import { ThreadAccessError } from './ihistory';
+import IHistory from './ihistory';
+import Thread from './thread';
+
+export default class HistoryWindows implements IHistory {
+  constructor(public readonly directory: string) {
+    console.warn('History is currently disabled. It is not yet implemented on Windows.');
+  }
+
+  token(): void {
+    // Do nothing
+  }
+
+  question(): void {
+    // Do nothing
+  }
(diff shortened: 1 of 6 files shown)


dustinbyrne:
Looking into one more potential issue

packages/cli/src/rpc/explain/navie/historyWindows.ts
```diff
@@ -0,0 +1,21 @@
...
+  token(): void {
+    // Do nothing
+  }
+
+  question(): void {
+    // Do nothing
+  }
+
+  load(threadId: string): Thread {
+    throw new ThreadAccessError(threadId, 'load', new Error('Not implemented on Windows'));
```
kgilpin, line 19: We just need to be aware that this will impact the UI if and when we use it there.

kgilpin (approved):
LGTM my only question is if the load thread error is going to bite us anywhere.

PR #1979, opened 2024-09-08
Pin files to context

apotterri:
Give the user the ability to pin files to the Chat context, including through drag-and-drop.

There are a couple of tests here, but this will be further tested by the e2e tests in the vscode-appland changes that use it.

### DIFF
--- a/packages/components/src/components/chat-search/Context.vue
+++ b/packages/components/src/components/chat-search/Context.vue
@@ -1,6 +1,6 @@
 <template>
   <div class="context" data-cy="context">
-    <template v-if="!contextResponse">
+    <template v-if="!contextResponse && !hasPinnedItems">
       <div class="context__notice-container">
         <div class="context__notice" data-cy="context-notice">
           <h2>Navie's context</h2>
@@ -15,7 +15,7 @@
     <template v-else>
       <div class="context__body">
         <h2>Context</h2>
-        <template v-if="pinnedItems && pinnedItems.length">
+        <template v-if="hasPinnedItems">
           <h3>Pinned Items</h3>
           <div class="context__body__table">
             <component
@@ -54,6 +54,7 @@
 <script lang="ts">
 //@ts-nocheck
 import VContextItem from '@/components/chat-search/ContextItem.vue';
+import VFile from '@/components/chat/File.vue';
 import VMarkdownCodeSnippet from '@/components/chat/MarkdownCodeSnippet.vue';
 import VMermaidDiagram from '@/components/chat/MermaidDiagram.vue';
 import VButton from '@/components/Button.vue';
@@ -61,14 +62,16 @@ import VButton from '@/components/Button.vue';
 const PinnedContextComponents = {
   'code-snippet': VMarkdownCodeSnippet,
   mermaid: VMermaidDiagram,
+  file: VFile,
 };
 
 export default {
   name: 'v-context',
 
   components: {
-    VContextItem,
     VButton,
+    VContextItem,
+    VFile,
     VMarkdownCodeSnippet,
     VMermaidDiagram,
   },
@@ -102,11 +105,14 @@ export default {
     contextTypeKeys() {
       return Object.keys(this.contextTypes);
     },
+    hasPinnedItems() {
+      return this.pinnedItems?.length > 0;
+    },
   },
 
   methods: {
     hasContext(type: string) {
-      return this.contextResponse.some((context) => context.type === type);
+      return this.contextResponse?.some((context) => context.type === type);
     },
     contextItems(type: string) {
       return this.contextResponse.filter((context) => context.type === type);
--- a/packages/components/src/components/chat/ContextContainer.vue
+++ b/packages/components/src/components/chat/ContextContainer.vue
@@ -1,14 +1,18 @@
 <template>
   <div
-    :class="{ 'context-container': 1, 'context-container--collapsed': collapsed }"
+    :class="{
+      'context-container': 1,
+      'context-container--collapsed': collapsed || !isAppliable,
+    }"
     data-cy="context-container"
     :data-handle="valueHandle"
     :data-reference="isReference"
+    :title="location"
   >
     <div
       :class="{
         'context-container__header': 1,
-        'context-container__header--collapsable': isCollapsable,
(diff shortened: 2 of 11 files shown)


packages/components/src/components/chat/ContextContainer.vue
```diff
@@ -217,7 +221,12 @@ export default Vue.extend({
       this.closeMenu();
     },
     onClickHeader() {
-      if (!this.isCollapsable) return;
+      if (!this.isCollapsable) {
+        if (this.isFile) {
+          this.$root.$emit('open-location', this.location, this.directory);
+        }
+        return;
+      }
```
dustinbyrne, line 229: At the moment there's no indication that these are clickable. Consider adding the `cursor: pointer` style to the element.

packages/components/src/lib/PinFileRequest.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * @class
+ */
+export class PinFileRequest {
+  /**
+   * @param {string} name - The name of the file.
+   * @param {string} [uri] - The URI of the file (optional).
+   * @param {string} [content] - The content of the file (optional).
+   */
```
dustinbyrne, line 9: These can be TypeScript as well

packages/components/src/components/chat-search/Context.vue
```diff
@@ -54,13 +54,15 @@
...
 import VContextItem from '@/components/chat-search/ContextItem.vue';
+import VFile from '@/components/chat/File.vue';
 import VMarkdownCodeSnippet from '@/components/chat/MarkdownCodeSnippet.vue';
 import VMermaidDiagram from '@/components/chat/MermaidDiagram.vue';
 import VButton from '@/components/Button.vue';
 
 const PinnedContextComponents = {
   'code-snippet': VMarkdownCodeSnippet,
   mermaid: VMermaidDiagram,
+  file: VFile,
```
dustinbyrne, line 65: I'd expect this to be registered within `components` below, is this working?

packages/components/src/components/chat-search/Context.vue
```diff
@@ -15,7 +15,7 @@
     <template v-else>
       <div class="context__body">
         <h2>Context</h2>
-        <template v-if="pinnedItems && pinnedItems.length">
+        <template v-if="hasPinnedItems">
           <h3>Pinned Items</h3>
           <div class="context__body__table">
             <component
```
dustinbyrne, line 21: Somehow the context is being written to an attribute as a literal, but I don't see that change here.
image

packages/components/src/components/chat/ContextContainer.vue
apotterri: Will do.

packages/components/src/lib/PinFileRequest.js
apotterri: Ok, I thought I tried that some point, and it caused a build failure. I just tried it now, though, and it worked ok. Could be that I moved it from somewhere else into `lib`, or that I was trying to import it into a different file (e.g. a `vue`) file....

packages/components/src/components/chat-search/Context.vue
apotterri: [This](https://vuejs.org/guide/components/registration) seems to suggest that components need to be registered so when they're used within a template, Vue knows how to find the implementation. `v-file` doesn't get used in this (or any other) template, which is probably why it works (which I believe you've now seen).

Doesn't seem like it would do any harm to add it to `components`, though, and that would probably bring it more in line with standard practice.

packages/components/src/components/chat-search/Context.vue
apotterri: The code that renders pinned items assumes that each item has a `content` property: https://github.com/getappmap/appmap-js/blob/pin-file_20240825/packages/components/src/components/chat/PinEvent.ts#L10 . When a file gets pinned, the `PinFile.content` for it gets set to the content of the file.

The other changes I made in this file insure that the content for a pinned file doesn't get revealed when you click on it.

apotterri:
**Please don't merge this until the `fixup!` commits get squashed.**

packages/components/src/components/chat/ContextContainer.vue
```diff
@@ -4,11 +4,12 @@
     data-cy="context-container"
     :data-handle="valueHandle"
     :data-reference="isReference"
+    :title="location"
   >
     <div
       :class="{
         'context-container__header': 1,
-        'context-container__header--collapsable': isCollapsable,
+        'context-container__header--collapsable': isCollapsable || !isAppliable,
```
dustinbyrne, line 12: There's a 1px border on the bottom of this header that shouldn't be present

apotterri: Should never be present, or shouldn't be present for items that aren't appliable (e.g. files)?

packages/components/src/pages/ChatSearch.vue
```diff
@@ -4,6 +4,10 @@
     @mousemove="makeResizing"
     @mouseup="stopResizing"
     @mouseleave="stopResizing"
+    @drop="onDrop"
+    @dragover.prevent
+    @dragevent.prevent
```
dustinbyrne, line 9: The event handler `dragevent` is not a standard DOM event. The correct event for handling drag operations is `dragover`, which is already present.

\- Navie

packages/components/src/pages/ChatSearch.vue
```diff
@@ -497,6 +503,47 @@ export default {
...
+              });
+            })
+          );
+        } else if (item.kind === 'file') {
+          const file = item.getAsFile();
+          const name = 'file:///' + file.name;
+          promises.push(Promise.resolve([{ name, file }]));
+        }
+      }
+      await Promise.all(promises).then((requests) => {
```
dustinbyrne, line 533: The `onDrop` method lacks error handling for asynchronous file reading and URI processing, which can lead to unhandled promise rejections.
\- Navie

DB: After looking through MDN, I didn't see any explicit mention of exceptions raised by the methods called, maybe it's not a consideration.

dustinbyrne (approved):

PR #1989, opened 2024-09-13
fix: Ignore `@suggest` request and completions when preparing chat history

dustinbyrne:
Considering the default limit is two messages, this can (will) completely overwrite any useful information.

This is more of an intermediate stop-gap to mitigate the issue. Fixing this properly will require some deeper abstraction.

### DIFF
--- a/packages/navie/src/lib/get-most-recent-messages.ts
+++ b/packages/navie/src/lib/get-most-recent-messages.ts
@@ -14,6 +14,22 @@ const DEFAULT_HISTORICAL_MESSAGES = (() => {
   return 2;
 })();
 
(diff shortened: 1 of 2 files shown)


kgilpin:
How about a `/history` / `/nohistory` option that will exclude a message (and the response) from the history? Then we could apply that automatically to @suggest.

dustinbyrne:
I'll play around with this idea. Though, `/history` and `/nohistory` are used in explain to set the number of historical messages to include in a completion. I can incorporate something similar.

I'm also going to merge this in the interim and put it on the server. That'll stop the issues we're seeing now, and I can follow up with a replacement incorporating the changes described above.

PR #2000, opened 2024-09-18
fix: Retry completion if OpenAI throws a server error while streaming

dividedmind:
This commit introduces a retry mechanism for handling server errors during the streaming of OpenAI API responses. Let's break down the changes and their implications:

## Error Handling and Retry Logic:

The main change is in the `complete` method of the `OpenAICompletionService` class. A new retry loop has been added to handle potential server errors during streaming.

```typescript
for (let attempt = 0; attempt < 5; attempt++) {
  const response = await fetchResponse();
  try {
    for await (const token of response) {
      // ... process tokens ...
    }
    break; // Exit loop if successful
  } catch (error) {
    if (!(error instanceof APIError) || error.type !== 'server_error') throw error;
    if (tokensYielded || attempt === 4) throw error;
    await new Promise((resolve) => setTimeout(resolve, 1000 * Math.pow(2, attempt)));
  }
}
```

This code attempts to fetch and process the response up to 5 times. If a server error occurs, it will retry with an exponential backoff delay.

## Error Handling Conditions:

- Only retry on server errors (`error.type === 'server_error'`).
- Stop retrying if tokens have already been yielded or if it's the last attempt.

## Exponential Backoff:

The retry delay increases exponentially with each attempt, starting at 1 second and doubling each time.

## Test Case:

A new test case has been added to verify this behavior:

```typescript
it('retries on server error with exponential backoff during response iteration', async () => {
  // ... test implementation ...
});
```

This test simulates a server error on the first attempt and a successful response on the second attempt.

## Import Changes:

The `APIError` from the 'openai' package is now imported to properly type-check the error handling.

# Implications:

1. Improved Reliability: This change makes the service more resilient to temporary server issues from OpenAI.

2. Potential Delay: In case of server errors, the response might be delayed due to the retry mechanism.

3. Partial Responses: The code now handles cases where some tokens have been yielded before an error occurs, preventing data loss or inconsistencies.

4. Resource Usage: The retry mechanism might lead to increased API calls in error scenarios, which could affect usage limits or costs.

Overall, this is a robust improvement to handle transient server errors from the OpenAI API, enhancing the reliability of the completion service.

### DIFF
--- a/packages/navie/src/services/openai-completion-service.ts
+++ b/packages/navie/src/services/openai-completion-service.ts
@@ -8,6 +8,8 @@ import { warn } from 'console';
 import Message, { CHARACTERS_PER_TOKEN } from '../message';
 import CompletionService, {
   Completion,
+  CompletionRetries,
+  CompletionRetryDelay,
   convertToMessage,
   JsonOptions,
   mergeSystemMessages,
@@ -296,7 +298,7 @@ export default class OpenAICompletionService implements CompletionService {
       });
     };
 
-    for (let attempt = 0; attempt < 5; attempt++) {
+    for (let attempt = 0; attempt < CompletionRetries; attempt++) {
       const response = await fetchResponse();
       try {
         for await (const token of response) {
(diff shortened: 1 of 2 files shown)


dustinbyrne (approved):

kgilpin (approved):

kgilpin:
Nice use of the Navie description

PR #2001, opened 2024-09-18
feat: Expand symbols parsed from Python source

dustinbyrne:
This now includes variables, type references, comments.

An example file with parsed keywords is available here:
https://gist.github.com/dustinbyrne/e208b9bfcb8ed106cef370454cbf8772


packages/cli/tests/unit/fulltext/querySymbols.spec.ts
```diff
@@ -8,27 +8,60 @@ describe('querySymbols', () => {
       const symbols = querySymbols(srcPath);
       expect(symbols.sort()).toStrictEqual(
         [
-          'ClassOne',
```
kgilpin, line 11: It's very important for the index that CamelizedWords and snake_case be left as-is. Because it's relying on that structure to be able to split words apart.

packages/cli/src/fulltext/querySymbols.ts
```diff
@@ -25,6 +27,18 @@ const genericSymbolRegex =
...
+  const results = sanitizeKeyword(symbol)
+    .flatMap((t) =>
+      splitCamelized(t)
+        .split(/[\s-_]/)
+        .map((s) => s.toLowerCase())
+    )
+    .filter((t) => t.trim().length > 1);
+  const unique = new Set([...results, symbol.toLowerCase()]);
+  return Array.from(unique).filter((t) => !STOP_WORDS.has(t));
+}
```
kgilpin, line 40: I think this happens when the words are added to the index, so does it also need to happen here?

kgilpin:
@dustinbyrne how should we proceed with this?

dustinbyrne:
Considering https://github.com/getappmap/appmap-js/issues/2038, I think it may be better to implement a secondary tokenizing strategy with different weighting as we've previously discussed.

Something as simple as splitting on `/[\w\d\-_]{3,}/` and filtering out language specific keywords would seem sufficient.

kgilpin:
Ok. See approach taken here - https://github.com/getappmap/appmap-js/tree/feat/upgrade-search

PR #2003, opened 2024-09-19
Support O1

dividedmind:
Here is a summarized description of the changes made to support the `o1-mini` and `o1-preview` models:

### Summary of Changes

1. **Add Cost Per Token for New Models**
   - Added entries for `o1-mini` and `o1-preview` models in `COST_PER_M_TOKEN`.
   
   
   ```ts
   const COST_PER_M_TOKEN: Record<string, { input: number; output: number }> = {
     ...
     'o1-preview': {
       input: 15,
       output: 60,
     },
     'o1-preview-2024-09-12': {
       input: 15,
       output: 60,
     },
     'o1-mini': {
       input: 3,
       output: 12,
     },
     'o1-mini-2024-09-12': {
       input: 3,
       output: 12,
     },
     ...
   };
   ```

2. **Merge System Messages and Log Sent Messages**
   - Added model-specific message handling for `o1-preview` and `o1-mini` models.
   
   
   ```ts
   const sentMessages = mergeSystemMessages(messages, isO1 ? 'user' : 'system');
   for (const message of sentMessages) this.trajectory.logSentMessage(message);
   
   const response = await (isO1
     ? // o1 currently doesn't support streaming or temperatures != 1
       this.model.completionWithRetry({
         messages: sentMessages,
         model,
         stream: false,
         temperature: 1,
       })
     : this.model.completionWithRetry({
         messages: sentMessages,
         model,
         stream: true,
         temperature: options?.temperature,
         stream_options: { include_usage: true },
       }));
   ```

These changes enhance the completion service by adding configurations and handling processes specific to the `o1-mini` and `o1-preview` models. This support ensures accurate cost tracking and message handling for these new models.


packages/navie/src/services/openai-completion-service.ts
```diff
@@ -279,15 +279,13 @@ export default class OpenAICompletionService implements CompletionService {
     return undefined;
   }
 
-  async *complete(messages: readonly Message[], options?: { temperature?: number }): Completion {
-    const promptTokensPromise = this.countTokens(messages);
-
+  async *complete(messages: readonly Message[], options?: JsonOptions): Completion {
```
dustinbyrne, line 292: If `JsonOptions` is now used in both methods, we should consider renaming it

packages/navie/src/services/openai-completion-service.ts
```diff
@@ -279,43 +289,60 @@ export default class OpenAICompletionService implements CompletionService {
...
 
-    const response = await this.model.completionWithRetry({
-      messages: sentMessages,
-      model: this.model.modelName,
-      stream: true,
-      temperature: options?.temperature,
-    });
+    const response = await (isO1
+      ? // o1 currently doesn't support streaming or temperatures != 1
+        this.model.completionWithRetry({
```
dustinbyrne, line 301: Do we know what langchains retry logic is if we hit our maximum RPM? Is there anything else we need to do here?

kgilpin: Retry definitely important to understand.

kgilpin:
So, to test this out in the benchmark:

1) o1 branch off the default branch
2) update appmap-js 
3) run the benchmark with “smoke” instance set

packages/navie/src/services/openai-completion-service.ts
dividedmind: Langchain uses p-retry with 6 exponentially backing off retries by default; `onFailedAttempt()` above decides whether an error is fatal or retryable.  I'm not sure what OpenAI will return on rate limit, but this is where we'll need to adjust if we encounter issues.

kgilpin:
Ok. So let’s keep working on this until the patches come out right.

dustinbyrne (approved):

PR #2004, opened 2024-09-19
fix: Retry on Claude overload error

dustinbyrne:
This pull request addresses an issue with the AnthropicCompletionService where requests would fail due to an overload error from Claude's API. The following changes have been implemented to resolve this issue:

- **Retry Mechanism for SSE Errors:** 
  - Introduced a method to parse SSE errors from error messages.
  - Implemented a retry mechanism within the `complete` method to handle overload errors by retrying up to a maximum of five attempts with exponential backoff.
  
- **Unit Tests:**
  - Enhanced the existing test suite to include a test that verifies the retry behavior when an overload error occurs.
  - Updated mocks and test utilities to support the new retry logic.

- **Dependency Updates:**
  - Updated the version of `@langchain/anthropic` to v0.3.1.

#### Testing

- **Automated Tests:** 
  - Added a new test case to ensure that the service retries requests on overload errors and ultimately throws an appropriate error if all retries fail.
  - All existing tests have been updated and validated to ensure compatibility with the changes.

- **Manual Testing:** 
  - Verified through manual testing that the retry mechanism works as expected, with appropriate logging and error handling.

#### Impact

These changes improve the robustness of the `AnthropicCompletionService` by handling API overload errors more gracefully, ensuring a more resilient interaction with the Claude API. The dependency updates also bring in the latest improvements and fixes from the updated libraries.

Fixes https://github.com/getappmap/appmap-js/issues/2002

### DIFF
--- a/packages/navie/src/services/anthropic-completion-service.ts
+++ b/packages/navie/src/services/anthropic-completion-service.ts
@@ -7,6 +7,7 @@ import { z } from 'zod';
 import Message from '../message';
 import CompletionService, {
   Completion,
+  CompletionRetries,
   convertToMessage,
   JsonOptions,
   mergeSystemMessages,
@@ -161,7 +162,7 @@ export default class AnthropicCompletionService implements CompletionService {
     const tokens = new Array<string>();
     for (const message of sentMessages) this.trajectory.logSentMessage(message);
 
-    const maxAttempts = 5;
+    const maxAttempts = CompletionRetries;
     for (let attempt = 0; attempt < maxAttempts; attempt += 1) {
       try {
         // eslint-disable-next-line no-await-in-loop
--- a/packages/navie/src/services/anthropic-completion-service.ts
+++ b/packages/navie/src/services/anthropic-completion-service.ts
@@ -8,6 +8,7 @@ import Message from '../message';
 import CompletionService, {
   Completion,
   CompletionRetries,
+  CompletionRetryDelay,
   convertToMessage,
   JsonOptions,
   mergeSystemMessages,
@@ -188,7 +189,7 @@ export default class AnthropicCompletionService implements CompletionService {
         if (attempt < maxAttempts - 1 && tokens.length === 0) {
           const sseError = getSseError(cause);
           if (sseError) {
-            const nextAttempt = 1000 * 2 ** attempt;
+            const nextAttempt = CompletionRetryDelay * 2 ** attempt;
             warn(`Received ${JSON.stringify(sseError.error)}, retrying in ${nextAttempt}ms`);
 
             // eslint-disable-next-line no-await-in-loop
--- a/packages/navie/src/services/anthropic-completion-service.ts
+++ b/packages/navie/src/services/anthropic-completion-service.ts
@@ -70,6 +70,29 @@ const COST_PER_M_TOKEN: Record<string, { input: number; output: number }> = {
   },
 };
 
+type SSEError = {
+  type: 'error';
+  error: {
+    type: string;
+    message: string;
+  };
+};
+
+// Parses Claude's SSE error format from an error message
+function getSseError(e: unknown): SSEError | undefined {
+  if (e instanceof Error) {
+    if ('status' in e && e.status === undefined) {
+      try {
+        const error = JSON.parse(e.message) as SSEError;
+        return error;
+      } catch {
+        // ignore
+      }
+    }
+  }
+  return undefined;
+}
+
 export default class AnthropicCompletionService implements CompletionService {
   constructor(
     public readonly modelName: string,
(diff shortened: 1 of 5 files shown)


dustinbyrne:
I was able to recreate the issue locally with a mock Anthropic API
https://gist.github.com/dustinbyrne/d03ec0b62bf533d447ea4982c7978fc8

packages/navie/src/services/anthropic-completion-service.ts
```diff
@@ -132,39 +155,59 @@ export default class AnthropicCompletionService implements CompletionService {
...
-      for (const message of sentMessages) this.trajectory.logSentMessage(message);
-
-      const response = await model.stream(sentMessages.map(convertToMessage));
+    const usage = new Usage(COST_PER_M_TOKEN[this.modelName]);
+    const model = this.buildModel(options);
+    const sentMessages = mergeSystemMessages(messages);
+    for (const message of sentMessages) this.trajectory.logSentMessage(message);
 
-      const usage = new Usage(COST_PER_M_TOKEN[this.modelName]);
+    const maxAttempts = 5;
```
kgilpin, line 163: Can this be an env var? Maybe used in other retry situations too.

kgilpin: With GPT 1o we will also want substantial retry back off.

dustinbyrne: We should update https://github.com/getappmap/appmap-js/pull/2000 as well (cc @dividedmind)

packages/navie/src/services/anthropic-completion-service.ts
```diff
@@ -132,39 +157,59 @@ export default class AnthropicCompletionService implements CompletionService {
...
-          usage.completionTokens += usage_metadata.output_tokens;
+        this.trajectory.logReceivedMessage({
+          role: 'assistant',
+          content: tokens.join(''),
+        });
+
+        break;
+      } catch (cause) {
+        if (attempt < maxAttempts - 1) {
+          const sseError = getSseError(cause);
```
dividedmind, line 191: You have to throw if you've already yielded any tokens. Compare with https://github.com/getappmap/appmap-js/pull/2000 (which is the same issues mod specifics of the error); btw, I wonder if we should DRY this up somehow, we're basically maintaining two separate implementation of almost the same thing.

packages/navie/src/services/completion-service.ts
```diff
@@ -83,3 +83,29 @@ export function convertToMessage(message: Message): BaseMessage {
   }
   return new Cons({ content });
 }
+
+// Number of retries to attempt a completion before giving up
+export const CompletionRetries = (() => {
+  const env = process.env.APPMAP_NAVIE_COMPLETION_RETRIES;
+  if (env) {
+    const value = parseInt(env, 10);
+    if (Number.isInteger(value) && value > 0) {
```
dividedmind, line 92: Or, just `if (value)`? In general these two functions are quite verbose for no good reason (that I can see). How about `export const CompletionRetries = parseInt(process.env.APPMAP_NAVIE_COMPLETION_RETRIES) || 5`?

packages/navie/test/services/anthropic-completion-service.spec.ts
```diff
@@ -72,6 +75,58 @@ describe('AnthropicCompletionService', () => {
...
+          /* eslint-enable jest/no-conditional-expect */
+        }
+      })();
+
+      const delays = [1000, 2000, 4000, 8000];
+
+      for (const delay of delays) {
+        jest.advanceTimersByTime(delay);
+
+        // Yield to the event loop to allow another attempt to be made
```
dividedmind, line 123: Ah, that's clever! I was trying to figure out a clean way to do this before and didn't think of this.

dividedmind (changes requested):

packages/navie/src/services/completion-service.ts
dustinbyrne: If the user's intent is to disable retries, `parseInt(...) || 5` doesn't allow the value to be set to zero. Similarly, `parseInt(...) ?? 5` allows the value to be set to `NaN`.

This function is slightly verbose, but it catches these corner cases.

packages/navie/src/services/anthropic-completion-service.ts
dustinbyrne: Good point. I've changed it to bail if any tokens have been written. I haven't made an effort to DRY this up as of yet, I'd like to see it working first.

kgilpin (approved):
Worked well in the benchmark!

packages/navie/src/services/anthropic-completion-service.ts
kgilpin: It definitely seemed to be working in the benchmark runs. Don't make any risky changes that would break it again :-)

PR #2005, opened 2024-09-19
fix: show pinning help

### LINKED ISSUE #1988: Context area explains the operation of "Pinned" content
When there is no Pinned content, the Context area should explain the concept of Pinned content and how to add pinned content.

1) Use the Pin button from a Navie answer
2) Drag and drop a file (Using Shift)
3) Use the command "AppMap: Add to Context"

### LINKED ISSUE #2011: When switching to Navie webview, the input area should get focus
The input area doesn't get focus. So, it's necessary to move the mouse over to the input area and click it before typing anything.

When switching between Navie webview and code, it's cumbersome to have to do so.

--

* Implement in the components package

apotterri:
Show help text when there are no pinned items in the Context.

Fixes #1988 .
Fixes #2011 .


kgilpin:
Calling the entire area "Navie's Context Sources" and then calling a subsection of that area "Navie's Context" is confusing.

Screen Shot 2024-10-02 at 9 43 30 AM

Remove the major heading "Navie's Context Sources".

kgilpin:
The (info) pin disappears when the search context arrives:

Screen Shot 2024-10-02 at 9 47 03 AM

packages/components/src/components/chat-search/IntelliJNotice.vue
```diff
@@ -0,0 +1,46 @@
+<template>
+  <div>
+    <button @click.prevent="addFiles">&#xFF0B; add files</button>
```
dustinbyrne, line 3: This should be an SVG, otherwise we have no control over how this character is rendered
image

dustinbyrne: Also, we can't have unstyled buttons. At least use the `v-button` component.

packages/components/src/components/chat/ChatInput.vue
```diff
@@ -116,6 +121,14 @@ export default {
     isSelectingCommand() {
       return this.$refs.autocomplete.isVisible;
     },
+    inputClass() {
+      return {
+        // See the comment in getappmap/vscode-appland/web/src/app.js for why `focus` is applied to
+        // the input
+        focus: true,
```
dustinbyrne, line 128: If we truly have to do this, use a data attribute and not an empty class

packages/components/src/components/AddFileButton.vue
```diff
@@ -0,0 +1,27 @@
...
+  </v-button>
+</template>
+
+<script lang="ts">
+//@ts-nocheck
+import PlusIcon from '../assets/plus.svg';
+export default {
+  name: 'v-chat-search',
+  components: {
+    PlusIcon,
```
dustinbyrne, line 13: `v-button` is not registered so the button does not render

dustinbyrne: image

packages/components/src/components/AddFileButton.vue
```diff
@@ -0,0 +1,27 @@
+<template>
+  <v-button class="clear add-file__button" size="small" kind="ghost">
```
dustinbyrne, line 2: `clear` is not a class

packages/components/src/components/chat-search/Context.vue
```diff
@@ -1,9 +1,8 @@
 <template>
   <div class="context" data-cy="context">
-    <template v-if="!contextResponse && !hasPinnedItems">
+    <template v-if="!contextResponse">
       <div class="context__notice-container">
```
dustinbyrne, line 4: The bottom area of the context notice is clipped
image

packages/components/src/pages/ChatSearch.vue
```diff
@@ -655,6 +789,27 @@ $border-color: darken($gray4, 10%);
...
+      display: flex;
+      flex-direction: column;
+      overflow-y: auto;
+      margin-bottom: 0;
+    }
+
+    .pinned-items-wrapper .search-container {
+      max-height: 33vh;
+      border-bottom: 1px dashed $white;
+    }
```
dustinbyrne, line 810: This border does not lay on the bottom margins of the 'pinned items' container, so it makes the 'selected by Navie' header appear have a lot of a lot of margin on top.
image

image

packages/components/src/pages/ChatSearch.vue
```diff
@@ -39,38 +40,89 @@
...
+      <div class="context-wrapper">
+        <h2>Selected by Navie</h2>
+        <div class="search-container">
+          <v-context
+            class="no-match-instructions"
+            :appmap-stats="appmapStats"
+            :context-response="contextResponse"
+          />
+        </div>
+      </div>
```
dustinbyrne, line 93: These aren't scrolling the correct area. The scroll bars should span the entire height of the container. The container should also not scroll unless the content is overflowing.
image

packages/components/src/pages/ChatSearch.vue
```diff
@@ -655,6 +789,27 @@ $border-color: darken($gray4, 10%);
...
+    .context-wrapper {
+      display: flex;
+      flex-direction: column;
+      overflow-y: auto;
+      margin-bottom: 0;
+    }
+
+    .pinned-items-wrapper .search-container {
+      max-height: 33vh;
+      border-bottom: 1px dashed $white;
```
dustinbyrne, line 809: `$white` is too bright here - try something like `rgba($white, 0.33)`

packages/components/src/pages/ChatSearch.vue
dustinbyrne: Notice the placement and sizing of the scrollbars in these two images. The container would ideally take up its maximum height even before filling up with context items.

image

image

apotterri: Yeah, that's the problem I fixed. I thought you were saying there was an additional problem, but maybe not?

dustinbyrne: It looks better now

packages/components/src/pages/ChatSearch.vue
```diff
@@ -423,9 +493,7 @@ export default {
         this.ask.on('stop', onStop);
 
         const explainRequest = {
-          question: message
-            .replace(/^@generate/, '@generate /format=xml')
-            .replace(/^@test/, '@test /format=xml'),
```
dustinbyrne, line 428: Rebase issue here? I think `@test /format=xml` should be kept.

apotterri:
Yeah, that definitely wasn't a change I intended. I'll make sure not to
merge it.

packages/components/src/pages/ChatSearch.vue
apotterri: I went back over my reflog and found where this change was introduced. It doesn't appear to have been part of a rebase, though the subject of the commit message is a little weird: it's labeled `!fixup !fixup`. I try to make a point of never fixing up a `!fixup` commit, so I'm not entirely sure how this happened. I think I missed it because it sort of looked like a formatting change, and I didn't notice that it was actually a behavior change.

In any event, this was the only unintentional change introduced in the `!fixup !fixup` commit. I also went back over all the changes in the PR, and didn't see anything else that looked out of place.

LMK if you have any other concerns I can address. Otherwise, I think this is ready to go.

dustinbyrne (approved):

PR #2021, opened 2024-10-01
feat: Ignore .appmap and .navie directories when selecting context

kgilpin:
These can be used to store metadata and information used by Navie itself, such as prompts and snippets, without risk of that material being selected as context.

### DIFF
--- a/packages/cli/tests/unit/rpc/explain/pattern.spec.ts
+++ b/packages/cli/tests/unit/rpc/explain/pattern.spec.ts
@@ -1,0 +1,27 @@
+import {
+  EXCLUDE_DOT_APPMAP_DIR,
+  EXCLUDE_DOT_NAVIE_DIR,
+} from '../../../../src/rpc/explain/collectContext';
+
+describe('Regex patterns', () => {
+  const testCases = [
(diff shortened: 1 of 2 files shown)


packages/cli/tests/unit/rpc/explain/pattern.spec.ts
```diff
@@ -0,0 +1,25 @@
+describe('Regex patterns', () => {
+  const EXCLUDE_DOT_APPMAP_DIR = /(^|[/\\])\.appmap([/\\]|$)/;
+  const EXCLUDE_DOT_NAVIE_DIR = /(^|[/\\])\.navie([/\\]|$)/;
```
dustinbyrne, line 3: Ideally these would be replaced by the [exports in `collectContext.ts`](https://github.com/getappmap/appmap-js/blob/9f1f6ca12255e20ffd0397ffa17e352d95d1bfde/packages/cli/src/rpc/explain/collectContext.ts#L11-L13)

dustinbyrne (approved):

packages/cli/tests/unit/rpc/explain/pattern.spec.ts
kgilpin: Sigh.

PR #2023, opened 2024-10-02
feat: Add a single pass token reduction strategy

dustinbyrne:
This PR introduces a `MessageTokenReducerService` to enhance the token reduction strategy for messages which overflow token limits within `AnthropicCompletionService` and `OpenAICompletionService`. The new service:

- Identifies token limits.
- Counts tokens using various strategies.
- Reduces overall token count efficiently in a single pass.

**Changes:**

- Replace `trimLargestUserMessage` with `MessageTokenReducerService`.
- Adjust constructors in `AnthropicCompletionService` and `OpenAICompletionService` to accept `MessageTokenReducerService`.
- Update test cases to mock and test `MessageTokenReducerService`.

The `MessageTokenReducerService` provides a scalable approach to handle tokens, aligning with the API limits dynamically, and improves the context length management by compressing messages more effectively.


packages/navie/src/services/message-token-reducer-service/token-count-strategies/estimate-token-count.ts
```diff
@@ -0,0 +1,10 @@
+import Message from '../../../message';
+import { TokenCountStrategy } from '../interfaces';
+
+export default class EstimateTokenCount implements TokenCountStrategy {
+  constructor(private readonly charactersPerToken = 3) {}
```
dividedmind, line 5: We could use the constant we already have defined somewhere so it doesn't grow inconsistent.

packages/navie/src/services/message-token-reducer-service/token-count-strategies/claude-token-count-error-parser.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export default class ClaudeTokenCountErrorParser implements TokenCountStrategy {
+  countTokens(_messages: Message[], _model: string, apiError?: ApiError): number | undefined {
+    if (!apiError) return undefined;
+
+    const match = /prompt is too long: (\d+) tokens > \d+ maximum/g.exec(apiError.message);
+    if (!match) return undefined;
+
+    const tokenLimit = parseInt(match[1], 10);
+    return tokenLimit;
```
dividedmind, line 12: This will return `NaN` if it can't parse (which won't be caught by `!== undefined` above). Consider changing the test there to truthiness test (`NaN` is falsy), or check here:
```suggestion
    return parseInt(match[1], 10) || undefined;
```

(There are a couple of other instances of the same problem.)

dividedmind (approved):
I can't say I like the Java-esque design very much (most of these classes should just have been functions), but good work! :) Good to merge IMO after fixing those `parseInt` corner cases.

PR #2030, opened 2024-10-03
feat: Support Gemini via Vertex AI

dividedmind:
Note to access the API you have to create a service account at https://console.cloud.google.com/apis/credentials — after creation you'll get a JSON file with credentials. This JSON has to be put into `GOOGLE_WEB_CREDENTIALS` environment variable.

Example configuration:
```
GOOGLE_WEB_CREDENTIALS='{   "type": "service_account",   "project_id": "gen-lang-client-0551265112",   "private_key_id": "1fd731b64e9543d54e2921d7cfbe40c5f6f2845a",   "private_key": "-----BEGIN PRIVATE KEY-----\n...\n-----END PRIVATE KEY-----\n",   "client_email": "divide-s-navie-experiments@gen-lang-client-0551265112.iam.gserviceaccount.com",   "client_id": "112105442388554579010",   "auth_uri": "https://accounts.google.com/o/oauth2/auth",   "token_uri": "https://oauth2.googleapis.com/token",   "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",   "client_x509_cert_url": "https://www.googleapis.com/robot/v1/metadata/x509/divide-s-navie-experiments%40gen-lang-client-0551265112.iam.gserviceaccount.com",   "universe_domain": "googleapis.com" }'
APPMAP_NAVIE_MODEL=gemini-1.5-pro-002
```

### DIFF
--- a/packages/navie/package.json
+++ b/packages/navie/package.json
@@ -41,6 +41,7 @@
   "dependencies": {
     "@langchain/anthropic": "^0.3.1",
     "@langchain/core": "^0.2.27",
+    "@langchain/google-vertexai-web": "^0.1.0",
     "@langchain/openai": "^0.2.7",
     "fast-xml-parser": "^4.4.0",
     "js-yaml": "^4.1.0",
--- a/packages/navie/src/services/google-vertexai-completion-service.ts
+++ b/packages/navie/src/services/google-vertexai-completion-service.ts
@@ -1,0 +1,137 @@
+import { warn } from 'node:console';
+import { isNativeError } from 'node:util/types';
+
+import { ChatVertexAI, type ChatVertexAIInput } from '@langchain/google-vertexai-web';
+import { zodResponseFormat } from 'openai/helpers/zod';
+import { z } from 'zod';
+
+import Trajectory from '../lib/trajectory';
+import Message from '../message';
+import CompletionService, {
+  CompleteOptions,
+  Completion,
+  CompletionRetries,
+  CompletionRetryDelay,
+  convertToMessage,
+  mergeSystemMessages,
+  Usage,
+} from './completion-service';
+
+export default class GoogleVertexAICompletionService implements CompletionService {
+  constructor(
+    public readonly modelName: string,
+    public readonly temperature: number,
+    private trajectory: Trajectory
+  ) {}
+
+  // Construct a model with non-default options. There doesn't seem to be a way to configure
+  // the model parameters at invocation time like with OpenAI.
+  private buildModel(options?: ChatVertexAIInput): ChatVertexAI {
+    return new ChatVertexAI({
+      model: this.modelName,
+      temperature: this.temperature,
+      streaming: true,
+      maxOutputTokens: 8192,
+      ...options,
+    });
+  }
+
+  get miniModelName(): string {
+    const miniModel = process.env.APPMAP_NAVIE_MINI_MODEL;
+    return miniModel ?? 'gemini-1.5-flash-002';
+  }
+
+  // Request a JSON object with a given JSON schema.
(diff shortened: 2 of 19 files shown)


kgilpin (commented):
Thanks! There are some build errors so I flipped this to Draft, temporarily.

packages/navie/package.json
```diff
@@ -41,6 +41,7 @@
   "dependencies": {
     "@langchain/anthropic": "^0.3.1",
     "@langchain/core": "^0.2.27",
+    "@langchain/google-vertexai-web": "^0.1.0",
```
dustinbyrne, line 44: Does this also work with the Gemini API described here?
https://ai.google.dev/gemini-api/docs/api-key

Note that they're referencing a `generativelanguage.googleapis.com` endpoint and not Vertex.

dividedmind: It's a different API: https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/overview#how-gemini-vertex-different-gemini-aistudio

Frankly, it's not clear to me which API we should use, but Langchain seems to only support Vertex.

kgilpin: Have we determined if the usage credits we have from Google will apply?

kgilpin:
Tests still failing.

dividedmind:
I forked the library that was ESM-only and added a CommonJS build (https://github.com/kriasoft/web-auth-library/pull/32) — hopefully this will make it work without fuss.

kgilpin:
Thanks @dividedmind 

Do you have thoughts on the error we were seeing in the benchmark? Would really like to get a clean run though with Gemini.

dividedmind:
> Thanks @dividedmind
> 
> Do you have thoughts on the error we were seeing in the benchmark? Would really like to get a clean run though with Gemini.

You mean the recitation problem? I'm looking at it next.

packages/navie/src/services/google-vertexai-completion-service.ts
```diff
@@ -0,0 +1,137 @@
...
+        break;
+      } catch (cause) {
+        if (attempt < maxAttempts - 1 && tokens.length === 0) {
+          const nextAttempt = CompletionRetryDelay * 2 ** attempt;
+          warn(`Received ${JSON.stringify(cause)}, retrying in ${nextAttempt}ms`);
+          await new Promise<void>((resolve) => {
+            setTimeout(resolve, nextAttempt);
+          });
+          continue;
+        }
```
dustinbyrne, line 119: Note that this doesn't attempt to prune input tokens

packages/navie/src/services/vector-terms-service.ts
```diff
@@ -90,15 +85,20 @@ Terms: test cases +logContext jest`,
...
+    }),
     role: 'assistant',
   },
 ];
 
-const parseText = (text: string): string[] => text.split(/\s+/);
+const schema = z.object({
+  context: z.string(),
+  instructions: z.string(),
+  terms: z.array(z.string()),
```
dustinbyrne, line 100: The description of these properties could move out of the prompt and in to the schema

dustinbyrne (approved):

packages/navie/src/services/google-vertexai-completion-service.ts
dividedmind: No, I figured it's unlikely to be necessary with 2M token input limit...

packages/navie/src/services/vector-terms-service.ts
dividedmind: True, although I tried to change it as little as possible as not to break anything.

PR #2044, opened 2024-10-04
fix: Punctuation in RTL text

dustinbyrne:
This change fixes punctuation rendering on the wrong side of the title. The `UserMessage` story is updated to validate this change.
image

It also removes the file tooltip from the buttons:
image

Resolves https://github.com/getappmap/appmap-js/issues/2036

### DIFF
--- a/packages/components/tests/unit/chat/MarkdownCodeSnippet.spec.js
+++ b/packages/components/tests/unit/chat/MarkdownCodeSnippet.spec.js
@@ -88,6 +88,6 @@ describe('components/MarkdownCodeSnippet.vue', () => {
       provide: { projectDirectories: ['/home/user/my-project'] },
     });
 
-    expect(wrapper.find('[data-cy="context-title"]').text()).toBe('app/models/user.rb');
+    expect(wrapper.find('[data-cy="context-title"]').text()).toContain('app/models/user.rb');
   });
 });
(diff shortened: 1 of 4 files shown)


packages/components/tests/unit/chat/MarkdownCodeSnippet.spec.js
```diff
@@ -88,6 +88,6 @@ describe('components/MarkdownCodeSnippet.vue', () => {
       provide: { projectDirectories: ['/home/user/my-project'] },
     });
 
-    expect(wrapper.find('[data-cy="context-title"]').text()).toBe('app/models/user.rb');
+    expect(wrapper.find('[data-cy="context-title"]').text()).toContain('app/models/user.rb');
```
apotterri, line 91: Won't these toContain checks pass even if the text is not rendering correctly (i.e. has the leading punctation show up on the right)?

dustinbyrne: Yes, though they passed before as well. I've added examples of the behavior to Storybook, so Chromatic will pick up any changes in the future should there be a regression.

dustinbyrne: To be clear, the issue wasn't present in the text of the element, just in how it was rendered.

apotterri (approved):
Works great, thanks for getting this sorted out.

PR #2046, opened 2024-10-04
feat: Ignore more directories of vendored code

kgilpin:
Ignore directories such as:

* .yarn
* venv
* .venv
* vendor
* node_modules

Most of these should not be in Git though, so maybe this isn’t needed?


dustinbyrne (approved):

kgilpin:
@dustinbyrne merge conflicts are resolved.

PR #2050, opened 2024-10-07
fix: add locations to explain requests

### LINKED ISSUE #2049: Pinned files should be refreshed when they change
When I Pin a file and then change that file, my changes are not picked up by Navie. The file is Pinned with the original file content, and doesn't seem to be refreshed when I edit the file.

apotterri:
Add the locations of pinned files to Navie requests, rather than their contents.

Fixes #2049 .

### DIFF
--- a/packages/cli/src/rpc/explain/explain.ts
+++ b/packages/cli/src/rpc/explain/explain.ts
@@ -1,6 +1,3 @@
-import EventEmitter from 'events';
-import { warn } from 'console';
-import { basename } from 'path';
 import {
   AI,
   ConversationThread,
@@ -10,19 +7,22 @@ import {
 } from '@appland/client';
 import { ContextV2, Help, ProjectInfo, UserContext } from '@appland/navie';
 import { ExplainRpc } from '@appland/rpc';
+import { warn } from 'console';
+import EventEmitter from 'events';
+import { basename } from 'path';
 
+import { LRUCache } from 'lru-cache';
+import detectAIEnvVar from '../../cmds/index/aiEnvVar';
+import collectHelp from '../../cmds/navie/help';
+import collectProjectInfos from '../../cmds/navie/projectInfo';
+import configuration, { AppMapDirectory } from '../configuration';
+import { getLLMConfiguration } from '../llmConfiguration';
 import { RpcError, RpcHandler } from '../rpc';
 import collectContext from './collectContext';
-import INavie, { INavieProvider } from './navie/inavie';
-import configuration, { AppMapDirectory } from '../configuration';
-import collectProjectInfos from '../../cmds/navie/projectInfo';
-import collectHelp from '../../cmds/navie/help';
-import { getLLMConfiguration } from '../llmConfiguration';
-import detectAIEnvVar from '../../cmds/index/aiEnvVar';
-import reportFetchError from './navie/report-fetch-error';
-import { LRUCache } from 'lru-cache';
 import { initializeHistory } from './navie/historyHelper';
 import { ThreadAccessError } from './navie/ihistory';
+import INavie, { INavieProvider } from './navie/inavie';
+import reportFetchError from './navie/report-fetch-error';
 import Thread from './navie/thread';
 
 const searchStatusByUserMessageId = new Map<string, ExplainRpc.ExplainStatusResponse>();
@@ -221,11 +221,12 @@ const codeSelections = new LRUCache<string, UserContext.Context>({
     const size = value.reduce((a, v) => {
       let ret = v.type.length;
       if (UserContext.isCodeSelectionItem(v)) {
-        ret += v.location.length + v.content.length;
+        ret += v.content.length;
       } else if (UserContext.isCodeSnippetItem(v)) {
         ret += v.content.length;
-      } else if (UserContext.isFileItem(v)) {
-        ret += v.location!.length;
+      }
+      if (UserContext.hasLocation(v)) {
+        ret += v.location.length;
       }
       return ret;
     }, 0);
--- a/packages/cli/src/rpc/explain/explain.ts
+++ b/packages/cli/src/rpc/explain/explain.ts
@@ -8,7 +8,7 @@ import {
   ProjectDirectory,
   ProjectParameters,
 } from '@appland/client';
-import { ContextV2, Help, ProjectInfo } from '@appland/navie';
+import { ContextV2, Help, ProjectInfo, UserContext } from '@appland/navie';
 import { ExplainRpc } from '@appland/rpc';
 
 import { RpcError, RpcHandler } from '../rpc';
@@ -51,7 +51,7 @@ export class Explain extends EventEmitter {
     public appmapDirectories: AppMapDirectory[],
     public projectDirectories: string[],
     public question: string,
-    public codeSelection: string | undefined,
+    public codeSelection: UserContext.Context | undefined,
     public appmaps: string[] | undefined,
     public status: ExplainRpc.ExplainStatusResponse,
     public codeEditor?: string,
@@ -104,7 +104,12 @@ export class Explain extends EventEmitter {
       }
     }
 
-    await navie.ask(this.status.threadId, this.question, this.codeSelection, this.prompt);
+    await navie.ask(
+      this.status.threadId,
+      this.question,
+      this.codeSelection,
+      this.prompt
+    );
   }
 
   async searchContext(data: ContextV2.ContextRequest): Promise<ContextV2.ContextResponse> {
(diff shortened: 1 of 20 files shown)


packages/navie/src/navie.ts
```diff
@@ -46,6 +46,7 @@ export interface ClientRequest {
   question: string;
   codeSelection?: string;
```
kgilpin, line 47: Historically, this `codeSelection` was the code that the user had selected in their IDE when they opened Navie. So, it was a string. And even then, as a string, it was deficient, because it didn't include the location from which the selected code was obtained.

So, I think there is an opportunity to redefine this codeSelection argument as a `string | UserContext` type, where `UserContext` can support all the different strategies that the user might employ to send context to Navie. 

This would include, but not necessarily be limited to:

- Code selections. A code selection is a range (start..end) from a file, plus the specific code from that range.
- Files. A file is just a file location. Navie should use the context API to retrieve the contents of that file. This operation is already supported in the context lookup facility. 

```
    type ContextV1.ContextRequest = {
        tokenCount: number;
        vectorTerms: string[]; <-- Leave this blank to fetch locations instead.
    };
    type ContextFilters = {
        recent?: boolean;
        locations?: string[]; <-- Can be populated to fetch the full contents of specified files.
        itemTypes?: ContextItemType[];
        labels?: ContextLabel[];
        exclude?: string[];
        include?: string[];
    };
    type ContextRequest = ContextV1.ContextRequest & ContextFilters;
```

apotterri: > Lookup by location is already supported like this:

Yes, I saw this, and a ContextRequest is ultimately the way file contents get retrieved. As far as I can tell, though, the existing interface for an Explain request doesn't allow the client to pass a ContextRequest. Instead, the pieces of the ContextRequest are passed as function parameters, and then get assembled into filters here: https://github.com/getappmap/appmap-js/blob/main/packages/navie/src/services/context-service.ts#L39 .

kgilpin: You can check out the generate agent to see how it fetches full file contents

apotterri: I guess I'm confused by this comment. The Explain agent uses the `searchContext` method of `ContextService`, which has this signature:
```
  async searchContext(
    options: AgentOptions,
    tokensAvailable: () => number,
    additionalTerms: string[] = [],
    locations?: string[]
  ): Promise<void> 
```
This sets the passed `locations` on a `ContextFilter`, which it then passes to `this.lookupContextService.lookupContext`.

The Generate agent uses `ContextService.locationContext`, which has this signature:
```
  async locationContext(fileNames: string[]): Promise<void>
```
`locationContext` sets `fileNames` as the `locations` on a `ContextFilter`, then it too calls `this.lookupContextService.lookupContext`.

Aren't these doing the same thing to fetch the full file contents? I understood your suggestion about creating `UserContextItem` to encapsulate all the different kinds of context a user can provide, and then passing an array of those, rather than having `codeSelection` and `locations` be separate arguments. Can't this array then be passed to `searchContext` as part of `AgentOptions`?

apotterri:
Changing `ClientRequest` seems like it necessitates changes on both sides of the RPC boundary, i.e. for both clients that make RPC calls, as well as for the server-side functions that handle the calls.

[89c769e](https://github.com/getappmap/appmap-js/pull/2050/commits/89c769eafae0a147d3cfa8a3356af52e70ca35b6) attempts to address the interface changes necessary.

packages/cli/src/rpc/explain/explain.ts
```diff
@@ -208,15 +213,35 @@ export class Explain extends EventEmitter {
...
+      let ret = v.type.length;
+      switch (v.type) {
+        case ContextV2.ContextItemType.CodeSelection:
+          ret += v.location!.length + v.content!.length;
+          break;
+        case ContextV2.ContextItemType.CodeSnippet:
+          ret += v.content!.length;
+          break;
+        case ContextV2.ContextItemType.File:
+          ret += v.location!.length;
```
dustinbyrne, line 232: We shouldn't need to use the non-null assertion operator, it generally indicates something is mis-typed.

Consider something like:
```ts
interface CodeSelectionContextItem {
    type: 'code-selection';
    location: string;
    content: string;
}

interface CodeSnippetContextItem {
    type: 'code-snippet';
    content: string;
    location?: never;
}

// This doesn't include the full list
type UserContextItem = CodeSelectionContextItem | CodeSnippetContextItem;
```

kgilpin (commented):
This seems directionally good.

packages/navie/src/user-context.ts
```diff
@@ -0,0 +1,69 @@
+export namespace UserContext {
+  export type CodeSelectionItem = {
+    type: 'code-selection';
+    location: string;
```
kgilpin, line 4: Since we know this is based on a file, that file can be a candidate for code generation output.

packages/navie/src/user-context.ts
```diff
@@ -0,0 +1,69 @@
...
+    content: string;
+  };
+
+  export type CodeSnippetItem = {
+    type: 'code-snippet';
```
kgilpin, line 9: These do not have an associated location; which actually might not be correct. If the user is Pinning a generated snippet, might that snippet have an associated location?

packages/navie/src/user-context.ts
```diff
@@ -0,0 +1,69 @@
...
+    const pinnedItems = items.filter(
+      (i) => UserContext.isCodeSelectionItem(i) || UserContext.isFileItem(i)
+    );
+    const snippets = items.filter((i) => UserContext.isCodeSnippetItem(i));
+    if (pinnedItems.length > 0 || snippets.length > 0) {
+      renderedItems.push('<user-provided-context>');
+      if (pinnedItems.length > 0) {
+        renderedItems.push(
+          `<!-- Pinned items are snippets that I have saved from earlier in the conversation -->`,
+          ...pinnedItems.map((i) => `<pinned-context>\n${renderItem(i)}\n</pinned-context>`)
```
kgilpin, line 56: And as discussed above, these may also have a location.

packages/navie/src/agent.ts
```diff
@@ -32,6 +33,11 @@ export class AgentOptions {
   buildContextFilters(): ContextV2.ContextFilters {
     const filters: ContextV2.ContextFilters = {};
     if (this.contextLabels) filters.labels = this.contextLabels;
+    if (this.codeSelection && typeof this.codeSelection !== 'string') {
+      filters.locations = this.codeSelection
+        .filter(UserContext.isFileItem)
```
kgilpin, line 38: May want to change this to a filter like `hasLocation`

apotterri:
@kgilpin I believe the changes in the `!fixup` commits address the concerns you raised.

One outstanding issue: how should pinned files be handled when Navie is remote? https://github.com/getappmap/appmap-js/blob/explain-locations_20241007/packages/cli/src/rpc/explain/navie/navie-remote.ts#L190

packages/components/src/components/chat/MarkdownCodeSnippet.vue
```diff
@@ -96,6 +96,11 @@ export default Vue.extend({
     },
     decodedLocation(): string | undefined {
       // The location may be URI encoded to avoid issues with special characters.
+      console.log(
+        `decodedLocation, this.location: ${
+          this.location
+        } decodeURIComponent(this.location): ${decodeURIComponent(this.location)}`
+      );
```
dustinbyrne, line 103: This looks like it can be dropped

packages/navie/src/user-context.ts
```diff
@@ -0,0 +1,75 @@
...
+    const renderItem = (i: UserContext.ContextItem) => {
+      if (UserContext.isCodeSelectionItem(i) && includeContent) {
+        return `<code-selection>\n${i.content}\n</code-selection>`;
+      } else if (UserContext.isCodeSnippetItem(i)) {
+        const content = includeContent ? `\n<content>${i.content}</content>\n'` : '';
+        return `<pinned-snippet>
+  <location>${i.location}</location>${content}
+</pinned-snippet>`;
+      } else if (UserContext.isFileItem(i)) {
+        return `<pinned-file><location>${i.location}</location></pinned-file>`;
```
dustinbyrne, line 48: Pinned files never include content?

packages/components/src/components/chat/MarkdownCodeSnippet.vue
apotterri: Thanks, fixed.

packages/navie/src/user-context.ts
apotterri: That's right, pinned files never include content. #2049 is about stale file contents. These changes avoid that by having context selection always read the file.

apotterri: One of the significant changes here is this: https://github.com/getappmap/appmap-js/blob/explain-locations_20241007/packages/navie/src/agent.ts#L36 . When a `ContextItem` has a `location`, it's included in the `ContextFilters`. As I understand it, this is the same mechanism that `@generate` uses to read the full contents of files.

You can see this in action in Navie's log:
```
[context-service] Searching for context
Tokens (prompt/compl/total): 657/26/683, cost: $0.00
Vector terms response:
Context: Exception handling in Java
Instructions: Explain how exception handling is managed
---
Terms: exception handling +Java try catch finally
Explain received context request: search
Collecting context with parameters: sourceDirectories: /Users/ajp/src/apotterri/ejb-example, charLimit: 16503, appmapDirectories: /Users/ajp/src/apotterri/ejb-example, keywords: exception, handling, java, try, catch, finally, locations: /Users/ajp/src/apotterri/ejb-example/src/main/java/org/superbiz/remote/App.java
Parsed locations: /Users/ajp/src/apotterri/ejb-example/src/main/java/org/superbiz/remote/App.java
[location-context] Candidate locations: /Users/ajp/src/apotterri/ejb-example/src/main/java/org/superbiz/remote/App.java
[location-context] Extracting snippet for location: /Users/ajp/src/apotterri/ejb-example/src/main/java/org/superbiz/remote/App.java (2127 bytes)
```

What you see under Code Snippets is in addition to the contents of files. This redundancy is what this issue is about: https://github.com/getappmap/appmap-js/issues/1984 .

dustinbyrne (approved):

PR #2055, opened 2024-10-07
fix: Use 4 historical chat messages

dustinbyrne:
Resolves https://github.com/getappmap/appmap-js/issues/2054


kgilpin (commented):
Does the LLM see the messages chronologically?

Can you or Navie think of any potential problems this could introduce?

Did you see any difference when you tried it with a long thread? You could simulate a long user interaction in which Navie may get confused.

dustinbyrne:
There are existing tests that validate the content and ordering of historical messages returned by this method.

We know that this will increase token usage, however I tried this with Copilot and didn't experience any issues.

kgilpin (approved):

PR #2059, opened 2024-10-09
feat: Navie + Copilot frontend support

dustinbyrne:
Work in progress

image

image


kgilpin:
Some suggestions.

## Provider order

Order the backend providers like this:

1. Copilot
2. AppMap hosted provider
3. BYOK
4. BYOM

## Rename "Navie backend"

All providers are technically a "Navie backend". So, rename the "Navie backend" option to "AppMap hosted provider", or something like that.

## Other API keys

We also support other API keys, such as Anthropic. So reword "OpenAI API Key" to reflect this.

dustinbyrne:
Left to do here is to update the link to the Copilot docs to something more appropriate

PR #2062, opened 2024-10-12
Expose RPC server on localhost instead of anyhost

dividedmind:
This change sets the RPC server to listen on localhost specifically, preventing network access which could potentially be problematic (and which can trigger firewall rules or prompts, specifically on Windows).

I've also fixed some lint problems while I was at it.


kgilpin (approved):

dividedmind:
Just a note that we might need to either add dual-binding (for IPv4 and v6) or bind to 127.0.0.1 — discussion at https://github.com/getappmap/vscode-appland/pull/1037#discussion_r1801793691

PR #2068, opened 2024-10-15
feat: Report AppMap usage stats

kgilpin:
Report run stats as the fingerprint watch command runs.

1. **Refactoring `emitUsage` to `writeUsage`, `collectUsageData`, and `sendUsageData`:**
   - **writeUsage:** 
     Writes usage statistics for AppMap events and files to a JSON file.
   - **collectUsageData:** 
     Collects usage data including metadata, number of events, and number of AppMap files.
   - **sendUsageData:** 
     Sends the usage data to a remote server; if it fails, it logs the data locally.

2. **Updated various files to use the new functions:**
   - `packages/cli/src/cmds/archive/index.ts`
   - `packages/cli/src/fingerprint/fingerprintDirectoryCommand.ts`
   - `packages/cli/src/fingerprint/fingerprintWatchCommand.ts`

### DIFF
--- a/packages/cli/src/cmds/archive/index.ts
+++ b/packages/cli/src/cmds/archive/index.ts
@@ -6,7 +6,7 @@ import { IndexResult } from '../index/IndexResult';
 import { basename, dirname, join } from 'path';
 import { readFile } from 'fs/promises';
 import { Metadata } from '@appland/models';
-import emitUsage from '../../lib/emitUsage';
+import writeUsage, { collectUsageData } from '../../lib/emitUsage';
 
 export async function index(
   workerPool: WorkerPool,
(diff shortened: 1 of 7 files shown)


packages/cli/src/cmds/archive/index.ts
```diff
@@ -39,7 +39,14 @@ export async function index(
   );
 
   if (sampleMetadata) {
-    await emitUsage(appMapDir, totalEvents, result.numProcessed, sampleMetadata);
+    const outputDir = join('.appmap', 'run-stats');
```
dustinbyrne, line 42: Is this meant to go in the cwd?
```suggestion
    const outputDir = join(homedir(), '.appmap', 'run-stats');
```

dustinbyrne (approved):

packages/cli/src/cmds/archive/index.ts
kgilpin: It actually is meant to go in the `cwd`, because this command is designed for use in CI. The data file gets written to `.appmap/run-stats` where it gets picked up by the api services backend.

You can see the old implementation was:

```
const runStatsDirectory = join('.appmap', 'run-stats');
```

So that directory is now an input parameter to `writeUsage`.

kgilpin:
I messed this up, there is actually a test failure in `test_the_rest`. I'll fix in a separate PR. Meanwhile the build on main will prevent this from releasing.

PR #2078, opened 2024-10-18
fix: Normalize text before keyword splitting

### LINKED ISSUE #2067: Context search misses matches due to case sensitivity & camel case analysis
The following questions don't yield the same context:

Does not produce good matches:

```
class diagram of searchrpc and explainrpc
```

Does produce good matches:

```
class diagram of SearchRPC and ExplainRPC
```

dustinbyrne:
This guarantees the full text search remains case insensitive. E.g., `HTTPServer` would split into `http server httpserver` Where `httpserver` would split into `httpserver`.

This does affect ranking in that the case sensitive match would have produced a higher score due to the additional matches.

Resolves #2067

### DIFF
--- a/packages/cli/src/fulltext/FileIndex.ts
+++ b/packages/cli/src/fulltext/FileIndex.ts
@@ -52,7 +52,7 @@ export class FileIndex {
   search(keywords: string[], limit = 10): FileIndexMatch[] {
     const query = `SELECT directory, file_name, (rank * -1) score FROM files WHERE files MATCH ? ORDER BY rank LIMIT ?`;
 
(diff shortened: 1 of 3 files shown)


dustinbyrne:
As mentioned, this does affect how well some search terms will perform. The alternative would be to attempt to split lower cased forms, likely using a word list.

dividedmind:
I don't think this is a good idea — this is just losing information. There is more information in "HTTPServer" than in "httpserver" and this change is throwing it away.

kgilpin:
It’s a difficult case, but I don’t think lower casing everything is better.

dustinbyrne:
Closed in favor of #2081

PR #2079, opened 2024-10-21
feat: Upgrade search

kgilpin:
## TODO

- [x] Ensure that include/exclude filters are being applied properly.

Full-text search is refactored out into its own package, `@appland/search`. In addition, the search functionality is upgraded and improved in a couple of different ways. Aside from these improvements, the code is generally moved over from `@appland/appmap`.

### Inclusion of all document words

All document words are now included in the index. This is an upgrade from current behavior with source code files, in which only the file path and recognized symbols are indexed.

Symbols are detected in the same ways as with the current behavior. Then, the document is split into words. Each word that is already recognized as a symbol is removed from the word list, so that it's not double-counted.

### Weighted terms

Each source file index entry includes three separate columns, which are essentially:

* File path: `weight=3`
* Symbols: `weight=2`
* Words: `weight=1`

SQLite query syntax makes it possible to perform a weighted search over the file path, symbols, and words.

### Boost factors

Each document can also be associated with a row in the boost factors table. This feature makes it possible to boost search results based on external factors. The intended usage for this feature is to boost the source files and snippets that appear in AppMap trace data files. Boost can also be applied based on other external factors, such as presence of a file in a relevant log entry, backtrace, or APM data file (to name three examples).

## Search exports

`@appland/search` exports functions and constants that can be used externally to perform a search. Most optional behavior is controlled by passing in a function that matches a particular type signature. Concrete implementations of these functions are provided, but not assigned by default.


dustinbyrne:
Looking good, and I'm glad to see search getting its own package :+1:. I've had similar thoughts about pulling out the local/remote Navie clients.

kgilpin:
Cool; I was wondering how you’d feel about that. The CLI package is so crowded.

kgilpin:
I've tested this in the benchmark and it adds about 1% to the solve rate. So, good to merge.

PR #2083, opened 2024-10-22
fix: Use git attributes to identify binary files

### LINKED ISSUE #2042: Use git attributes to identify binary files
Git metadata can indicate which files are binary. Use this information as well as file extension.

dustinbyrne:
Fixes #2042

Add support for identifying binary files using git attributes in addition to file extensions.

* Add `getGitAttributes` function to read and parse the `.gitattributes` file in `packages/cli/src/fulltext/FileIndex.ts`.
* Update `isBinaryFile` function to check git attributes information in addition to file extensions in `packages/cli/src/fulltext/FileIndex.ts`.
* Update `filterFiles` function to use the `getGitAttributes` function to determine if a file is binary based on git attributes in `packages/cli/src/fulltext/FileIndex.ts`.
* Add tests to verify that the context search/lookup/collector correctly identifies binary files using git attributes information in `packages/cli/tests/unit/fulltext/FileIndex.spec.ts`.

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/getappmap/appmap-js/issues/2042?shareId=7a68523a-affa-44f8-80f9-d10a1d5a1e10).

### DIFF
--- a/packages/cli/src/fulltext/FileIndex.ts
+++ b/packages/cli/src/fulltext/FileIndex.ts
@@ -259,11 +259,9 @@ const DATA_FILE_EXTENSIONS: string[] = [
   'xml',
 ].map((ext) => '.' + ext);
 
-const isBinaryFile = (fileName: string, gitAttributes?: ReturnType<typeof getGitAttributes>) => {
-  if (BINARY_FILE_EXTENSIONS.some((ext) => fileName.endsWith(ext))) return true;
-  if (gitAttributes?.some((attr) => attr.binary && minimatch(fileName, attr.pattern))) return true;
-  return false;
-};
+const isBinaryFile = (fileName: string, gitAttributes?: ReturnType<typeof getGitAttributes>) =>
+  BINARY_FILE_EXTENSIONS.some((ext) => fileName.endsWith(ext)) ||
+  gitAttributes?.some((attr) => attr.binary && minimatch(fileName, attr.pattern, { dot: true }));
 
 const isDataFile = (fileName: string) => {
   return DATA_FILE_EXTENSIONS.some((ext) => fileName.endsWith(ext));
--- a/packages/cli/src/fulltext/FileIndex.ts
+++ b/packages/cli/src/fulltext/FileIndex.ts
@@ -1,10 +1,12 @@
 import { stat } from 'node:fs/promises';
 import path, { join } from 'node:path';
+import { readFileSync } from 'node:fs';
 
 import sqlite3 from 'better-sqlite3';
 import assert from 'assert';
 import makeDebug from 'debug';
 import { existsSync } from 'fs';
+import minimatch from 'minimatch';
 
 import listProjectFiles from './listProjectFiles';
 import queryKeywords from './queryKeywords';
@@ -80,11 +82,14 @@ export class FileIndex {
             ? await listGitProjectFiles(directory)
             : await listProjectFiles(directory);
 
+        const gitAttributes = getGitAttributes(directory);
+
         const filteredFileNames = await filterFiles(
           directory,
           fileNames,
           excludePatterns,
-          includePatterns
+          includePatterns,
+          gitAttributes
         );
 
         const options = {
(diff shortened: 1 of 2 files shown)


packages/cli/src/fulltext/FileIndex.ts
```diff
@@ -298,3 +304,20 @@ export async function filterFiles(
...
+  const lines = gitAttributesContent.split('\n').filter(Boolean);
+
+  return lines.map((line) => {
+    const [pattern, ...attributes] = line.split(/\s+/);
+    return {
+      pattern,
+      binary: attributes.includes('binary'),
+    };
+  });
+}
```
dividedmind, line 323: Note this is much simpler than actual [gitattributes(5)](https://git-scm.com/docs/gitattributes):

> Each line in gitattributes file is of form:
> 
> pattern attr1 attr2 ...
> 
> That is, a pattern followed by an attributes list, separated by whitespaces. Leading and trailing whitespaces are ignored. Lines that begin with # are ignored. Patterns that begin with a double quote are quoted in C style. When the pattern matches the path in question, the attributes listed on the line are given to the path.

> When deciding what attributes are assigned to a path, Git consults $GIT_DIR/info/attributes file (which has the highest precedence), .gitattributes file in the same directory as the path in question, and its parent directories up to the toplevel of the work tree (the further the directory that contains .gitattributes is from the path in question, the lower its precedence). Finally global and system-wide files are considered (they have the lowest precedence).

This might be fine for our purposes, but the simplification warrants a comment at least — and I think skipping comments is easy enough to just do it.

packages/cli/src/fulltext/FileIndex.ts
```diff
@@ -298,3 +304,20 @@ export async function filterFiles(
   }
   return result;
 }
+
+// Reads git attribute patterns from a .gitattributes file.
+export function getGitAttributes(directory: string) {
+  const gitAttributesPath = join(directory, '.gitattributes');
+  if (!existsSync(gitAttributesPath)) return [];
+
+  const gitAttributesContent = readFileSync(gitAttributesPath, 'utf-8');
```
dividedmind, line 313: [From node.js doc](https://nodejs.org/api/fs.html#fsexistspath-callback):

> Using fs.exists() to check for the existence of a file before calling fs.open(), fs.readFile(), or fs.writeFile() is not recommended. Doing so introduces a race condition, since other processes may change the file's state between the two calls. Instead, user code should open/read/write the file directly and handle the error raised if the file does not exist.

(Essentially, exists returning true doesn't guarantee that read will succeed, so you have to handle the error anyway, so might as well skip the exists.)

packages/cli/src/fulltext/FileIndex.ts
```diff
@@ -298,3 +304,20 @@ export async function filterFiles(
...
+  const lines = gitAttributesContent.split('\n').filter(Boolean);
+
+  return lines.map((line) => {
+    const [pattern, ...attributes] = line.split(/\s+/);
+    return {
+      pattern,
+      binary: attributes.includes('binary'),
+    };
```
dividedmind, line 321: If it's only pulling binary like that, why even bother returning other patterns? Perhaps it should just return a list of binary patterns instead. This will also simplify upstream code — you can just add those to excluded patterns.

packages/cli/src/fulltext/FileIndex.ts
```diff
@@ -259,11 +259,9 @@ const DATA_FILE_EXTENSIONS: string[] = [
...
 ].map((ext) => '.' + ext);
 
-const isBinaryFile = (fileName: string, gitAttributes?: ReturnType<typeof getGitAttributes>) => {
-  if (BINARY_FILE_EXTENSIONS.some((ext) => fileName.endsWith(ext))) return true;
-  if (gitAttributes?.some((attr) => attr.binary && minimatch(fileName, attr.pattern))) return true;
-  return false;
-};
+const isBinaryFile = (fileName: string, gitAttributes?: ReturnType<typeof getGitAttributes>) =>
+  BINARY_FILE_EXTENSIONS.some((ext) => fileName.endsWith(ext)) ||
+  gitAttributes?.some((attr) => attr.binary && minimatch(fileName, attr.pattern, { dot: true }));
```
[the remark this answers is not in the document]
kgilpin: This looks like a good fix to have. Is it in any other merged PRs?

kgilpin:
@dustinbyrne @dividedmind Elsewhere, we implemented binary file detection by inspecting the file contents. How do we want to reconcile / merge these approaches?

Issue #2071, opened 2024-10-16
@diagram when the diagram has unresolved errors, helpful information should be displayed


kgilpin:
In place of the empty Diagram box:

<img width="902" alt="Screen Shot 2024-10-16 at 4 35 23 AM" src="https://github.com/user-attachments/assets/bf28e0b5-f3c0-4ebb-b9bc-1b62b966f06e">

We should display information such as:

1. The diagram markdown that's available
2. Information about the error
3. How to edit the diagram manually to fix the error?
4. How to open in Mermaid Live to fix it?
5. Fix it inline? 
6. A follow up prompt that might help (e.g. "simplify object and property names")

dividedmind:
Note: #2084 makes sure the code is displayed if the backend cannot fix the diagram. While it's not a complete fix (we could make more UX improvements), this should fix the apparent regression.

PR #2084, opened 2024-10-23
More robust Mermaid diagram fixing

dividedmind:
This improves mermaid diagram fixing in the backend by:
- simplifying the code;
- implementing a hack to make dompurify work in the backend; this ensures broken diagrams are detected correctly (previously, when diagram validation attempted to call dompurify, the error would cause the validation to end prematurely and make the code ignore the spurious error and assume the diagram was valid even though an error could have been detected later) — while this does not solve #2071 entirely it at least makes the user see the broken code as text,
- show a message to the user when repairing a diagram so it doesn't look stuck.


packages/navie/package.json
```diff
@@ -45,6 +45,7 @@
     "@langchain/openai": "^0.2.7",
     "fast-xml-parser": "^4.4.0",
     "js-yaml": "^4.1.0",
+    "jsdom": "^16.6.0",
     "langchain": "^0.2.16",
     "mermaid": "^9",
```
apotterri, line 50: As long as you're here, can you take a look at upgrading to mermaid 10? I believe that would allow vscode-appland to link to a local appmap-js repo.

dividedmind: I actually did try, but mermaid 10 is ESM-only.

dustinbyrne (approved):

PR #2087, opened 2024-10-23
Visual adjustments to signup form

Daniel-Warner-X:
This is part of the [Get New Users Into Chat](https://github.com/getappmap/board/issues/335) epic. Along with updates to the UI of adjacent components, this contributes to conveying a sense of organization and stability to new users. It also aims to drive more email activations but being clearer about how lightweight that process is.

Updates the button label on email signup, the visual design of the 'your data is your data' message, and the responsive behavior of the form at large sizes.

New View:

image

image

Existing View:

image

image

### DIFF
--- a/packages/components/src/components/SignIn.vue
+++ b/packages/components/src/components/SignIn.vue
@@ -2,7 +2,7 @@
   <div class="signin-sidebar">
     <AppMapLogo width="120" />
     <div v-if="!submitted" class="content-wrap">
-      <div class="sidebar-title" data-cy="title">Activate AppMap</div>
+      <div data-cy="title">Activate AppMap</div>
 
       <div class="signin-buttons">
         <p v-if="error" class="error">{{ error }}</p>
@@ -20,7 +20,7 @@
           data-cy="email-activation-button"
           @click.prevent="activateWithEmail"
         >
-          Activate with<span><EmailIcon class="small-logo" /> Email</span>
+          Get activation code by <span><EmailIcon class="small-logo" /> Email</span>
         </a>
 
         <div class="or-divider-container">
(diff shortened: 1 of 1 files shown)


packages/components/src/components/SignIn.vue
```diff
@@ -384,7 +388,6 @@ export default {
   .your-data {
     display: flex;
     flex-direction: row;
-    align-items: center;
```
dustinbyrne, line 387: When the "your data is your data" messaging collapses into rows, the shield is no longer aligned with this change
image

packages/components/src/components/SignIn.vue
```diff
@@ -396,6 +399,10 @@ export default {
       max-width: 33px;
       height: 42px;
     }
+
+    .your-data-text {
+      font-size: 0.85rem;
```
dustinbyrne, line 404: I've noticed that `rem` units tend to be larger once in VS Code. You may want to consider using `em` here. Here's what it currently looks like in VS Code:
image

packages/components/src/components/SignIn.vue
dustinbyrne: Looks like there's a `@media (max-width: 365px)` query responsible for this somewhere.

packages/components/src/components/SignIn.vue
```diff
@@ -250,6 +250,10 @@ export default {
   padding: 1.5rem;
```
dustinbyrne, line 250: This looks unrelated to this particular change, but I'm noticing that the `height: 100%` on `.signin-sidebar` will cause text to overflow the container if the view is too narrow.
image

dustinbyrne:
This test will need to be updated
https://github.com/getappmap/appmap-js/blob/d764365be5f0ccf3554d654fdcbc563d88afe3c2/packages/components/tests/e2e/specs/sidebarSignIn.spec.js#L21-L23

`Activate with` just needs to change to `Get activation code by`

kgilpin:
Closing this due to lack of activity.

PR #2092, opened 2024-10-27
fix: Strip options from chat history given to the LLM

dividedmind:
Giving options to the LLM tends to confuse it; for example seeing `@generate /format=xml` often leads to it suggesting things like `@generate a plan in XML format`. Since the options are a side channel meant for Navie, not the LLM, it should be safe to strip them.

### DIFF
--- a/packages/navie/src/navie.ts
+++ b/packages/navie/src/navie.ts
@@ -238,10 +238,26 @@ export default function navie(
     async *execute(): AsyncIterable<string> {
       assert(command, 'Command not specified');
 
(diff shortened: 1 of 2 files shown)


kgilpin (approved):
Good catch! Should we also strip the command name (eg `@generate`)?

dividedmind:
> Good catch! Should we also strip the command name (eg `@generate`)?

I was thinking about it, but I think it's fine to leave it in — IMO it's important context and it should be quite understandable on its own.

PR #2097, opened 2024-10-28
fix: Adjust the first line of a patch when it has 0 indent

kgilpin:
Fixes https://github.com/getappmap/appmap-js/issues/2095


dividedmind:
I don't think this actually works... let me double check. I had a simpler solution in mind, if this doesn't work then I'll try mine, ok?

kgilpin:
Yeah, it doesn't. Try your way.

dividedmind:
> Yeah, it doesn't. Try your way.

I will. I pushed a commit that makes the tests more robust so it's apparent that it doesn't work.

dividedmind:
Another approach at #2100

PR #2120, opened 2024-11-22
feat: Update appmap file search

kgilpin:
Somehow this looks like a huge set of changes, but it isn't really. Only one commit really matters:

https://github.com/getappmap/appmap-js/pull/2120/commits/ac0004717147a095f1fa609c2aa341dec6e6c7bc

What's happening here is that the AppMap search [based on lunr](https://github.com/getappmap/appmap-js/pull/2120/commits/ac0004717147a095f1fa609c2aa341dec6e6c7bc#diff-d98dba68a2edf4f99b569e8401c3f2abb0af6152d984108f03acf79566a8a85b) is migrated to use [sqlite and the new search package](https://github.com/getappmap/appmap-js/pull/2120/commits/ac0004717147a095f1fa609c2aa341dec6e6c7bc#diff-b2de4d44acb99b50a421ea070f8bcee3eab528a58abdf0aab7af1b3c61a7fcb9).

A fair amount of code is moved around. Why? Because follow-on work will:

a) continue to replace Lunr search with sqlite.
b) combine the AppMap event searches and code snippet searches into one integrated query, so that both runtime event content and code snippet content are ranked together, resulting in better relevance of the search context.


packages/cli/src/fingerprint/canonicalize/packageDependencies.js
```diff
@@ -9,7 +9,7 @@ function packageOf(codeObject) {
   }
 
   const ancestors = [codeObject, ...codeObject.ancestors()];
-  let packageObject = ancestors.find((a) => a.type === CodeObject.PACKAGE);
+  let packageObject = ancestors.find((a) => a.type === CodeObjectType.PACKAGE);
```
kgilpin, line 12: This is a bug that resulted in the package object being missed or mis-identified, resulting in the package dependencies list in the index file being empty.

packages/search/src/build-snippet-index.ts
```diff
@@ -16,14 +17,16 @@ type Context = {
 };
 
 async function indexFile(context: Context, file: File) {
-  const fileContent = await context.contentReader(file.filePath);
+  const filePath = isAbsolute(file.filePath) ? file.filePath : join(file.directory, file.filePath);
```
kgilpin, line 20: In most cases, the  file path was already absolute, so this bug didn't manifest. But in a test case that passed a relative path for the file path, this bug was revealed. It should do the right thing in either case.

packages/search/src/snippet-index.ts
```diff
@@ -47,22 +42,63 @@ ORDER BY
   score DESC
 LIMIT ?`;
 
+export enum SnippetType {
+  FileChunk = 'file-chunk',
+}
+
+export type SnippetId = {
```
kgilpin, line 49: Adding structure to the SnippetId makes it easier for clients to interpret the different types of search results that they may get.

PR #2123, opened 2024-11-23
feat: Search snippets and appmap events using sqlite

kgilpin:
In the old/existing algorithm, the search context collects events from AppMap data, then looks up the source code for the code objects represented in that event list. Then it supplements these search results with some text snippets. The ratio of event-based results to text-search-based results is 3 to 1:

https://github.com/getappmap/appmap-js/pull/2123/files#diff-b247cb5820889b6fdee4a942ce59747b9dd5285ee212a7b89a840470b41c600dL129

In this new implementation, events and code snippets are combined together in the same index, and search and ranked together. 

This prevents the search results from being dominated by AppMap events, in the case that the AppMap data is not truly very relevant to the search.


packages/cli/src/rpc/explain/collect-context.ts
```diff
@@ -0,0 +1,136 @@
+import { ContextV2 } from '@appland/navie';
```
kgilpin, line 1: This file is essentially `buildContext`, but with a more (IMO) clear name, since the job of the function is not just to build the context, but to collect it from the available assets.

packages/cli/src/rpc/navie/metadata.ts
```diff
@@ -39,18 +39,6 @@ export function navieMetadataV1(): RpcHandler<
         {
           name: '@test',
           description: 'Write tests for your code.',
-          referenceUrl: 'https://appmap.io/docs/navie-reference/navie-commands.html#test',
```
kgilpin, line 42: Restore these changes before merge

packages/cli/tests/unit/rpc/explain/review.spec.ts
```diff
@@ -1,83 +0,0 @@
-import { EventEmitter } from 'stream';
```
kgilpin, line 1: Be sure to restore this.

PR #2157, opened 2024-12-12
feat: Navie welcome message UI

### LINKED ISSUE #2148: Suppress question suggestions when there is a code selection
When Navie chat opens with a code selection, suppress / short cut the question suggestions. 

We don't need to suggest project-related questions when we can see that the user is clearly about to ask about the code selection they have made.

### LINKED ISSUE #2155: Support a structured response for the Welcome Message and render suggested questions as links or buttons
Currently, the welcome message handler returns a string. The string is a mashup of the user's named activity along with suggested questions that she might ask. The format of the string is Markdown.

Formatting the welcome message in this manner makes it awkward for the UI to render the suggested questions as links or buttons. 

Provide a separate method for inferring the activity name and suggesting questions:

```
  export namespace V1 {
    export namespace Welcome {
      export const Method = 'v1.navie.welcome';
      export type Params = {
        codeSelection?: string;
      };
      export type Response = {
        activity: string;
        suggestions: string[];
      };
    }
  }
```

This new message offers the following features:

* The `codeSelection` is available to the welcome message generator
* The activity name and suggested questions are separate fields
* Input place holder and command list are available immediately. 

The frontend can test whether the RPC service supports this new welcome method, and use it if it exists. Otherwise, it can fall back to the existing V1 behavior.

kgilpin:
Introduces a new version of the welcome message UI that provides a more structured and context-aware experience. The key changes include:

- Splitting the welcome message into static and dynamic parts to improve initial load time
- Adding support for displaying the current activity name and related suggestions
- Introducing new RPC methods for metadata (v2) and welcome message content
- Preserving backward compatibility with v1 welcome message format

The new UI shows a standard greeting header followed by either:
- A custom welcome message in markdown format, or
- The current activity name with relevant suggestions presented as a bulleted list

Resolves #2155
Resolves #2148

### DIFF
--- a/packages/components/src/components/chat/WelcomeMessageV2.vue
+++ b/packages/components/src/components/chat/WelcomeMessageV2.vue
@@ -1,20 +1,27 @@
 <template>
-  <div class="welcome-message" data-cy="welcome-message">
+  <div class="welcome-message" data-cy="welcome-message" data-version="2">
     <div class="welcome-message-static">
       <h3>Hi, I'm Navie!</h3>
     </div>
     <div class="welcome-message-dynamic" v-if="welcomeLoaded">
-      <div v-if="welcomeMessage" v-html="welcomeMessage" class="welcome-message-text" />
+      <div v-if="welcomeMessage" v-html="renderedWelcomeMessage" class="welcome-message-text" />
       <div v-else-if="activityName" class="welcome-message-structured">
-        <p>I see you're working on {{ activityName }}. Here are some things you can try:</p>
+        <p>I see you're working on {{ activityName }}.</p>
+        <p>Here are some suggestions for what you can do next:</p>
         <ul>
           <li v-for="suggestion in trimmedSuggestions" :key="suggestion">
-            {{ suggestion }}
+            <a
+              href="#"
+              @click.prevent="onSuggestionClick(suggestion)"
+              v-html="renderMarkdown(suggestion)"
+            />
           </li>
         </ul>
       </div>
     </div>
-    <v-skeleton-loader class="welcome-message__loading" v-else />
+    <div class="welcome-message__loading" data-cy="loading" v-else>
+      <v-skeleton-loader />
+    </div>
   </div>
 </template>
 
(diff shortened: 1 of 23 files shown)


packages/components/src/components/chat/WelcomeMessageV2.vue
```diff
@@ -14,18 +14,23 @@
         </ul>
       </div>
     </div>
-    <div class="welcome-message-dynamic-placeholder" v-else>Analyzing workspace...</div>
+    <v-skeleton-loader class="welcome-message__loading" v-else />
```
kgilpin: I noticed you dropped the cue "Analyzing workspace..."

packages/components/src/components/chat/WelcomeMessageV2.vue
```diff
@@ -1,20 +1,24 @@
...
-  <div class="welcome-message" data-cy="welcome-message">
+  <div class="welcome-message" data-cy="welcome-message" data-version="2">
     <div class="welcome-message-static">
       <h3>Hi, I'm Navie!</h3>
     </div>
     <div class="welcome-message-dynamic" v-if="welcomeLoaded">
-      <div v-if="welcomeMessage" v-html="welcomeMessage" class="welcome-message-text" />
+      <div v-if="welcomeMessage" v-html="renderedWelcomeMessage" class="welcome-message-text" />
       <div v-else-if="activityName" class="welcome-message-structured">
         <p>I see you're working on {{ activityName }}. Here are some things you can try:</p>
```
kgilpin, line 9: How does this feel with "Here are some next steps:"

packages/navie/src/commands/welcome-command.ts
```diff
@@ -22,22 +22,22 @@ You have two tasks:
...
+a) Improve the code, tests, or documentation - These should be posed as a statement about the requirements for a change.
 
 b) Creating diagrams - These should be posed as a statement about the requirements for a diagram.
   Supported diagram types are sequence diagram, class diagram, entity-relationship diagram (ERD) and flowchart.
 
-Don't use an "@" command prefix in your questions.
+Don't use an "@" command prefix in your suggestion.
 
-Questions should try not to reference "changes", instead reference the specific concepts or symbols from the diff.
+Suggestions should try not to reference "changes", instead reference the specific concepts or symbols from the diff.
```
kgilpin, line 41: In a next step, it would be interesting to see if Navie is able to determine when the user might want to create or run test cases to reproduce the issue.

packages/components/src/components/chat/WelcomeMessageV2.vue
kgilpin: We could provide the text for this in the welcome message?

kgilpin (commented):
FYI I can't Approve this because I'm an author. LGTM, I left a few comments you may want to fixup.

packages/components/src/components/chat/WelcomeMessageV2.vue
dustinbyrne: Typically this style of loading element does not include any text. Generally it only takes a few seconds to complete and reveal it's purpose.

dustinbyrne (approved):

PR #2162, opened 2024-12-17
feat: Show subscription status

kgilpin:


packages/cli/src/rpc/navie/register.ts
```diff
@@ -0,0 +1,62 @@
...
+    directoryCount: configurationDirectories.length,
+    directories,
+  };
+  if (codeEditor) projectParameters.codeEditor = codeEditor;
+
+  let thread: ConversationThread | undefined;
+  try {
+    thread = await AI.createConversationThread({ modelParameters, projectParameters });
+  } catch (err) {
+    warn(`Failed to create conversation thread`);
```
kgilpin, line 46: TODO: The frontend may be expecting this to be an error, rather than a result where the thread is undefined. Maybe it would be better to update the Response type such that the thread is cannot be undefined, and raise an error here if the conversation can't be enrolled.

PR #2163, opened 2024-12-17
feat: Add a save button to assistant messages

dustinbyrne:
# Add a save button to assistant messages

Adds a new button to save the assistant's message to a file. When clicked, the button emits a `save-message` root event containing:

- `messageId` - The unique ID of the message
- `threadId` - The ID of the chat thread containing the message  
- `content` - The text content of the message

## Changes

- Added a save icon button next to copy button on assistant messages
- Updated styles to handle multiple action buttons consistently 
- Added tests to verify event emission with correct payload
- Updated `download.svg` icon to work with dynamic coloring

## Screenshots

image

## Testing

- Unit tests added to verify event payload
- Manually verified save button appears only on assistant messages
- Tested button hover and click states

The implementation assumes the code editor frontend will implement the file save dialog and write operations when receiving the `save-message` event.


packages/components/src/components/chat/UserMessage.vue
```diff
@@ -383,6 +388,13 @@ export default {
         await this.writeToClipboard(text);
       }
     },
+    saveMessage() {
+      this.$root.$emit('save-message', {
+        messageId: this.id,
+        threadId: this.threadId,
+        content: this.message,
```
kgilpin, line 395: Curious, is this going to include the gatherer output? @dividedmind 

We should probably move the gatherer feedback into its own UI element and propagate that back to the frontend via the RPC. If properly presented, it would be interesting / useful to the user to see what the gatherer is doing.

kgilpin (approved):

PR #2164, opened 2024-12-19
fix: Working changes include untracked files

dustinbyrne:
Updated the Git working directory diff functionality to include both tracked and untracked files. Previously, the system would only return changes to files already tracked by Git, which limited visibility into new files added to the project.

## Changes
- Added functionality to detect and include untracked files in working directory diffs
- Implemented platform-specific handling for null device paths
- Added comprehensive integration tests for Git operations
- Enhanced the execute command to support custom exit codes
- Added proper error handling and logging for Git operations

## Testing
New integration tests verify:
- Changes to tracked files
- New untracked files
- Command injection protection
- Error cases and edge conditions

## Technical Details
The system now combines diffs from both tracked files (via `git diff HEAD`) and untracked files (via `git diff --no-index`) to provide a complete view of working directory changes. Platform-specific handling ensures compatibility across Windows and Unix-like systems.

## Review Summary
The changes have been reviewed and found to be production-ready. Key strengths include:
- Comprehensive test coverage
- Secure command execution
- Cross-platform compatibility
- Clear error handling
- Maintainable code structure

Minor suggestions for future improvements include adding additional edge case tests and implementing a queue system for handling large numbers of untracked files, but these are not blocking issues.

### DIFF
--- a/packages/cli/src/lib/git.ts
+++ b/packages/cli/src/lib/git.ts
@@ -6,6 +6,8 @@ import { execute } from './executeCommand';
 import { warn } from 'node:console';
 import { verbose } from '../utils';
 
+const DEV_NULL = process.platform === 'win32' ? 'NUL' : '/dev/null';
+
 // Attempt to parse special syntax allowed by git:
 // - scp-style urls ([user@]host.xz:path/to/repo.git/),
 // - bare file paths.
@@ -148,6 +150,21 @@ export async function findBaseCommit(headRef = 'HEAD', cwd?: string): Promise<st
 }
 
 /**
+ * Get the list of untracked files in the working directory.
+ *
+ * @param {string} [cwd] - The working directory to run the command in.
+ * @returns {Promise<string[]>} A promise containing the list of untracked files.
+ * @throws {Error} If the command fails.
+ */
+export async function getUntrackedFiles(cwd?: string): Promise<string[]> {
+  const result = await execute('git', ['ls-files', '--others', '--exclude-standard'], { cwd });
+  return result
+    .split('\n')
+    .map((f) => f.trim())
+    .filter(Boolean);
+}
+
+/**
  * This method produces the diff between your working directory and HEAD, returning the changes that
  * have not yet been committed.
  *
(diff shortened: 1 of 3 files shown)


packages/cli/src/lib/git.ts
```diff
@@ -155,8 +180,34 @@ export async function findBaseCommit(headRef = 'HEAD', cwd?: string): Promise<st
  * @param {string} [cwd] - The working directory to run the command in.
  * @returns {Promise<string>} A promise containing the diff.
  */
-export function getWorkingDiff(cwd?: string): Promise<string> {
-  return execute('git', ['diff', 'HEAD'], { cwd });
+export async function getWorkingDiff(cwd?: string): Promise<string> {
+  const untrackedFiles = await getUntrackedFiles(cwd);
+  const untrackedDiffs: string[] = [];
```
kgilpin, line 185: We also list project files for the purpose of building the file index. So it could be useful to compare that with this and see if any other changes should be made in either place, or if the code can be shared.

packages/cli/src/lib/git.ts
```diff
@@ -155,8 +180,34 @@ export async function findBaseCommit(headRef = 'HEAD', cwd?: string): Promise<st
...
+
+  // Do this serially to avoid spawning too many processes.
+  // TODO: Perhaps this could be done in a queue.
```
kgilpin, line 188: From our experience building the file search index, nothing fancier than a loop like this really seems to be needed.

packages/cli/src/lib/git.ts
kgilpin: Also, when listing files for search, we detect and ignore binary files and also look at how big they are.

dustinbyrne: It looks like these are also not considered by search. I can create an issue.

In this instance, binary files aren't much of a consideration as they won't render in the diff view, but there's still potential for a large diff to overflow the token limit. In that case, it'd fall back to the default welcome message.

packages/cli/tests/integration/git.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+  const commitFile = async (file: string, message: string) => {
+    await execute('git', ['add', file], { cwd: projectDirectory });
+    await execute('git', ['commit', '-m', message], { cwd: projectDirectory });
+  };
+
+  const createBranch = (branch: string) =>
+    execute('git', ['checkout', '-b', branch], { cwd: projectDirectory });
+
+  beforeEach(async () => {
+    projectDirectory = await mkdtemp(join(tmpdir(), 'appmap-git-test-'));
```
dividedmind, line 25: This package already uses `tmp` and `tmp-promise` packages which might be more convenient and structured for tmpdir creation.

packages/cli/tests/integration/git.spec.ts
```diff
@@ -0,0 +1,73 @@
+import { tmpdir } from 'node:os';
+import { mkdir, mkdtemp, rm, writeFile } from 'node:fs/promises';
+import { dirname, join } from 'node:path';
```
dividedmind, line 3: Pet peeve: unsorted imports (there's a vscode command for that!) ;)

packages/cli/src/lib/git.ts
```diff
@@ -147,6 +149,29 @@ export async function findBaseCommit(headRef = 'HEAD', cwd?: string): Promise<st
...
+        .filter(Boolean)
+    )
+    .catch((e: unknown) => {
+      if (verbose()) {
+        warn(`Failed to retrieve untracked files`);
+        warn(String(e));
+      }
+      return [];
+    });
+}
```
dividedmind, line 173: I feel this would be a bit clearer as an `async` function.

dividedmind (approved):

packages/cli/tests/integration/git.spec.ts
dustinbyrne: Can we configure ESLint to do enforce or fix this for us?

PR #2168, opened 2024-12-20
fix: Fix a typo in `/gather` option

dividedmind:
In refactoring the `/gather` flag somehow got changed to `/gatherer`. This change changes it back.


kgilpin (approved):

kgilpin:
@dividedmind keep in mind, as always, this will not be visible to users until the CLI is released.

PR #2184, opened 2025-01-08
feat: Disable the selection of help mode by default

kgilpin:
Fixes https://github.com/getappmap/appmap-js/issues/2182


packages/cli/src/cmds/sequenceDiagram/browserRenderer.ts
```diff
@@ -49,6 +49,10 @@ export default class BrowserRenderer {
     const puppeteer = require('puppeteer');
 
     if (verbose()) console.warn(`Preparing browser for PNG rendering`);
-    this._browser = await puppeteer.launch({ timeout: 120 * 1000, headless: !this.showBrowser });
+    this._browser = await puppeteer.launch({
+      timeout: 120 * 1000,
+      headless: !this.showBrowser,
+      args: ['--no-sandbox'],
```
dustinbyrne, line 55: I ended up skipping the test that broke instead of disabling the sandbox. It's in `main` now, but feel free to replace it with this.

dustinbyrne: The ruby action is fixed as well. I'm not sure why these decided to break all of a sudden.

dustinbyrne (approved):

PR #2189, opened 2025-01-09
feat: Update search and rpc packages in CLI

dividedmind:
I've noticed recent search improvements haven't been released with the newest CLI; this bumps the version. I also bumped RPC while I was at it. I didn't know if components needs to be updated too, so I left it alone.


dividedmind:
There seems to be something else going on with the search; I don't think this is necessary after all.

PR #2213, opened 2025-01-21
fix: Tokenization performance

dustinbyrne:
## Overview

This pull request addresses several issues related to the tokenization process and efficiency improvements across search. The key changes are described below:

### `search`

- **Tokenization Process Improvement**: 
  - The tokenization process has been updated to prevent the process from hanging. Tokenization is now performed in batches which periodically yield back to the main event loop, enhancing the system's responsiveness.
  - The `indexFile` function now handles asynchronous tokenization, and adjustments were made to accommodate the change to a promise-based return from the `tokenizer` function.
  - A new `batch` function was introduced to handle line batching, ensuring efficient processing by taking into account line lengths and batch sizes.

- **Accuracy Improvements**: 
  - The file extension is now considered in the tokenization process, which aids in determining the appropriate parsing strategy for each file type.

- **Test Additions**: 
  - Tests for batch processing and adjustments to the tokenization workflow have been included to validate the new functionality and performance of the tokenization strategy.

Resolves https://github.com/getappmap/appmap-js/issues/2204

### DIFF
--- a/packages/cli/src/rpc/explain/index/appmap-index.ts
+++ b/packages/cli/src/rpc/explain/index/appmap-index.ts
@@ -126,7 +126,9 @@ export async function readAppMapContent(appmapFile: string): Promise<string> {
   appmapWords.push(...parameters);
   appmapWords.push(...types);
 
-  return appmapWords.join(' ');
+  // Words are separated by new lines to reduce the time spent tokenizing the content.
+  // Long lines may be slower to tokenize or be skipped altogether.
+  return appmapWords.join('\n');
 }
 
 export function trueFilter(): Promise<boolean> {
--- a/packages/cli/tests/unit/rpc/explain/index/appmap-index.spec.ts
+++ b/packages/cli/tests/unit/rpc/explain/index/appmap-index.spec.ts
@@ -63,7 +63,7 @@ describe('readAppMapContent', () => {
     expect(content).toContain('sql');
     expect(content).toContain('database');
 
-    expect(content.split(' ')).toEqual([
+    expect(content.split(/\s+/)).toEqual([
       'Test',
       'AppMap',
       'test',
--- a/packages/search/src/splitter.ts
+++ b/packages/search/src/splitter.ts
@@ -47,6 +47,7 @@ export async function langchainSplitter(content: string, fileExtension: string):
     splitter = new RecursiveCharacterTextSplitter();
   }
   const documents = await splitter.createDocuments([content]);
+  const contentLines = content.split('\n');
 
   // metadata includes:
   // { loc: { lines: { from: 1, to: 14 } } }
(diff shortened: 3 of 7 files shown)


packages/cli/src/rpc/explain/index/appmap-index.ts
```diff
@@ -126,7 +126,9 @@ export async function readAppMapContent(appmapFile: string): Promise<string> {
   appmapWords.push(...parameters);
   appmapWords.push(...types);
 
-  return appmapWords.join(' ');
+  // Words are separated by new lines to reduce the time spent tokenizing the content.
+  // Long lines may be slower to tokenize or be skipped altogether.
+  return appmapWords.join('\n');
```
dustinbyrne, line 131: My assessment here is that changing this whitespace is inconsequential, except in the scenario described in the comment.

packages/cli/tests/unit/rpc/explain/index/appmap-index.spec.ts
```diff
@@ -63,7 +63,7 @@ describe('readAppMapContent', () => {
     expect(content).toContain('sql');
     expect(content).toContain('database');
 
-    expect(content.split(' ')).toEqual([
+    expect(content.split(/\s+/)).toEqual([
```
dustinbyrne, line 66: The content is no longer joined by spaces, so splitting on whitespace restores the previous behavior. This is functionally the same as how we're [splitting words in tokenize](https://github.com/getappmap/appmap-js/blob/355d931458dded962a186de86722c451e75c530d/packages/search/src/tokenize.ts#L59), even though the regex is different.

packages/search/src/splitter.ts
```diff
@@ -47,6 +47,7 @@ export async function langchainSplitter(content: string, fileExtension: string):
     splitter = new RecursiveCharacterTextSplitter();
   }
   const documents = await splitter.createDocuments([content]);
+  const contentLines = content.split('\n');
```
dividedmind, line 50: Ouch! :/ Good catch, thanks!

kgilpin (approved):
I know you guys have been over this a few times, and it looks good to me.

packages/search/src/tokenize.ts
```diff
@@ -59,23 +59,55 @@ export function words(content: string): string[] {
...
+ */
+export function batch(content: string, batchSize = 1000, maxLineLength = 1000): string[] {
+  const lines = content.split('\n').filter(({ length }) => length <= maxLineLength);
+  const result = [];
+  for (let i = 0; i < lines.length; i += batchSize) {
+    result.push(lines.slice(i, i + batchSize).join('\n'));
+  }
+
+  return result;
+}
```
dividedmind, line 79: Have you considered implementing this as an async generator instead? IMO it'll make the flow clearer and you won't have to muck with the event loop directly.

dividedmind (approved):
Looks good, great catches! I'd refactor to use an async generator instead, but I'll leave it up to you whether that's something you want to do.

packages/search/src/tokenize.ts
dividedmind: If you're generating batches asynchronously then it goes through the event loop, so you don't need `setImmediate`. Ie.
```typescript
async *batch(): AsyncGenerator<string> {
  for (...) yield chunk;
}

for await (const chunk of batch())
  doSomething(chunk);
```

PR #2217, opened 2025-01-23
feat: Skip large diffs when generating context for the LLM

### LINKED ISSUE #2173: Ignore large files when building the diff for welcome and review purposes
When examining the diff to suggest user questions and to perform code review, ignore large file diffs. A typical example would be yarn.lock.

dividedmind:
This skips diffs larger than 1000 characters when generating input for `@welcome` and `@review`. Large diffs are replaced with a message, for example

    diff --git a/file2.txt b/file2.txt
    [Change of size 2078]

Fixes #2173


packages/cli/src/rpc/explain/review.ts
```diff
@@ -27,7 +28,10 @@ export default async function handleReview(
       {
         type: 'code-snippet',
         location: REVIEW_DIFF_LOCATION, // eslint-disable-line @typescript-eslint/no-unsafe-assignment
-        content: diffContent.filter(Boolean).join('\n\n'),
+        content: diffContent
+          .filter(Boolean)
+          .map((diff) => processPatchset(diff))
+          .join('\n\n'),
```
kgilpin, line 34: I have been looking into this type of limiting as well, and one of the issues I see is that `getDiffLog` returns, as the name suggests, `git log` output not `git diff` output. So, splitting this output would seem to be more naturally done on the `commit <sha>` (author) rather than on the `diff --git` line.

WDYT?

```
commit 5f7fed0d7fab55e9b2cdc6cddbebd9644bca416c
Author: Kevin Gilpin <kgilpin@gmail.com>
Date:   Thu Jan 23 09:49:53 2025 -0500

    refactor: Remove unused code
    
    Locating the git repository - can't find any references to this any more

diff --git a/packages/cli/src/lib/git.ts b/packages/cli/src/lib/git.ts
index 5bd8b0a03..139b10eca 100644
--- a/packages/cli/src/lib/git.ts
+++ b/packages/cli/src/lib/git.ts
```

kgilpin: In other words, maybe it would be better / easier if `getDiffLog` and `getWorkingDiff` returned an output that had some structure (at least an array of chunks?) rather than just a string that the caller then has to figure out how to split.

dustinbyrne (approved):

PR #2231, opened 2025-02-04
fix: Windows in-memory history support

dustinbyrne:
WIP, might need some basic persistence to load active threads on restart.


packages/search/src/snippet-index.ts
```diff
@@ -60,7 +60,10 @@ export type SnippetId = {
 export function fileChunkSnippetId(filePath: string, startLine?: number): SnippetId {
   return {
     type: 'file-chunk',
-    id: [filePath, startLine].filter(Boolean).join(':'),
+    id: [filePath, startLine]
+      .filter((t): t is string | number => Boolean(t))
```
kgilpin, line 64: Can you describe how the old behavior was causing the drive letter bug?

dustinbyrne: Sure, given the following snippet identifier via `[filePath, startLine]`:
```
> ['C:\\Users\\Dustin\\README.md', '1'].filter(Boolean).join(':')
'C:\\Users\\Dustin\\README.md:1'
```

When this string is split into `filePath`, `startLine` below in `parseFileChunkSnippetId`:
```
> 'C:\\Users\\Dustin\\README.md:1'.split(':')
[ 'C', '\\Users\\Dustin\\README.md', '1' ]
```

We then take the first element as the filePath and the second as the startLine, resulting in the path being `C`.
URL encoding the filepath prevents a colon in the path from breaking this process.

dustinbyrne: I'll add a test case for this.

kgilpin: Got it, thanks.

kgilpin (approved):

PR #2233, opened 2025-02-08
fix: Normalize Windows path locations

dustinbyrne:
In practice, Windows paths are represented in many different forms. Path separators are interchangable and paths are case insensitive. Additionally, paths constructed via a file scheme URI may contain a leading slash, similar to an absolute path in POSIX.

This was preventing pinned items from being picked up in Windows. A pinned file path may be something like:
```
/c:/Users/Administrator/Project/app.rb:2-24
```

This is problematic for a few reasons:
1. The leading slash corrupts the path and it is not usable. Ideally clients would not send us paths formatted in this manner, but it's already happening. This behavior is observed both from JetBrains and Visual Studio Code.
2. The path separators are valid, but don't match other cases in our code, such as when we compare the absolute file path against known project directories. In this case, known project directories will use the default separator (\\), so [checks to see if a file starts with one of these paths will always fail](https://github.com/getappmap/appmap-js/blob/3e0f4ae7aabb3d7b1a27b8f54135d39a6e4b79d9/packages/cli/src/rpc/explain/collect-location-context.ts#L40).
3. A case inconsistency in the drive letter will also cause the above scenario to fail. Windows doesn't actually care, but our string comparisons obviously do.

### DIFF
--- a/packages/cli/src/rpc/explain/location.ts
+++ b/packages/cli/src/rpc/explain/location.ts
@@ -1,4 +1,21 @@
 import { warn } from 'node:console';
+import { platform } from 'node:os';
+
+function normalizePath(location: string): string {
+  if (platform() !== 'win32') return location;
+
+  // This fixes up a few issues observed with Windows paths:
+  // 1. Leading slashes can be present in the path. This is likely an artifact of how the path is
+  //    constructed in the client (e.g., via a file scheme URI).
+  // 2. Drive letters are down cased. Again, likely due to file URIs.
+  // 3. Path separators are normalized to backslashes.
+  //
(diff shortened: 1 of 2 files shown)


packages/cli/src/rpc/explain/location.ts
```diff
@@ -34,19 +51,21 @@ export default class Location {
   }
 
   static parse(location: string): Location {
-    const tokens = location.split(':');
```
dustinbyrne, line 37: I forgot to mention this in the PR description, but splitting here on `:` is also problematic for absolute paths in Windows. I've rewritten this method to perform the path and line range parsing via regex.

kgilpin (approved):

PR #2236, opened 2025-02-11
Enhance JSON parsing and improve vector search with pinned items

dustinbyrne:
This PR addresses the following:

1. JSON Parsing Enhancements:
   - Improved parsing of JSON responses, especially for cases with or without code fences
   - Added better support for multi-line JSON objects
   - Implemented a more robust JSON object detection algorithm
   - Added fallback parsing strategies for different response formats

2. Vector Search Improvements:
   - Updated system prompt terminology from "question" to "inquiry" for better LLM compatibility
     - I've observed Sonnet saying things like "you didn't provide a question" to inquiries such as "analyze this".
   - Fixed pinned items inclusion in vector terms generation

### DIFF
--- a/packages/navie/src/lib/parse-json.ts
+++ b/packages/navie/src/lib/parse-json.ts
@@ -16,7 +16,7 @@ function findArray(text: string): string | undefined {
  * If found, return the text within the braces. This is a fallback
  * in case proper JSON cannot be found within trimmed fences.
  */
-function findObject(text: string): string | undefined {
+export function findObject(text: string): string | undefined {
   const object = text.match(/\{.*\}/s);
   return object ? object[0] : undefined;
 }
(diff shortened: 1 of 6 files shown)


packages/navie/src/services/context-service.ts
```diff
@@ -45,7 +45,7 @@ export default class ContextService {
 
       const searchTerms = await transformSearchTerms(
         termsEnabled,
-        options.aggregateQuestion,
+        aggregateQuestion.join('\n\n'),
```
dividedmind, line 48: Great catch!

packages/navie/src/lib/parse-json.ts
```diff
@@ -16,9 +16,21 @@ function findArray(text: string): string | undefined {
...
+    if (text[i] === '{') {
+      depth++;
+      if (depth === 1) start = i;
+    } else if (text[i] === '}') {
+      depth--;
+      if (depth === 0) {
+        return text.slice(start, i + 1);
+      }
+    }
+  }
```
dividedmind, line 33: Is this change necessary? The spec doesn't include any examples where the old, simpler code doesn't work.

dividedmind (approved):
I'm not sure about the change to `findObject`, but other than that, it looks good.

packages/navie/src/lib/parse-json.ts
dustinbyrne: The previous implementation only found JSON objects in a single line. I've observed Claude Sonnet generating multi-line JSON outside of code fences, so I decided to make this change to parse it our manually.

PR #2247, opened 2025-02-17
Improving generation consistency

dustinbyrne:
1. Pinned items are now presented solely in the user context. Previously they were split between user and system context, where a path to the file would be presented in user context but the actual contents were written to the system context. It'd be up to the LLM to make the association.
2. Some changes to the `@plan` prompt to nudge it more towards presenting relative file paths in the plan output. With gpt-4o, plan would often only include a basename or an identifier found in the code.
3. User selected context paths are now normalized on ingress so that they don't need to be spot fixed where ever we're comparing paths internally.
4. Some additional logging is added to help troubleshoot cases where Navie becomes unresponsive
5. `@diagram` now has a default prompt. Without user input, some models would tend to completely ignore the system prompt.

### DIFF
--- a/packages/navie/src/agents/diagram-agent.ts
+++ b/packages/navie/src/agents/diagram-agent.ts
@@ -1,7 +1,7 @@
 import { Agent, AgentOptions } from '../agent';
 import InteractionHistory, { PromptInteractionEvent } from '../interaction-history';
 import MermaidFilter from '../lib/mermaid-filter';
-import { PROMPTS, PromptType } from '../prompt';
+import { buildPromptValue, PROMPTS, PromptType } from '../prompt';
 import ContextService from '../services/context-service';
 import MermaidFixerService from '../services/mermaid-fixer-service';
 import { DIAGRAM_FORMAT_PROMPT } from './explain-agent';
@@ -17,6 +17,8 @@ Your job is to generate software diagrams based on a description provided by the
 The user is an experienced software developer who will review the diagrams you generate, and use them
 to understand the code and design code solutions. You can expect the user to be proficient in software development.
 `;
+
+const DEFAULT_QUESTION = 'Create a diagram from the information available.';
 
 export default class DiagramAgent implements Agent {
   public temperature = undefined;
@@ -53,6 +55,8 @@ export default class DiagramAgent implements Agent {
   }
 
   applyQuestionPrompt(question: string): void {
-    this.history.addEvent(new PromptInteractionEvent(PromptType.Question, 'user', question));
+    this.history.addEvent(
+      new PromptInteractionEvent(PromptType.Question, 'user', question.trim() || DEFAULT_QUESTION)
+    );
   }
 }
--- a/packages/navie/src/services/openai-completion-service.ts
+++ b/packages/navie/src/services/openai-completion-service.ts
@@ -19,6 +19,7 @@ import { APIError } from 'openai';
 import MessageTokenReducerService from './message-token-reducer-service';
 import { findObject, tryParseJson } from '../lib/parse-json';
 import trimFences from '../lib/trim-fences';
+import { performance } from 'node:perf_hooks';
 
 /*
   Generated on https://openai.com/api/pricing/ with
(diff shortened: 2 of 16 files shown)


packages/navie/src/services/openai-completion-service.ts
```diff
@@ -255,7 +255,11 @@ export default class OpenAICompletionService implements CompletionService {
     const maxRetries = options?.maxRetries ?? 3;
     for (let i = 0; i < maxRetries; i += 1) {
       const generateJson = this.isLocalModel ? generateLocal : generateOpenAi;
+      warn(`Requesting JSON completion`);
+      const startTime = new Date();
       const response = await generateJson(); // eslint-disable-line no-await-in-loop
+      const endTime = new Date();
+      warn(`Received JSON response in ${(endTime.valueOf() - startTime.valueOf()) / 1000}s`);
```
dividedmind, line 262: Not that it matters terribly here, but this [is not the correct way to measure elapsed time](https://blog.insiderattack.net/how-not-to-measure-time-in-programming-11089d546180) — `Date()` is not designed for this and can jump around eg. due to NTP adjustments. You can use `process.hrtime()` or `performance.now()` to get a monotonic clock. (Also, `node:console` offers `time()` and `timeEnd()` functions for measuring intervals which might be easier to use.)

Also, I'd skip the `requesting completion` message; Navie is already chatty as it is on the console and this message will just add to the noise.

packages/navie/src/services/openai-completion-service.ts
```diff
@@ -335,7 +339,13 @@ export default class OpenAICompletionService implements CompletionService {
 
     for (let attempt = 0; attempt < CompletionRetries; attempt++) {
       try {
+        warn(`Requesting completion`);
+        const startTime = new Date();
         const response = await fetchResponse();
+        const endTime = new Date();
+        warn(
+          `Received completion response in ${(endTime.valueOf() - startTime.valueOf()) / 1000}s`
+        );
```
dividedmind, line 348: As above.

packages/navie/src/agents/diagram-agent.ts
```diff
@@ -53,6 +55,12 @@ export default class DiagramAgent implements Agent {
   }
 
   applyQuestionPrompt(question: string): void {
-    this.history.addEvent(new PromptInteractionEvent(PromptType.Question, 'user', question));
+    this.history.addEvent(
+      new PromptInteractionEvent(
+        PromptType.Question,
+        'user',
+        buildPromptValue(PromptType.Question, question.trim() || DEFAULT_QUESTION)
```
dividedmind, line 62: This change will add `<question>` tags around the question — is this intended?

dividedmind (approved):

packages/navie/src/user-context.ts
```diff
@@ -36,17 +39,38 @@ export namespace UserContext {
...
+        // If it's found, use its content, otherwise only the location is rendered
+        const content =
+          options.interactionHistory && options.includeContent
+            ? options.interactionHistory.events.find((e): e is ContextItemEvent => {
+                if (e.type !== 'contextItem') return false;
+                const { location, directory } = e as ContextItemEvent;
+                return (
+                  location === i.location ||
+                  join(...([directory, location].filter(Boolean) as string[])) === i.location
+                );
```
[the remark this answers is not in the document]
dustinbyrne: I agree, the coupling here is not ideal. However, having said that, I don't love the idea of having to mutate context from the user either. It _should_ have content, yes, I agree, but by doing so it'd be classified as a code selection (or code snippet? what's the difference?) and I'm just not sure if there's any repercussions to doing so.

I think the way we're handling user context / pinned items / code snippets / code selections in general needs an overhaul. Maybe the context service should be responsible for resolving user context items, or maybe, it should be its own service. I kind of like the separate service idea. It feels to me like a lot of things need to move around but I'm most concerned with getting this working in the short term.

I've added a TODO to the code to pull out this dependency.

packages/navie/src/agents/diagram-agent.ts
dustinbyrne: I don't know if it'd cause any issue. I pulled it out.

PR #2255, opened 2025-02-28
feat: Replace LLM configuration with model selector

dustinbyrne:
This PR implements a new model selector UI and supporting backend services, allowing users to choose between different AI models for Navie.

### @appland/rpc
- Added new RPC method definitions for model management (`v1.navie.models`)
- Includes methods for adding, listing, selecting, and getting model configurations

### @appland/navie
- Added support for per-message model overrides
- Implemented Ollama completion service
- Enhanced completion service factory to handle different model providers
- Added configuration for model-specific token limits
- Added support for custom API endpoints and keys

### @appland/components
- Added new ModelSelector.vue component with dropdown and configuration UI
- Integrated model selector with Chat.vue
- Updated ChatSearch.vue to handle model selection
- Added styling for model badges and configuration UI
- Added unit tests for ModelSelector component

### @appland/cli
- Implemented ModelRegistry to manage available AI models
- Added handlers for model-related RPC methods
- Added model refresh functionality to fetch available models from providers
- Added support for OpenAI, Anthropic, and Ollama model discovery
- Added unit tests for ModelRegistry

The changes allow users to select and configure different AI models through the UI, with support for standard providers (OpenAI, Anthropic) and local models (Ollama). The implementation includes proper configuration management, model discovery, and a user-friendly interface for model selection.

This should **not affect** existing code editor extensions. Code editor extensions must be upgraded to provide external models and, export `APPMAP_NAVIE_MODEL_SELECTOR`, set and persist the selected model.

https://github.com/user-attachments/assets/58abdd13-a7e6-4165-8e27-a5fe7b52769f

### DIFF
--- a/packages/cli/src/rpc/navie/isCustomWelcomeMessageEnabled.ts
+++ b/packages/cli/src/rpc/navie/isCustomWelcomeMessageEnabled.ts
@@ -1,5 +1,18 @@
+import ModelRegistry from './models/registry';
+
 export default function isCustomWelcomeMessageEnabled(): boolean {
   // This behavior is DEFAULT ON, but can be disabled by setting the environment variable
-  // NAVIE_CUSTOM_WELCOME_MESSAGE=false. Any other value will be treated as true.
-  return process.env.NAVIE_CUSTOM_WELCOME_MESSAGE !== 'false';
+  // NAVIE_CUSTOM_WELCOME_MESSAGE=false.
+  if (process.env.NAVIE_CUSTOM_WELCOME_MESSAGE === 'false') {
+    return false;
+  }
+
+  // If the model selector is available, the custom message can only be rendered if a model is
+  // selected.
+  if (process.env.APPMAP_NAVIE_MODEL_SELECTOR) {
+    return ModelRegistry.instance.selectedModel !== undefined;
+  }
+
+  // Otherwise, default on.
+  return true;
 }
--- a/packages/cli/src/rpc/navie/models/handlers/getConfig.ts
+++ b/packages/cli/src/rpc/navie/models/handlers/getConfig.ts
@@ -1,0 +1,38 @@
+import { NavieRpc } from '@appland/rpc';
+import { RpcHandler } from '../../../rpc';
+
+/**
+ * This function is used to print a secret value as asterisks. If the value doesn't exists,
+ * it returns undefined.
+ * @param value the secret value
+ * @returns the secret value as asterisks or undefined
+ */
+const formatSecret = (value?: string) => (value?.length ? '*'.repeat(value.length) : undefined);
+
+export function navieModelsGetConfigV1(): RpcHandler<
+  NavieRpc.V1.Models.GetConfig.Params,
+  NavieRpc.V1.Models.GetConfig.Response
+> {
+  return {
+    name: NavieRpc.V1.Models.GetConfig.Method,
+    handler: () => [
+      {
+        provider: 'openai',
+        apiKey: formatSecret(process.env.OPENAI_API_KEY),
+        endpoint: process.env.OPENAI_BASE_URL,
+      },
+      {
+        provider: 'anthropic',
+        apiKey: formatSecret(process.env.ANTHROPIC_API_KEY),
+      },
+      {
+        provider: 'google',
+        apiKey: formatSecret(process.env.GOOGLE_WEB_CREDENTIALS),
+      },
+      {
+        provider: 'ollama',
+        endpoint: process.env.OLLAMA_BASE_URL,
+      },
+    ],
+  };
+}
--- a/packages/cli/src/rpc/navie/models/registry.ts
+++ b/packages/cli/src/rpc/navie/models/registry.ts
@@ -1,0 +1,131 @@
+import { NavieRpc } from '@appland/rpc';
+import { verbose } from '../../../utils';
+import { fetchAnthropicModels } from './providers/anthropic';
+import { fetchOllamaModels } from './providers/ollama';
+import { fetchOpenAIModels } from './providers/openai';
+
+const findModel = (
+  arr: NavieRpc.V1.Models.Model[],
+  id: string,
+  provider?: string
+): NavieRpc.V1.Models.Model | undefined => {
+  const lowerId = id.toLowerCase();
+  const lowerProvider = provider?.toLowerCase();
+  return arr.find(
+    (model) =>
+      model.id.toLowerCase() === lowerId &&
+      (!provider || model.provider.toLowerCase() === lowerProvider)
+  );
+};
+
+export default class ModelRegistry {
+  private readonly recommendedModels = [/claude-3[-.]5-sonnet/, /gpt-4o/];
+  private readonly notRecommendedModels = [/^o1-?.*$/, /^o3-?.*$/, /^claude-3[-.]7-sonnet-?.*$/];
+  private clientModels = new Map<string, NavieRpc.V1.Models.ClientModel>();
+  private apiModels: NavieRpc.V1.Models.Model[] = [];
+  private _selectedModel?: NavieRpc.V1.Models.Model;
+
+  static readonly instance = new ModelRegistry();
+
+  private constructor() {
+    // no-op
+  }
+
+  select(modelId: string | undefined) {
+    if (!modelId) {
+      if (verbose()) console.log(`[ModelRegistry] Unselecting active model`);
+      this._selectedModel = undefined;
+      return;
+    }
+
+    let provider: string | undefined;
+    let id: string;
+    const lowerModelId = modelId.toLowerCase();
+    const delimiterIndex = lowerModelId.indexOf(':');
+    if (delimiterIndex === -1) {
+      id = decodeURIComponent(lowerModelId.toLowerCase());
+    } else {
+      id = decodeURIComponent(lowerModelId.slice(delimiterIndex + 1));
+      provider = decodeURIComponent(lowerModelId.slice(0, delimiterIndex));
+    }
+
+    let model = findModel(this.apiModels, id, provider);
+    model = model ?? findModel(Array.from(this.clientModels.values()), id, provider);
+    if (model) {
+      console.log(`[ModelRegistry] Selecting model "${modelId}"`);
+    } else if (!model) {
+      console.warn(`[ModelRegistry] Model ${modelId} not found`);
+    }
+
+    this._selectedModel = model;
+  }
+
+  get selectedModel(): NavieRpc.V1.Models.Model | undefined {
+    return this._selectedModel;
+  }
+
+  add(model: NavieRpc.V1.Models.ClientModel) {
+    const key = `${model.provider.toLowerCase()}:${model.id.toLowerCase()}`;
+    this.clientModels.set(key, model);
+  }
+
+  list(): NavieRpc.V1.Models.ListModel[] {
+    const listModels: NavieRpc.V1.Models.ListModel[] = Array.from(this.clientModels.values())
+      .map((m) => {
+        const model = {
+          id: m.id,
+          name: m.name,
+          provider: m.provider,
+          createdAt: m.createdAt,
+        } as NavieRpc.V1.Models.Model;
+        if (m.maxInputTokens !== undefined) {
+          model.maxInputTokens = m.maxInputTokens;
+        }
+        return model;
+      })
+      .concat(this.apiModels)
+      .map((m) => ({ ...m })); // shallow copy to avoid mutating the original objects
+
+    const recommendation = this.recommendedModels.find((regex) =>
+      listModels.some((m) => regex.test(m.id))
+    );
+    for (const model of listModels) {
+      if (recommendation?.test(model.id)) {
+        model.tags = ['recommended'];
+      }
+      if (model.id === this.selectedModel?.id) {
+        if (model.tags) model.tags.push('primary');
+        else model.tags = ['primary'];
+      }
+      if (this.notRecommendedModels.some((regex) => regex.test(model.id))) {
+        if (model.tags) model.tags.push('alpha');
+        else model.tags = ['alpha'];
+      }
+    }
+
+    return listModels;
+  }
+
+  async refresh() {
+    const promises: Promise<NavieRpc.V1.Models.Model[]>[] = [];
+    const { ANTHROPIC_API_KEY, OPENAI_API_KEY, OPENAI_BASE_URL } = process.env;
+
+    if (ANTHROPIC_API_KEY) {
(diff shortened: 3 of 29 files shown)


kgilpin:
Impressive :-)

packages/cli/src/rpc/navie/isCustomWelcomeMessageEnabled.ts
```diff
@@ -1,5 +1,18 @@
...
+  // NAVIE_CUSTOM_WELCOME_MESSAGE=false.
+  if (process.env.NAVIE_CUSTOM_WELCOME_MESSAGE === 'false') {
+    return false;
+  }
+
+  // If the model selector is available, the custom message can only be rendered if a model is
+  // selected.
+  if (process.env.APPMAP_NAVIE_MODEL_SELECTOR) {
+    return ModelRegistry.instance.selectedModel !== undefined;
+  }
```
dustinbyrne, line 14: This `APPMAP_NAVIE_MODEL_SELECTOR` environment variable controls the behavior of the welcome message. If the model selector is available, this environment variable should be set. When it is set, it prevents the welcome message from attempting to generate completions unless a model is actively selected.

packages/components/tests/e2e/specs/chat/chatSearch.spec.js
```diff
@@ -6,13 +6,6 @@ context('Chat search', () => {
     cy.viewport(1280, 720);
   });
 
-  it('assistant response can be stopped', () => {
-    cy.get('[data-cy="chat-input"]', { timeout: 25000 }).clear().type('Hello world{enter}');
-    cy.get('[data-cy="stop-response"]').should('exist');
-    cy.get('[data-cy="stop-response"]').click();
-    cy.get('[data-cy="stop-response"]').should('not.exist');
-  });
-
```
dustinbyrne, line 15: This is already a unit test.

packages/navie/src/services/openai-completion-service.ts
```diff
@@ -181,11 +181,17 @@ export default class OpenAICompletionService implements CompletionService {
...
+    private apiUrl?: string,
+    private apiKey?: string
   ) {
     this.model = new ChatOpenAI({
       modelName: this.modelName,
       temperature: this.temperature,
+      configuration: {
+        baseURL: this.apiUrl ?? process.env.OPENAI_BASE_URL,
+      },
+      apiKey: this.apiKey ?? process.env.OPENAI_API_KEY,
```
dividedmind, line 194: Maybe leave just `this.apiUrl`/`this.apiKey` and add a comment that if they're undefined langchain will automatically use values from the environment if present? As it is, it puzzled me for a moment to consider what would happen in the Azure case (it will in fact work correctly, at least I think so).

packages/navie/test/services/completion-service-factory.spec.ts
```diff
@@ -2,6 +2,9 @@ import createCompletionService from '../../src/services/completion-service-facto
 import OpenAICompletionService from '../../src/services/openai-completion-service';
 import AnthropicCompletionService from '../../src/services/anthropic-completion-service';
 import Trajectory from '../../src/lib/trajectory';
+import { NavieModel } from '../../src/navie';
+import OllamaCompletionService from '../../src/services/ollama-completion-service';
+import GoogleVertexAICompletionService from '../../src/services/google-vertexai-completion-service';
```
dividedmind, line 7: nits: NavieModel should be `import type` and I like to (incrementally) sort the imports when I make changes — in this case the completion services naturally want to be together.

packages/navie/test/services/completion-service-factory.spec.ts
```diff
@@ -12,15 +15,41 @@ describe('CompletionServiceFactory', () => {
...
+    process.env.ANTHROPIC_API_KEY = 'anthropic-key';
+    const trajectory: Trajectory = jest.createMockFromModule<Trajectory>(
+      '../../src/lib/trajectory'
+    );
+    const examples: (NavieModel & { expected: unknown })[] = [
+      { id: 'gpt-4o', provider: 'openai', expected: OpenAICompletionService },
+      { id: 'claude-2', provider: 'anthropic', expected: AnthropicCompletionService },
+      { id: 'deepseek-r1:8b', provider: 'ollama', expected: OllamaCompletionService },
+      { id: 'gemini-1.5-turbo', provider: 'vertex-ai', expected: GoogleVertexAICompletionService },
+    ];
```
dividedmind, line 44: I think you can avoid explicitly typing this by saying `const examples = [ /* ... */ ] as const;`. (Then you shouldn't need the NavieModel import.)

packages/navie/test/services/completion-service-factory.spec.ts
```diff
@@ -12,15 +15,41 @@ describe('CompletionServiceFactory', () => {
 
   it('creates an OpenAICompletionService by default', () => {
     process.env.OPENAI_API_KEY = 'openai-key';
-    const trajectory: Trajectory = jest.genMockFromModule<Trajectory>('../../src/lib/trajectory');
+    const trajectory: Trajectory = jest.createMockFromModule<Trajectory>(
+      '../../src/lib/trajectory'
+    );
```
dividedmind, line 20: Just wondering, since you're already changing this — I think that if you add `jest.mock('../../src/lib/trajectory')` in the preamble then a simple `new Trajectory()` should work? Actually I don't know if there's even any point mocking.

packages/cli/src/rpc/navie/models/handlers/getConfig.ts
```diff
@@ -0,0 +1,38 @@
+import { NavieRpc } from '@appland/rpc';
+import { RpcHandler } from '../../../rpc';
+
+/**
+ * This function is used to print a secret value as asterisks. If the value doesn't exists,
+ * it returns undefined.
+ * @param value the secret value
+ * @returns the secret value as asterisks or undefined
+ */
+const printSecret = (value?: string) => (value?.length ? '*'.repeat(value.length) : undefined);
```
dividedmind, line 10: IMO it should just print a constant number of asterisks. Not that it matters terribly here ;) but a secret's length is also secret! (Also, a nit: this doesn't actually print anything ;p)

packages/cli/src/rpc/navie/models/handlers/getConfig.ts
```diff
@@ -0,0 +1,38 @@
...
+        apiKey: printSecret(process.env.OPENAI_API_KEY),
+        endpoint: process.env.OPENAI_BASE_URL,
+      },
+      {
+        provider: 'anthropic',
+        apiKey: printSecret(process.env.ANTHROPIC_API_KEY),
+      },
+      {
+        provider: 'google',
+        apiKey: printSecret(process.env.GOOGLE_WEB_CREDENTIALS),
```
dividedmind, line 30: I wonder — why even bother returning the API keys?

packages/cli/src/rpc/navie/models/registry.ts
```diff
@@ -0,0 +1,281 @@
...
+        createdAt: normalizeDate(model.modified_at),
+      }))
+    );
+  } catch (e) {
+    if (verbose()) {
+      console.error(`failed to fetch ollama models: ${String(e)}`);
+    }
+  }
+  return models;
+}
```
dividedmind, line 155: Feels like these should live in their respective providers.

packages/cli/src/rpc/navie/models/registry.ts
```diff
@@ -0,0 +1,281 @@
...
+  async refresh() {
+    const promises: Promise<NavieRpc.V1.Models.Model[]>[] = [];
+    const { ANTHROPIC_API_KEY, OPENAI_API_KEY, OPENAI_BASE_URL } = process.env;
+
+    if (ANTHROPIC_API_KEY) {
+      promises.push(fetchAnthropicModels(ANTHROPIC_API_KEY));
+    }
+
+    if (OPENAI_API_KEY && (!OPENAI_BASE_URL || OPENAI_BASE_URL === 'https://api.openai.com')) {
+      promises.push(fetchOpenAIModels(OPENAI_API_KEY));
```
dividedmind, line 268: I think this won't work with azure.

packages/cli/tests/unit/rpc/navie/models/registry.spec.ts
```diff
@@ -0,0 +1,212 @@
+import ModelRegistry from '../../../../../src/rpc/navie/models/registry';
+
+describe(ModelRegistry, () => {
+  beforeEach(() => {
+    (ModelRegistry as any).instance = new (ModelRegistry as any)();
+  });
```
dividedmind, line 6: That's naughty ;) Is there a reason not to just use a `registry = new ModelRegistry()` in the tests?

dividedmind (approved):
I left a few nitpicks, but it looks good in general. A couple thoughts (maybe for future improvements):
- At this point, it feels kind of simplistic to assume that the completion providers are effectively singletons. I imagine someone might want to eg. have several openai-compatible providers to choose from. But this is probably good enough, especially for now.
- I noticed you're using `this.$root.$emit` && `this.$root.$on` — as far as I understand this is removed in Vue 3 and generally frowned upon (and flagged by some linters) because it causes spooky action at a distance between unrelated components; as someone who spent quite some time trying to track down these events I agree that it might make the code less maintainable. Maybe we should consider not adding new usages of the event bus paradigm and instead explicitly pass handlers down the stack up from the instantiation. (I tried to implement a more explicit architecture with `openNewChat`, grep the code (in vscode extension too) to see what I mean.)

packages/navie/test/services/completion-service-factory.spec.ts
dustinbyrne: Yeah, you're right, the mock seems to have no effect. I'll pull it out. I believe I only changed this because I noticed `genMockFromModule` was deprecated.

packages/cli/src/rpc/navie/models/handlers/getConfig.ts
dustinbyrne: I went back and forth on this. Here's the frontend:
image

The frontend doesn't actually know the value of the secret (unless you've just changed it). I considered simply setting a flag that says a value is present, but I worried that rendering a password input of the wrong length may confuse the user. 

E.g., "I just pasted in the value, why does it say it's only four characters long?"

Maybe hiding the value altogether and adding an explicit edit/save button would be a viable alternative.

packages/navie/src/services/openai-completion-service.ts
```diff
@@ -181,11 +181,17 @@ export default class OpenAICompletionService implements CompletionService {
     public readonly modelName: string,
     public readonly temperature: number,
     private trajectory: Trajectory,
-    private readonly messageTokenReducerService: MessageTokenReducerService
+    private readonly messageTokenReducerService: MessageTokenReducerService,
+    private apiUrl?: string,
+    private apiKey?: string
```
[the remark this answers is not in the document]
dustinbyrne: What would you prefer to see here? Additional fields be added to an options object or all of the constructor fields in each completion service moved to options?

dustinbyrne:
> * I noticed you're using `this.$root.$emit` && `this.$root.$on` — as far as I understand this is removed in Vue 3 and generally frowned upon (and flagged by some linters) because it causes spooky action at a distance between unrelated components; as someone who spent quite some time trying to track down these events I agree that it might make the code less maintainable. Maybe we should consider not adding new usages of the event bus paradigm and instead explicitly pass handlers down the stack up from the instantiation. (I tried to implement a more explicit architecture with `openNewChat`, grep the code (in vscode extension too) to see what I mean.)

Emitting an event to the root is essentially how we've been defining outbound callbacks for the code editor extensions to respond to. They're not meant to be used for communication internally within the frontend (though, this is violated to some degree in ChatSearch.vue). We're in a bit of a niche here by providing our frontend as a library to be used by another application.

In this case, we're emitting `select-model` so that the code extension can listen for this event and persist the selected model. No internal component listens for this event.

It looks like Vue 3 removes the [event listener methods (`$on`, `$off`, `$once`)](https://v3-migration.vuejs.org/breaking-changes/events-api.html#_3-x-update) on Vue components altogether. If we migrate we can either introduce our own global event bus or inject some sort of typed callback object. Getting rid of internal dependencies on `$root` events would be a plus, because that pattern should have never occurred to begin with.

packages/navie/src/services/openai-completion-service.ts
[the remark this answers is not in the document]
dividedmind: I think refactor everything to options, WDYT?

packages/cli/src/rpc/navie/models/handlers/getConfig.ts
dividedmind: IMO, during editing the UI should keep the key as a local state and so know the length. If the configuration is later fetched (eg. to edit it) the UI should just show a placeholder (eg. `<unchanged>`) and on update send `null` to the backend if it is indeed still unchanged; so the user can edit some fields (or eg. fix a typo in a URL) without having to paste the key again.

dividedmind:
> > * I noticed you're using `this.$root.$emit` && `this.$root.$on` — as far as I understand this is removed in Vue 3 and generally frowned upon (and flagged by some linters) because it causes spooky action at a distance between unrelated components; as someone who spent quite some time trying to track down these events I agree that it might make the code less maintainable. Maybe we should consider not adding new usages of the event bus paradigm and instead explicitly pass handlers down the stack up from the instantiation. (I tried to implement a more explicit architecture with `openNewChat`, grep the code (in vscode extension too) to see what I mean.)
> 
> Emitting an event to the root is essentially how we've been defining outbound callbacks for the code editor extensions to respond to. They're not meant to be used for communication internally within the frontend (though, this is violated to some degree in ChatSearch.vue). We're in a bit of a niche here by providing our frontend as a library to be used by another application.
> 
> In this case, we're emitting `select-model` so that the code extension can listen for this event and persist the selected model. No internal component listens for this event.
> 
> It looks like Vue 3 removes the [event listener methods (`$on`, `$off`, `$once`)](https://v3-migration.vuejs.org/breaking-changes/events-api.html#_3-x-update) on Vue components altogether. If we migrate we can either introduce our own global event bus or inject some sort of typed callback object. Getting rid of internal dependencies on `$root` events would be a plus, because that pattern should have never occurred to begin with.

What bothers me is an internal component deep within the hierarchy emitting a global event. IMO it's clearer what happens (and is more modular and testable) when it bubbles up with handlers up to the root and then the root can invoke an external handler given by the editor in a property on instantiation (again, like with `openNewChat`). This provides explicit flow, typo and (if vue and typescript gods are generous) type checking.

Anyway, I'm fine with leaving this as is for now, we have plenty of this pattern around the code anyway, I'm just signaling something to think about :)

PR #2262, opened 2025-03-13
fix: Notify users of termination due to content restrictions

dustinbyrne:
This pull request addresses a critical update to the `OpenAICompletionService` by implementing a mechanism to notify users when a response is prematurely terminated due to content filtering restrictions. Additionally, it ensures proper handling of code blocks within responses under these circumstances.

### Features

- **Content Restriction Notification**: The system now detects when the LLM provider terminates a response due to content filtering (e.g., matches with public code). The user will be clearly notified with a message to explain the situation.
  
- **Proper Code Block Handling**: In cases where content filtering occurs during an ongoing code block response, the update will close any open code blocks before notifying the user, thus ensuring cleaner output.

### Tests

- Updated test scenarios verify that users are appropriately informed when content restrictions occur.
- Added checks to ensure that code blocks are correctly closed in the event of a content restriction.

---

image

### DIFF
--- a/packages/navie/src/services/openai-completion-service.ts
+++ b/packages/navie/src/services/openai-completion-service.ts
@@ -173,6 +173,9 @@ function onFailedAttempt(error: unknown) {
   }
 }
 
+const countCodeFences = (text: string): number =>
+  Array.from(text.matchAll(/^\s*?`{3,}\w*?$/gm)).length;
+
 export default class OpenAICompletionService implements CompletionService {
   constructor(
     public readonly modelName: string,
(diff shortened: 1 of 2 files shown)


dividedmind (approved):
Good catch! LGTM. This might need support from the copilot proxies, right?

SgtAirRaid:
Great find @dustinbyrne

dustinbyrne:
With the addition of https://github.com/getappmap/vscode-appland/pull/1075 and https://github.com/getappmap/appmap-intellij-plugin/pull/871 - this is confirmed working in both IDEs.

PR #2278, opened 2025-04-15
Thread log improvements

### LINKED ISSUE #2271: Don't index empty threads
[Threads are flushed to disk upon initialization](https://github.com/getappmap/appmap-js/blob/main/packages/cli/src/rpc/navie/thread/index.ts#L182). This allows clients to reconnect to the same state (e.g., presence of pinned items, code selections) should the RPC service restart for any reason.

However, when being written to disk, they're also indexed. An option should be added to `Thread.flush` which allows a thread to be written to disk _without_ indexing. Then, once a user message (or assistant message) is sent, the thread can be indexed as normal.

### LINKED ISSUE #2272: Navie responds [object Object] on error
Response is not flushed

Error raised from Claude 3.7 Sonnet via Copilot

### LINKED ISSUE #2274: File paths are not rendered
Apply button is present, so path is present
Image

dustinbyrne:
## Overview

This PR includes several fixes and improvements for the new thread event log, focusing on error handling, performance optimizations, and UI enhancements.

### CLI
- Enabled WAL mode for thread index database to improve performance and prevent database lock issues 
- Fixed thread indexing to only occur when users send messages, preventing empty "Untitled" threads (resolves #2271)
- Added strict typing for event errors and improved error message handling (resolves #2272)
- Websocket connection upgrades are now forwarded from the shadow server properly

### Frontend
- Improved file name display in generated code snippets when available (resolves #2274)
- Added fallback to static welcome messages when dynamic message loading fails


packages/cli/src/rpc/navie/services/threadIndexService.ts
```diff
@@ -7,6 +7,7 @@ import { mkdirSync } from 'node:fs';
 
 const INITIALIZE_SQL = `
 PRAGMA foreign_keys = ON;
+PRAGMA journal_mode = WAL;
```
dividedmind, line 10: Are you sure this is the problem with locks? I think it might have been (perhaps in addition to this) that you're preparing statements but then never disposing of them. I've seen this leave stale locks around after the program exits and requiring manual removal of lock files in order to be able to open the database.

In general, I don't think it makes sense to hold prepared statements long-term like this. If it's not in a hot loop, just execute it without preparing. If it is, prepare it before the loop and dispose after.

(Also in some cases it actually can be more efficient, since the statistics might have changed since the program started and a different query plan is better now.)

packages/cli/src/rpc/navie/thread/index.ts
```diff
@@ -154,7 +159,7 @@ export class Thread {
   /**
    * Flush the event log to disk. This is called automatically when a new message is finalized.
    */
-  private async flush() {
+  private async flush({ updateIndex = false }: FlushOptions = {}) {
```
dividedmind, line 164: Maybe updating index should be a separate function? Or, actually, on the other hand — why even flush out threads that are not meant to be indexed? Maybe it's flushing too much, and should only flush when there's something that needs to be indexed. I don't think there'd be any significant risk of data loss if implemented this way.

packages/cli/src/lib/shadowLocalhost.ts
```diff
@@ -29,6 +29,7 @@ function createShadow(server: Server): Promise<Server> {
           server.on('close', () => shadow.close());
           resolve(shadow);
         })
+        .on('upgrade', (...args) => server.emit('upgrade', ...args))
```
dividedmind, line 32: 🥳

packages/cli/tests/unit/rpc/navie/thread/index.spec.ts
```diff
@@ -53,7 +53,13 @@ type ThreadPrivate = Omit<Thread, 'log' | 'listeners' | 'logEvent'> & {
...
   let thread: ThreadPrivate;
   let conversationThread: ConversationThread;
-  let mockNavie: INavie;
+  let mockNavie: {
+    providerName: string;
+    setOption: jest.Mock;
+    ask: jest.Mock;
+    on: jest.Mock;
+    terminate: jest.Mock;
+  };
```
dividedmind, line 62: Why the change? It erases type information.

packages/components/src/components/chat/MarkdownCodeSnippet.vue
```diff
@@ -94,7 +94,7 @@ export default Vue.extend({
     // project directory, we'll just return the full path, what ever it may be.
     shortPath(): string | undefined {
       // If we're using a URI and it's not a file, skip this.
-      if (this.uriComponents && this.uriComponents?.scheme !== 'file') return undefined;
+      if (!this.decodedLocation) return undefined;
```
dividedmind, line 97: The comment above is out of date now, I believe.

dividedmind (approved):
Looks good, thanks! I've had a couple of comments, up to you if you want to apply them. 

One thing that worries me is those undisposed prepared statements — obviously this hasn't been changed in this PR, but perhaps it's a good excuse to address this.

packages/cli/src/rpc/navie/thread/index.ts
dustinbyrne: That's the way it was working when the initial release went out. I had to alter it slightly to flush on certain state changes like including a code selection, pinning a file, etc.

The issue was that the user could open a new chat, pin a file, then change their configuration — triggering a reboot.

When the RPC service comes back up, the frontend will attempt to resubscribe to the thread. If we haven't flushed the events out to disk, it won't exist. We can then start a new thread, but then pinned file state will be missing.

Because of this it's possible that a thread is written to disk while not indexed. It's not likely that the user wants to come back to it later anyway as it's likely empty. Currently, those threads will live on indefinitely in an orphaned state.

packages/cli/tests/unit/rpc/navie/thread/index.spec.ts
dustinbyrne: https://github.com/getappmap/appmap-js/pull/2278/commits/977ab2d9f26ace80334a8e3102cf3982e2c58487

packages/components/src/components/chat/MarkdownCodeSnippet.vue
dustinbyrne: https://github.com/getappmap/appmap-js/pull/2278/commits/c763dd8e922175c2fbb737aed453d04ad24bc2b7

PR #2280, opened 2025-04-17
fix: Thread index database migrations are now atomic

dustinbyrne:
We're still seeing some cases where the database is locked upon startup. This change aims to resolve this entirely by enforcing the process hold a lock file in order to migrate and setting a busy timeout of 3s on all db connections.

This change adds proper database migration handling to the thread index service. Previously, idempotent migrations were run every time the RPC service started, which could lead to race conditions when multiple instances tried to migrate simultaneously.

Now:
- A single process will manage database migrations using file-based locking
- Migrations are only run when necessary (schema version mismatch)
- Added proper error handling and rollback support
- Migrations are wrapped in a transaction for atomicity
- Added retry logic with a reasonable timeout
- Database connections use a busy timeout to handle concurrent access

This change improves reliability and prevents potential database corruption when multiple AppMap instances are running.

Dependencies:
- Added `proper-lockfile` for file locking
- Added corresponding TypeScript type definitions


packages/cli/src/rpc/navie/services/threadIndexService.ts
```diff
@@ -63,11 +71,64 @@ export interface ThreadIndexItem {
...
+      if (!this.shouldMigrate()) {
+        console.info('ThreadIndexService: migration completed by another process');
+        return;
+      }
+
+      console.info(`ThreadIndexService: migrating schema to v${SCHEMA_VERSION}`);
+      try {
+        this.db.exec('PRAGMA journal_mode = WAL');
+        this.db.exec('BEGIN TRANSACTION');
+        this.db.exec(INITIALIZE_DB_SQL);
```
dustinbyrne, line 111: This isn't fully future-proof. We'll need to change this a bit if we ever need to mutate the schema in a non-idempotent manner, but that's an issue for another day.

PR #2297, opened 2025-05-10
feat: Navie review2

kgilpin:


packages/cli/src/lib/git.ts
```diff
@@ -121,18 +121,24 @@ export async function getWorkingDiff(cwd?: string): Promise<string> {
     try {
       untrackedDiffs.push(
         // When using `--no-index`, an exit code of 1 is expected.
-        await execute('git', ['--no-pager', 'diff', '--no-index', DEV_NULL, file], {
-          cwd,
-          exitCode: 1,
-        })
+        await execute(
+          'git',
+          ['--no-pager', 'diff', '--no-index', '--function-context', DEV_NULL, file],
```
kgilpin, line 126: This is more of a feature than a chore, but no big deal.

PR #2320, opened 2025-06-30
fix: Improve worst case performance of file indexing

dustinbyrne:
This pull request introduces several optimizations and a critical bug fix to the file indexing process, significantly reducing the time required to index a project and ensuring the index is correctly maintained between sessions.

#### Summary of Changes

This PR addresses performance bottlenecks and a correctness issue within the file search indexer:

*   **Optimized Binary File Detection with Caching:** To minimize disk I/O during indexing, the binary file detection logic now uses a cache. This cache is pre-populated with a list of common binary and source-code file extensions to avoid hitting the disk for most files. For unknown file types, the result of the check is cached for subsequent lookups.

*   **Improved I/O Concurrency for Faster Indexing:** The file-type checking process has been made asynchronous. This allows multiple file-read operations to be performed concurrently, which enables the OS and drive to optimize the I/O queue. In testing, this change cut the end-to-end indexing time in half for large projects on fast storage. In my testing, this alone cut index times in half.

*   **Prevented Unnecessary Index Re-creation:** A bug was fixed where the file index schema version was being read incorrectly from the metadata table. This caused the version check to fail on every run, leading to the entire index being dropped and rebuilt unnecessarily. This fix ensures the index is correctly preserved.

### DIFF
--- a/packages/search/src/file-index.ts
+++ b/packages/search/src/file-index.ts
@@ -88,13 +88,10 @@ export default class FileIndex {
   get schemaVersion(): string | undefined {
     try {
       const version = this.database
-        .prepare<
-          [],
-          { version?: string }
-        >("SELECT value FROM metadata WHERE key = 'schema_version'")
+        .prepare<[], { value?: string }>("SELECT value FROM metadata WHERE key = 'schema_version'")
         .get();
-      if (!version?.version) return undefined;
-      return String(version.version);
+      if (!version?.value) return undefined;
+      return String(version.value);
     } catch (error) {
       debug('Error retrieving schema version: %s', error);
       return;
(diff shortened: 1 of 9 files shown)


packages/search/src/file-index.ts
```diff
@@ -88,13 +88,10 @@ export default class FileIndex {
...
-        .prepare<
-          [],
-          { version?: string }
-        >("SELECT value FROM metadata WHERE key = 'schema_version'")
+        .prepare<[], { value?: string }>("SELECT value FROM metadata WHERE key = 'schema_version'")
         .get();
-      if (!version?.version) return undefined;
-      return String(version.version);
+      if (!version?.value) return undefined;
+      return String(version.value);
```
dividedmind, line 94: Ah, good catch! My mistake during the migration from wasm sqlite. Perhaps it's worth a test case?

dividedmind (approved):

PR #2337, opened 2025-10-21
Add username to splunk event information

dividedmind:
Splunk telemetry is typically used in corporate environments where username is useful for tracking usage.

### DIFF
--- a/packages/telemetry/src/backends/splunk.ts
+++ b/packages/telemetry/src/backends/splunk.ts
@@ -1,5 +1,6 @@
 import * as http from 'node:http';
 import * as https from 'node:https';
+import * as os from 'node:os';
 import { URL } from 'node:url';
 import * as fs from 'node:fs';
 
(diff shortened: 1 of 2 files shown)


packages/telemetry/src/client.ts
```diff
@@ -218,7 +218,6 @@ export class TelemetryClient implements ITelemetryClient {
       if (this.enabled) {
         this.backend.sendEvent(event);
         this.session.touch();
-        this.backend.flush();
```
hleb-rubanau: @dividedmind seems like a focused but significant behaviour change -- maybe is innnocent and even useful, but I still wonder -- is it worth making flush/noflush a configurable option?

Typically there are tradeoffs between buffering and data persistence. Few questions arise, and answers are not immediately obvious without full code inspection:
1) does the change imply that system will be *always* buffering data? 
2) If so, does it match user expectations and documentation? Is flush guaranteed eventually?
3) Will there be arisk of losing some telemetry (buffered but not flushed), is it acceptable?

I understand that answers could be trivial but I think questions are still worth asking.
Would very much appreciate if you'll find time to reflect on them. Thanks!

PR #2353, opened 2026-03-11
chore(ci): cache yarn packages per native build job

hleb-rubanau:
Swaps 'actions/cache/restore' to 'action/cache' inside actions/setup-node.yml

Supposedly, it should heal always-cold yarn cache in native builds without breaking anything.

(The reason for always cold yarn cache was the lack of corresponding save step anywhere in the flow).


dividedmind (approved):

dividedmind:
A bit of a simplistic fix, but certainly an improvement.

PR #2365, opened 2026-06-24
revive enriched scan:completed telemetry (watch mode)

dividedmind:
Re-emit the batched scan:completed event from the scan watcher, with per-rule and impact-domain finding counts as sorted JSON. Drop the removed scan:started event and the env-var-names attachment.

Bind EventAggregator to a single emitter at construction so cancel() fully detaches it (event + beforeExit listeners) and flushes any final pending batch -- it can no longer leak a process listener.

The scan:completed event was removed in https://github.com/getappmap/appmap-js/commit/c7d0a928661351bc10e51c0531689caa3ff98b1f ("feat: Clean up telemetry"); this reinstates it for corporate visibility.


kgilpin (approved):
LGTM pending tests passing

PR #2375, opened 2026-07-07
feat(cli): appmap sanitize — equality-preserving value tokenization

kgilpin:
## What

A new `appmap sanitize` command that makes committed AppMaps **structurally incapable of carrying a secret**, plus this repo adopting it end to end.

1. **`appmap sanitize <files..>`** (`4b2cd63d2`) — replaces every captured runtime value string with a per-file, equality-preserving token (`<v1>`, `<v2>`, …) assigned by first appearance. Equality within one AppMap still reads as data-flow; zero content characters are retained. Details:
   - **Positional, not content-sniffing**: only the format's captured-data fields are touched (parameter/receiver/return/message values, exception messages, header values with names kept, concrete paths with route templates kept, client URLs). Code-object names, classes, labels are schema and survive.
   - **Kind-annotated tokens** (`<uuid:v3>`, `<int:v7>`) — recognition informs the label, never exemption (session tokens *are* UUIDs).
   - **SQL parameterized, fail-closed**: literals stripped via `normalizeSQL` keeping statement shape; a statement it can't reliably strip (it returns the original — fail-open for its display use) is downgraded to a whole-statement token.
   - **Allowlist**: built-in enumerables (`true`/`false`/`null`…) plus curated vocabularies via `--allow`/`--allow-file`, exact whole-value match only.
   - Idempotent; atomic writes; malformed files skipped, not fatal. 9-test unit spec.
2. **All 20 committed gold-trace baselines sanitized retroactively** (`234284a8c`) — verified against the gold-traces engine's own pipeline: **bless digests identical before/after for all 20** (values are outside the digest by design), so no false re-bless ever. 17% smaller as a side effect.
3. **Review workflow → released `review-action@v1`** (`b3484ec51`), dropping the merged `skills-ref` override; v1's per-matrix sticky comments fix this repo's two-package matrix.
4. **AppMap skills vendored as a submodule** (`491e32dcb`) at `.claude/appland-skills` with committed relative symlinks, replacing machine-local links.

## Why

Gold traces are committed, cloneable, and diff-invisible (`*.appmap.json binary`) — the one place a captured secret would live forever. Truncation caps leakage but retains whole short secrets (PINs, passwords ≤ cap); digests of low-entropy values are crackable offline. Tokenization retains zero bits while preserving the equality structure reviews actually use.

## Review

A behavioral review of this branch ran against the sanitized baselines (asymmetric inputs: sanitized head vs trimmed base): compare clean 20/20, one 🟡 follow-up — unrecognized format fields currently pass through silently; a spec-driven audit + warn-on-unknown-keys mode is the fast follow.

## Follow-ups (not this PR)

- Gold-traces skill switch `trim` → `sanitize` at bless (drafted, gated on this shipping in a release).
- Gold-trace manifest entry for the sanitize command itself.
- Format-version field audit (the 🟡 above).

## Examples

Screenshot 2026-07-08 at 10 45 11 AM


packages/cli/src/lib/sanitizeAppMap.ts
```diff
@@ -0,0 +1,197 @@
...
+// unlinkability across files/revisions is the anti-correlation guarantee.
+
+// Kind recognizers annotate a token with the *shape* of the original value
+// (`<uuid:v3>`) — readability only. Recognition never exempts a value from
+// sanitization: a UUID is exactly the shape session tokens take.
+const KIND_PATTERNS: [string, RegExp][] = [
+  ['uuid', /^\{?[0-9a-fA-F]{8}-(?:[0-9a-fA-F]{4}-){3}[0-9a-fA-F]{12}\}?$/],
+  ['iso8601', /^\d{4}-\d{2}-\d{2}([T ]\d{2}:\d{2}(:\d{2}(\.\d+)?)?(Z|[+-]\d{2}:?\d{2})?)?$/],
+  ['hex', /^(0x)?[0-9a-fA-F]{16,}$/],
+  ['int', /^-?\d+$/],
```
dividedmind, line 39: I still don't think stripping numbers makes much sense; especially up to a certain length.

packages/cli/src/lib/sanitizeAppMap.ts
```diff
@@ -0,0 +1,197 @@
...
+];
+
+// An already-assigned token passes through unchanged, making sanitization
+// idempotent — re-running it on a sanitized AppMap is byte-identical.
+const TOKEN_PATTERN = /^<(?:[a-z0-9]+:)?v\d+>$/;
```
dividedmind, line 46: Should we consider the risk that the application would have sensitive strings like this? The format is fairly generic.

packages/cli/src/lib/sanitizeAppMap.ts
```diff
@@ -0,0 +1,197 @@
...
+  'true',
+  'false',
+  'True',
+  'False',
+  'TRUE',
+  'FALSE',
+  'null',
+  'NULL',
+  'nil',
+  'None',
```
dividedmind, line 63: `undefined` could be a good addition.

packages/cli/src/lib/sanitizeAppMap.ts
```diff
@@ -0,0 +1,197 @@
...
+]);
+
+export class ValueTokenizer {
```
dividedmind: The name `tokenizer` is a bit unfortunate – it has the established meaning of, well, tokenizing a string (splitting it into tokens),which makes reading this unnecessarily difficult. It seems to me this is masking the input strings, not tokenizing them?

packages/cli/src/lib/sanitizeAppMap.ts
```diff
@@ -0,0 +1,208 @@
...
+      sanitizeHeaders(clientRequest.headers, tokenizer);
+    }
+    sanitizeHeaders(event.http_server_response?.headers, tokenizer);
+    sanitizeHeaders(event.http_client_response?.headers, tokenizer);
+
+    const sqlQuery = event.sql_query;
+    if (sqlQuery && typeof sqlQuery.sql === 'string')
+      sqlQuery.sql = sanitizeSql(sqlQuery.sql, sqlQuery.database_type, tokenizer);
+  }
+  return appmap;
```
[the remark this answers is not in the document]
dividedmind: This is a real concern. Perhaps instead of crawling the event updates, it's a good idea to normalize the appmap before continuing.

packages/cli/tests/unit/cmds/sanitize.spec.ts
```diff
@@ -0,0 +1,179 @@
...
+    sanitizeAppMap(appmap);
+    const event = appmap.events[0] as any;
+    expect(event.defined_class).toEqual('app.Auth');
+    expect(event.method_id).toEqual('verify_token');
+    expect(event.path).toEqual('server/auth.py');
+    expect(event.labels).toEqual(['security.authentication']);
+    expect(event.parameters[0].name).toEqual('token');
+    expect(event.parameters[0].class).toEqual('myapp.Token');
+  });
+});
```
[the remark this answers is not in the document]
dividedmind: If we decide to normalize, this should check that event updates are absent and that all events are masked instead.

packages/sequence-diagram/gold_traces/manifest.yaml
```diff
@@ -14,10 +14,11 @@
 schema_version: 1
 
 commands:
-  # Record ONE gold test, then trim its captured values so the committed baseline
-  # stays lean (see gold_traces/trim.mts). appmap-node names the output by the full
-  # describe/it path, so we record the whole test file and trim just the entry's
-  # appmap_path. Recording from the package dir is required for the ts-jest preset.
+  # Record ONE gold test. No trim step here — the engine trims each committed
+  # baseline via `appmap trim`. appmap-node names the output by the full
```
dividedmind, line 18: I wonder, should this be changed to sanitize instead? I suppose this will still use trim now... maybe the comment should be hedged.

dividedmind (changes requested):
My biggest worry are the event updates; we need to also sanitize them, or (preferably) normalize the appmap before sanitization.

packages/cli/src/lib/sanitizeAppMap.ts
```diff
@@ -0,0 +1,208 @@
+import { normalizeSQL } from '@appland/models';
+
+// Sanitize an AppMap so its committed form is structurally incapable of
+// carrying a secret. Every captured runtime value string is replaced with a
+// per-AppMap token (`<v1>`, `<v2>`, …) assigned in order of first appearance,
```
dividedmind, line 8: BTW, is it just me, or do these look awfully like version numbers? I wonder if this could be confusing.

kgilpin: Yeah, I don't really love "v" either. Maybe "str1", "str2"? LMK what you propose.

packages/sequence-diagram/gold_traces/manifest.yaml
kgilpin: Regarding normalization, should the `sanitize` command do that? It's a bit of an unexpected side effect, though harmless I suppose.

packages/cli/src/lib/sanitizeAppMap.ts
kgilpin: Yeah; apparently "tokenization" is the security lingo for what we would call masking. So maybe `ValueMasker`.

packages/cli/src/lib/sanitizeAppMap.ts
kgilpin: Roger

packages/cli/src/lib/sanitizeAppMap.ts
kgilpin: The rationalization for this is things like PINs and OTP codes. We are unlikely to ever see this in practice - but I'm not sure how to make the tradeoff, now that we are trying to "guarantee" security (as much as such a thing can ever be done).

packages/cli/tests/unit/cmds/sanitize.spec.ts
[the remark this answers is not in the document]
kgilpin: Yeah. We could also add `sanitized: true` to the AppMap metadata section; and maybe `trimmed: true`.

packages/cli/src/lib/sanitizeAppMap.ts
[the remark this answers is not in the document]
kgilpin: As I commented elsewhere, I like normalization as an implementation strategy, but it's an unexpected side effect for `sanitize`.

packages/cli/src/lib/sanitizeAppMap.ts
dividedmind: Yeah; I know it's a nitpick, but it catches me off guard every time I look at it. Shouldn't be difficult to get an LLM to do it quickly before it's set in stone ;) `ValueMasker` sounds fine.

packages/sequence-diagram/gold_traces/manifest.yaml
dividedmind: On one hand, perhaps; on the other, if we're storing and comparing gold traces it's probably a good idea to have them normalized anyway. I don't see the downside of that? Canonicalizing would be a step too far. Here's quick gemini's analysis https://gist.github.com/dividedmind/1ea9dca5419a2248723921f808707cbd

packages/cli/src/lib/sanitizeAppMap.ts
dividedmind: Even `<val1>`... would be more obvious.

PR #2377, opened 2026-07-11
fix(cli): sanitize SQL in the class map, not just events

kgilpin:
## Problem
`appmap sanitize` promises a committed AppMap that is "structurally incapable of
carrying a secret", but `sanitizeAppMap` only walked `events`, `eventUpdates`, and
`metadata` — **not the `classMap`**. A `query` code object's `name` is the raw SQL
statement, literals included, so a query's secret/PII literals survived there.

On a fixture, after `appmap sanitize`:

```
event sql_query.sql :  INSERT INTO "api_keys" (…) VALUES (?, ?)                      ← sanitized
classMap query node :  INSERT INTO "api_keys" (…) VALUES ('admin', 'example api key') …  ← RAW
literal 'admin' in classMap? true      ← leaks
metadata.sanitized present?  true      ← and stamped "sanitized"
```

The design comment even states "code object names … are schema … never modified" —
but a query node's name is a SQL statement, not schema, so the assumption breaks for
that one node type.

## Fix
Walk the `classMap` and run each `type: 'query'` node's SQL through the existing
`sanitizeSql` (parameterize via `normalizeSQL`, or fail closed to a `<vN>` token on
residue), sharing the same `ValueMasker`. The SQL→adapter map is captured from the
events **before** they're rewritten, so class-map SQL sanitizes identically to its
event. Everything else in the class map (packages, classes, functions, routes) is
schema and untouched.

## Verification
- New tests: no event literal survives **anywhere** in the sanitized output (class
  map included); a class-map query that can't be parameterized fails closed to a token.
- End-to-end on the fixture: class-map SQL becomes `VALUES (?, ?)`, and `'admin'` /
  `'example api key'` are gone from the class map.
- Full `sanitize.spec.ts` 25/25; `verify` (lint + tsc) clean.

### DIFF
--- a/packages/cli/src/lib/sanitizeAppMap.ts
+++ b/packages/cli/src/lib/sanitizeAppMap.ts
@@ -23,7 +23,10 @@ export const CLI_VERSION: string = require('../../package.json').version;
 // Sanitization is positional, driven by the AppMap format: it touches only
 // the fields defined to hold captured runtime data. Code object names,
 // classes, paths, labels, and route templates are schema, not data, and are
-// never modified.
+// never modified — with one exception: a `query` code object's name IS a SQL
+// statement, literals included, so it is parameterized exactly like an event's
+// sql_query (otherwise a literal that was scrubbed from the events would still
+// survive in the class map).
 //
 // The per-file token namespace is deliberate: unlinkability across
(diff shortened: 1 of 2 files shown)


dividedmind (approved):
Good catch, lgtm.

PR #2401, opened 2026-09-17
fix: leave MongoDB query statements alone when normalizing SQL

evlawler:
The Node and Java agents are gaining `sql_query` events for MongoDB operations, with `database_type: "mongodb"` and a statement whose argument values are already placeholders:

```
db.users.updateOne({"_id": ?}, {"$set": {"name": ?}, "$inc": {"n": ?}})
```

`normalizeSQL` has no case for that database type, so the fallback obfuscation applies. It treats the quoted keys as string literals and a pair of `$operators` as a dollar-quoted string, and the statement above becomes `db.users.updateOne({?: ?}, {?: {?: ?}, ?: {?: ?}})` in the `sqlNormalized` index, in the sequence diagram digest (through `abstractSqlAstJSON`, which falls back to `normalizeSQL` when the parser fails) and in change reports.

Fix. Return the statement unchanged for `mongodb`. `analyzeSQL` still returns null for it, so no tables are reported, which is the existing behavior for anything the SQL parser cannot read.

Test. `sql.normalize.spec.js` covers a known dialect, the fallback, and MongoDB statements passing through untouched.

Companion changes: getappmap/appmap-node#239 and getappmap/appmap-java#330 emit the events.

Written by Claude in a Claude Code session for Elizabeth Lawler. The commit carries Claude as author.

https://claude.ai/code/session_01LntddoBsqjRDRBepBx7oLZ


evlawler:
Closing. This change only existed to let MongoDB statements pass through `normalizeSQL` for getappmap/appmap-node#239, and the discussion there concluded that recording Mongo operations as `sql_query` is not the right shape. The redesign will give non-SQL queries their own structure, so this patch is not needed.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #2403, opened 2026-09-18
Fix broken/noisy npm/npx installs (fsevents build failure, stale engines)

dividedmind:
## Summary

`npx @appland/appmap` and `npm install -g @appland/appmap` produce confusing `gyp ERR!` output, and on Linux/Windows `npm install -g` can fail outright. Root causes, reproduced against the real published `3.203.0` tarball:

- **`fsevents` breaks global installs.** `fsevents` (chokidar's macOS-only optional dependency) was only a transitive optionalDependency, two levels below the installed package (`appmap → chokidar → fsevents`). npm reliably tolerates a failed optional dependency of the package you install directly, but not one nested this deep in a large dependency graph: `npm install -g @appland/appmap` hard-failed on Linux, rolling back the whole install and leaving a dangling `bin/appmap` symlink, because the current `fsevents` release ships only a prebuilt macOS binary with no build sources for `node-gyp` to fall back to. `@appland/scanner` has the identical `chokidar` dependency and is equally exposed. A plain local `npm install` and `npx` were unaffected — only global installs hit this.
- **`engines.node: ">15"` is fiction.** Both packages already depend on packages (`express@5`, `langchain`) that require Node >=18, and `better-sqlite3@^12` only ships prebuilt binaries for `20.x || 22.x-26.x` — anything else falls back to a slow, noisy `node-gyp` source compile.

## Changes

- `packages/cli` and `packages/scanner`: declare `fsevents` directly as an `optionalDependencies` entry (same range chokidar itself requests), so npm's os-based filtering skips it outright on non-macOS instead of attempting and failing to build it.
- `packages/cli`, `packages/scanner`, `packages/search`: bump `engines.node` from `>15` to `>=18`, matching reality.
- `packages/search`: add a `typecheck` script so it participates in `yarn verify` going forward.

An earlier commit also wrapped the `better-sqlite3` open call in `packages/search` with a friendlier error message, but Copilot's review correctly pointed out it couldn't work — `better-sqlite3` is imported statically at module load time, so a failed native binding throws before that try/catch ever runs. That change was reverted; the `search` package's `engines`/`typecheck` fix (a `fix:`-tagged commit touching its files) is enough on its own to get `semantic-release-monorepo` to cut a release, since its commit filtering is purely file-path based.

## Verification

Downloaded the real published `@appland/appmap@3.203.0` tarball, patched only the `engines`/`optionalDependencies` fields to match this branch, repacked it, and ran `npm install -g` against a throwaway prefix — no registry publish involved:

| | Unpatched (control) | Patched (this branch's fields) |
|---|---|---|
| `npm install -g` | Exit 1, `gyp ERR!` on `fsevents`, rolled back | Exit 0, no gyp/fsevents activity, `appmap --version` runs |

Also verified `yarn lint`/`tsc --noEmit` pass for the touched `search` files.

https://claude.ai/code/session_01FnDF7psx9u4BqXwupSFRWz


kgilpin:
This may have failed in CI

dividedmind:
> This may have failed in CI

Yeah thanks, it looks like a fluke, I'll prod a rerun.

